PR #9, opened 2024-05-06
NEW: @W-15652642@ - Add in starting files to prepare monorepo

stephen-carter-at-sf:
Added in monorepo files, eslint (using newest recommended flat config), jest, github run_tests script, etc. Note that I'm just using npm run all for now... but the build script might evolve in the future to break out the steps if we find it necessary. Just trying to keep things simple for now.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,90 @@
+# Contributing Guide
+> Currently, we aren't accepting external contributions. To know when we open for contributions, periodically check this guide.
+
+To contribute, read this guide to understand its operational governance model. This guide also provides contribution requirements and recommendations. Thanks for your interest and contributions.
+
+# Governance Model
+
+## Salesforce Sponsored
+
+The intent and goal of open sourcing this project is to increase the contributor and user base. Note that Salesforce employees have `admin` rights and are the final arbiters which contributions are accepted.
+
+# Issues, requests & ideas
+
+To submit issues, enhancement requests, and discuss ideas, create an issue on the [Issues page](https://github.com/forcedotcom/code-analyzer-core/issues).
+
+### Bug Reports and Fixes
+-  If you find a bug, search for it in the [Issues](https://github.com/forcedotcom/code-analyzer-core/issues), and if it isn't already tracked,
+   [create a new issue](https://github.com/forcedotcom/code-analyzer-core/issues/new). Fill out the "Bug Report" section of the issue template. Even if an issue is closed, feel free to comment and add details, it will still be reviewed.
+-  Issues that have been confirmed as a bug will be labelled `bug`.
+-  If you want to submit a bug fix, [create a pull request](#creating_a_pull_request) and mention the issue number.
+-  In your bug fix pull request, include tests that isolate the bug and verify that it's fixed.
+
+### New Features
+-  To add new functionality, describe the problem you want to solve - or the enhancement you identifed - in a [new issue](https://github.com/forcedotcom/code-analyzer-core/issues/new).
+-  Issues that are identified as a feature request are labelled `enhancement`.
+-  Before writing the code for your newly requested feature, first wait for feedback from the project maintainers. In some cases, requested `enhancements` don't align well with the current project objectives.
+
+### Tests, Documentation, Miscellaneous
+-  We welcome your contributions to:
+    - improving tests
+    - clarifying documentation
+    - requesting alternative implementations
+-  If it's a trivial change, go ahead and [create a pull request](#creating_a_pull_request) with the changes you have in mind.
+-  If your request is larger in size, [open an issue](https://github.com/forcedotcom/code-analyzer-core/issues/new) to get feedback on your idea.
+
+# Contribution Checklist
+
+- [x] Clean, simple, well styled code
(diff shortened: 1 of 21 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,90 @@
+# Contributing Guide
```
stephen-carter-at-sf, line 1: This is the new standard template (which we recently used in our run_code_analyzer repo) modified to fit this repo.

package.json
```diff
@@ -0,0 +1,43 @@
...
+    "build": "npm run build --workspaces",
+    "test": "jest --coverage",
+    "lint": "npm run lint --workspaces --if-present",
+    "package": "npm run package --workspaces --if-present",
+    "all": "npm run build && npm run test && npm run lint && npm run package",
+    "clean": "npm run clean --workspaces",
+    "postclean": "rimraf coverage"
+  },
+  "devDependencies": {
+    "jest": "^29.0.0",
```
stephen-carter-at-sf, line 17: Note to get code coverage numbers to role up to the top, I added jest as a top level dependency and have the "test" script above work on all packages as a whole. This way we can check the coverageThreshold at the top level instead of at each independent's subproject level.

packages/code-analyzer-core/eslint.config.mjs
```diff
@@ -0,0 +1,9 @@
+// @ts-check
```
stephen-carter-at-sf, line 1: If you are unfamiliar with these new eslint config files, read https://typescript-eslint.io/getting-started

Unfortunately, we don't get a whole lot of intellij editor support for this yet. But it is highly recommended on the internet for folks to move to this new eslint configuration model even though we might wait for intellij to catch up.

packages/code-analyzer-core/tsconfig.json
```diff
@@ -0,0 +1,12 @@
+{
+  "extends": "./tsconfig.build.json",
```
stephen-carter-at-sf, line 2: This turns out to be the industry recommended way of allowing your IDE to incorporate tests, without shipping tests in the dist folder. Although we end up with multiple tsconfig files... things just work really nicely and our packages will be constructed using the correct src folder as the root directory.

tsconfig.json
```diff
@@ -0,0 +1,6 @@
+{
+  "extends": "./tsconfig.base.json",
+  "compilerOptions": {
+    "noEmit": true
```
stephen-carter-at-sf, line 4: This top level tsconfig.json file really helps out IDEs like vscode and intellij know that the root project is a typescript project. Using noEmit = true is needed for jest to work properly from the top level as well. But again, all our true build and packaging uses the tsconfig.build.json files anyway - so this trick (used all over the place by top packages) seems to give you the best of both worlds... a clean build - with a great development experience.

I used https://github.com/typescript-eslint/typescript-eslint as a reference project to learn about all these tricks.

package-lock.json
jfeingold35, line 1: Question: Had we decided that we were using NPM for the core module instead of yarn? That sounds right, but I genuinely don't remember.

package.json
jfeingold35: Does that mean that the lower bound applies to the average of all projects, or does it still apply to each project individually?

package.json
```diff
@@ -0,0 +1,43 @@
...
+    "ts-jest": "^29.0.0",
+    "rimraf": "*"
+  },
+  "jest": {
+    "preset": "ts-jest",
+    "testEnvironment": "node",
+    "coverageThreshold":  {
+      "global": {
```
jfeingold35, line 25: 95% is pretty ambitious. Is the idea that we shoot for 95% so we have room to lower it if we need while still staying well within the 75% org-wide mandate?

jfeingold35 (approved):
I believe the standard is for every file to have the BSD-3 license short-version in addition to the LICENSE file. Maybe i'm misremembering?
Also, I'm seeing lots of two-space indents. My personal preference is to use one tab, and then everyone can set the tab to display as however many spaces they feel like.

package.json
stephen-carter-at-sf: Basically this top level jest dependency is for running the entire repo's typescript tests as a whole. Individual projects can have their own jest versions if they want to... with their own coverage limits, etc. Primarily, this is to satisfy the IDE experience of being able to develop all projects at once and run any test anywhere without having to have separate IDEs. Without this top level jest, then IntelliJ and VSCode get confused as to what is the working directory for each test run. It took me a while to understand this, but it is what a lot of other big typescript monorepos do out there.

package.json
stephen-carter-at-sf: My goal is actually 100%. So 95% was to give a tiny bit of wiggle room for now. If we need to drop this lower, then we should have a conversation first to see why we can't keep this high level.

jag-j: I disagree. Code coverage is notorious for giving false confidence. It is a good indicator at a lower number i.e. when a project has say < 50% code coverage, you know there is room for improvement. But it doesn't add much value beyond the 75-80% mark.

package-lock.json
stephen-carter-at-sf: Starting with npm and going to keep the dependencies super clean so that by the time vs code extension depends on core, hopefully we won't have anymore packaging issues.

PR #10, opened 2024-05-07
NEW: @W-15652656@: Add public interfaces before adding any business logic

stephen-carter-at-sf:
As a first step, I'm simply adding in all the public interfaces so that we are ready at any time to publish a 0.1.0-alpha package of the core module so that client development may begin in parallel. I went ahead and added in interfaces for both the core package and the engine-api packages.


package.json
```diff
@@ -23,10 +23,10 @@
...
     "coverageThreshold":  {
       "global": {
-        "branches": 95,
-        "functions": 95,
-        "lines": 95,
-        "statements": 95
+        "branches": 80,
+        "functions": 80,
+        "lines": 80,
+        "statements": 80
```
stephen-carter-at-sf, line 29: Lowering per the suggestion of Jag and Josh. But we are still at 100% coverage and I hope to keep it up there as much as possible.

packages/code-analyzer-core/src/code-analyzer.ts
```diff
@@ -0,0 +1,20 @@
...
+import { Event } from "./events"
+
+export type RunOptions = {
+    filesToInclude: string[]
+    entryPoints?: string[]
+}
+
+// Temporarily making this an interface
+export interface CodeAnalyzer {
+    addEnginePlugin(enginePlugin: EnginePlugin): void
```
stephen-carter-at-sf, line 13: I'm purposely choosing not to add in any documentation comments since this is all subject to change. Only after things get stable does it make sense for us to add in all the comments for us to manage.

packages/code-analyzer-core/src/events.ts
```diff
@@ -0,0 +1,37 @@
...
+    Debug = 4,
+    Fine = 5
+}
+
+export type LogEvent = {
+    type: EventType.LogEvent,
+    timestamp: Date,
+    logLevel: LogLevel,
+    message: string
+}
```
jfeingold35, line 22: We might be able to reuse some of the Log4JS code that we use in the current scanner repo. Might not have to reinvent the wheel.

packages/code-analyzer-core/src/results.ts
```diff
@@ -0,0 +1,34 @@
+import { Rule, SeverityLevel } from "./rules"
+
+export interface CodeLocation {
+    getFile(): string
+    getStartLine(): number
+    getStartColumn(): number
+    getEndLine(): number
+    getEndColumn(): number
```
jfeingold35, line 8: Should these three be `number|null`?

packages/code-analyzer-core/src/rules.ts
```diff
@@ -0,0 +1,27 @@
+export enum SeverityLevel {
+    High = 1,
```
jfeingold35, line 2: Did we decide against having a `Critical = 0` for engine failures?

jfeingold35 (approved):

packages/code-analyzer-core/src/results.ts
stephen-carter-at-sf: Will do in my next PR. Sorry I missed your comments.

packages/code-analyzer-core/src/rules.ts
stephen-carter-at-sf: Oh that's a good idea. Will update in my next PR.

PR #11, opened 2024-05-08
NEW: @W-15652656@: Add in manual workflow to publish to npm

stephen-carter-at-sf:
Added in a workflow that will allow us to manually publish to npm for now as we are internally developing these packages. In the future we'll add in smoke tests, etc. But I wanted to keep it simple for now.

### DIFF
--- a/packages/code-analyzer-core/package.json
+++ b/packages/code-analyzer-core/package.json
@@ -7,7 +7,7 @@
   "homepage": "https://developer.salesforce.com/docs/platform/salesforce-code-analyzer/overview",
   "repository": {
     "type": "git",
-    "url": "https://github.com/forcedotcom/code-analyzer-core.git",
+    "url": "git+https://github.com/forcedotcom/code-analyzer-core.git",
     "directory": "packages/code-analyzer-core"
   },
   "main": "dist/index.js",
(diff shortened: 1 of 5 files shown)


packages/code-analyzer-core/package.json
```diff
@@ -7,7 +7,7 @@
   "homepage": "https://developer.salesforce.com/docs/platform/salesforce-code-analyzer/overview",
   "repository": {
     "type": "git",
-    "url": "https://github.com/forcedotcom/code-analyzer-core.git",
+    "url": "git+https://github.com/forcedotcom/code-analyzer-core.git",
```
stephen-carter-at-sf, line 10: Updated this per the warning that a dry run of npm publish gives us. npm pkg fix is what updated this.

packages/code-analyzer-core/src/rules.ts
```diff
@@ -1,4 +1,5 @@
 export enum SeverityLevel {
+    Critical = 0, // Only used in the event of an engine failure
```
stephen-carter-at-sf, line 2: Per the suggestion of Josh that I missed in the last submission.

jfeingold35 (approved):

PR #12, opened 2024-05-10
NEW: @W-15652729@: Implement rule selection ability to core (initially without configuration)

stephen-carter-at-sf:


packages/code-analyzer-core/test/add-engines.test.ts
```diff
@@ -0,0 +1,74 @@
...
+    beforeEach(() => {
+        codeAnalyzer = new CodeAnalyzer(CodeAnalyzerConfig.withDefaults());
+        logEvents = [];
+        codeAnalyzer.onEvent(EventType.LogEvent, (event: LogEvent) => logEvents.push(event));
+    })
+
+    it('When adding engine plugin then all its engines are correctly added', () => {
+        codeAnalyzer.addEnginePlugin(new stubs.StubEnginePlugin());
+
+        expect(codeAnalyzer.getEngineNames().sort()).toEqual(["stub1","stub2"])
```
jag-j: To test the boundary conditions, can we test engines with one stub engine as well?

packages/code-analyzer-core/test/stubs.ts
```diff
@@ -0,0 +1,227 @@
+import * as engApi from "@salesforce/code-analyzer-engine-api"
+
+/**
+ * StubEnginePlugin - A plugin stub with preconfigured outputs to help with testing
+ */
+export class StubEnginePlugin extends engApi.EnginePluginV1 {
+    getAvailableEngineNames(): string[] {
+        return ["stub1", "stub2"];
```
jag-j: Nit pick: Too many "stubs". Can we rename this stub-engine-1 and stub-engine-2 or something similar?

packages/code-analyzer-core/src/messages.ts
```diff
@@ -0,0 +1,47 @@
+const messageCatalog : { [key: string]: string } = {
+    EngineFromFutureApiDetected:
```
jag-j, line 2: Do you plan to do doc review for these? Not sure if that would be an overkill. Worth asking @jshackell-sfdc if she follows any doc review for SDR library.

stephen-carter-at-sf: I plan on doing a doc review before GA... there will be too many changes until then and its best to do it all at once then every few days. This is one of the reasons why I'm putting all the messages into a message catalog... to make it easier to review in the end.

packages/code-analyzer-core/test/add-engines.test.ts
stephen-carter-at-sf: The FutureEnginePlugin returns one engine and thus effectively tests this already on line 21.

packages/code-analyzer-core/test/stubs.ts
stephen-carter-at-sf: Sure. Renamed them to stubEngine1 and stubEngine2

jag-j (approved):

packages/code-analyzer-core/test/rule-selection.test.ts
```diff
@@ -0,0 +1,155 @@
...
+    LogEvent,
+    Rule,
+    RuleSelection,
+    RuleType,
+    SeverityLevel
+} from "../src";
+import { SeverityLevel as EngApi_SeverityLevel } from "@salesforce/code-analyzer-engine-api"
+import {StubEnginePlugin} from "./stubs";
+
+describe('Tests for selecting rules', () => {
```
jfeingold35, line 14: I'd like to see a test for when a tag overlaps with a rule name, or some other weird corner cases like that. My assumption would be that the rule with that name and all rules with the tag would be returned.

stephen-carter-at-sf: Yes that is a correct assumption. I'll add a test for this in my next PR.

PR #13, opened 2024-05-15
NEW: @W-15652762@: Add configuration parsing and validation abilities

stephen-carter-at-sf:
Adding in configuration parsing and validation which will allow clients to now pass in configuration data according the schema:

log_folder
- Specify a location to write log files to. If not specified, then your machine’s default temporary directory will be used.

rules.<engine_name>.<rule_name>.severity
- Specify the severity 1 (Critical), 2 (High), 3 (Moderate), 4 (Low), or 5 (Info) level as an integer that you want to set for the specified rule.

rules.<engine_name>.<rule_name>.tags
- Specify a list of tags to set for the specified rule.

engines.<engine_name>.<engine_specific_property>
- For the specified engine, you can specify any engine specific properties and values as needed.

Clients may pass in a yaml or json file, a yaml or json string, or an object.

### DIFF
--- a/.github/workflows/verify-pr.yml
+++ b/.github/workflows/verify-pr.yml
@@ -18,7 +18,10 @@ jobs:
               exit 1
           fi
   run_tests:
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-node@v4
--- a/packages/code-analyzer-core/src/code-analyzer.ts
+++ b/packages/code-analyzer-core/src/code-analyzer.ts
@@ -4,25 +4,11 @@ import {Event, EventType, LogLevel} from "./events"
 import {getMessage} from "./messages";
 import * as engApi from "@salesforce/code-analyzer-engine-api"
 import {EventEmitter} from "node:events";
+import {CodeAnalyzerConfig} from "./config";
 
 export type RunOptions = {
     filesToInclude: string[]
     entryPoints?: string[]
-}
-
-// Currently we have no configuration abilities implemented. So this is just a placeholder for now.
-export class CodeAnalyzerConfig {
-    public static withDefaults() {
-        return new CodeAnalyzerConfig();
-    }
-
-    private constructor() {
-    }
-
-    public getEngineSpecificConfig(_engineName: string): engApi.ConfigObject {
-        // To be implemented soon
-        return {}
-    }
 }
 
 export class CodeAnalyzer {
@@ -42,7 +28,7 @@ export class CodeAnalyzer {
         const enginePluginV1: engApi.EnginePluginV1 = enginePlugin as engApi.EnginePluginV1;
 
         for (const engineName of getAvailableEngineNamesFromPlugin(enginePluginV1)) {
-            const engConf: engApi.ConfigObject = this.config.getEngineSpecificConfig(engineName);
+            const engConf: engApi.ConfigObject = this.config.getEngineSettingsFor(engineName);
             const engine: engApi.Engine = createEngineFromPlugin(enginePluginV1, engineName, engConf);
             this.addEngineIfValid(engineName, engine);
         }
--- a/packages/code-analyzer-core/src/config.ts
+++ b/packages/code-analyzer-core/src/config.ts
@@ -6,6 +6,14 @@ import * as yaml from 'js-yaml';
 import {getMessage} from "./messages";
 import {SeverityLevel} from "./rules";
 
+const FIELDS = {
+    LOG_FOLDER: 'log_folder',
+    RULE_SETTINGS: 'rule_settings',
+    ENGINE_SETTINGS: 'engine_settings',
+    SEVERITY: 'severity',
+    TAGS: 'tags'
+} as const
+
 export type EngineRuleSettings = {
     severity?: SeverityLevel
     tags?: string[]
@@ -37,11 +45,11 @@ export class CodeAnalyzerConfig {
         }
         const fileContents: string = fs.readFileSync(file, 'utf8');
 
-        const fileExt : string = file.split('.').pop()?.toLowerCase() || "";
+        const fileExt : string = path.extname(file).toLowerCase();
 
-        if (fileExt == 'json') {
+        if (fileExt == '.json') {
             return CodeAnalyzerConfig.fromJsonString(fileContents);
-        }  else if (fileExt == 'yaml' || fileExt == 'yml') {
+        }  else if (fileExt == '.yaml' || fileExt == '.yml') {
             return CodeAnalyzerConfig.fromYamlString(fileContents);
         } else {
             throw new Error(getMessage('ConfigFileExtensionUnsupported', file, 'json,yaml,yml'))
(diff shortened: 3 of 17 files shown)


packages/code-analyzer-core/src/code-analyzer.ts
```diff
@@ -42,7 +28,7 @@ export class CodeAnalyzer {
         const enginePluginV1: engApi.EnginePluginV1 = enginePlugin as engApi.EnginePluginV1;
 
         for (const engineName of getAvailableEngineNamesFromPlugin(enginePluginV1)) {
-            const engConf: engApi.ConfigObject = this.config.getEngineSpecificConfig(engineName);
+            const engConf: engApi.ConfigObject = this.config.getEngineSettingsFor(engineName);
```
jfeingold35, line 31: Is there a reason you switched from "engineConfig" to "engineSettings"?

packages/code-analyzer-core/src/config.ts
```diff
@@ -0,0 +1,181 @@
...
+    public static fromFile(file: string) {
+        file = toAbsolutePath(file);
+        if (!fs.existsSync(file)) {
+            throw new Error(getMessage('ConfigFileDoesNotExist', file));
+        }
+        const fileContents: string = fs.readFileSync(file, 'utf8');
+
+        const fileExt : string = file.split('.').pop()?.toLowerCase() || "";
+
+        if (fileExt == 'json') {
```
jfeingold35: Question: Are we supporting JSON-formatted configs? I thought we were just doing YAML.

packages/code-analyzer-core/src/config.ts
```diff
@@ -0,0 +1,181 @@
...
+    }
+
+    public static fromFile(file: string) {
+        file = toAbsolutePath(file);
+        if (!fs.existsSync(file)) {
+            throw new Error(getMessage('ConfigFileDoesNotExist', file));
+        }
+        const fileContents: string = fs.readFileSync(file, 'utf8');
+
+        const fileExt : string = file.split('.').pop()?.toLowerCase() || "";
```
jfeingold35: Node gives us this sort of thing for free: https://nodejs.org/api/path.html#pathextnamepath

packages/code-analyzer-core/src/config.ts
```diff
@@ -0,0 +1,181 @@
...
+        return this.config.rule_settings[engineName] || {};
+    }
+
+    public getEngineSettingsFor(engineName: string): engApi.ConfigObject {
+        return this.config.engine_settings[engineName] || {};
+    }
+}
+
+function validateAndExtractLogFolderValue(data: object): string {
+    if (!('log_folder' in data)) {
```
jfeingold35: Can `log_folder` and others be constants?

packages/code-analyzer-core/src/config.ts
```diff
@@ -0,0 +1,181 @@
...
+    if ('tags' in valueObj ) {
+        validateTagsValue(valueObj['tags'], `${valueKey}.tags`);
+    }
+}
+
+function validateObject(value: unknown, valueKey: string) {
+    return validateType<object>('object', value, valueKey);
+}
+
+function validateType<T>(expectedType: string, value: unknown, valueKey: string): T {
```
jfeingold35, line 136: Did you consider doing this with type guards instead of casting? https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards

.github/workflows/verify-pr.yml
```diff
@@ -5,6 +5,7 @@ on:
 
 jobs:
   validate_pr_title:
+    if: github.base_ref == 'dev' && github.event.action == 'opened' || github.event.action == 'reopened' || github.event.action == 'edited'
```
jfeingold35: Is this `if` really necessary? Also, for my own sanity, I'd prefer if it was `x && (y || z)` as opposed to `x && y || z`.

.github/workflows/verify-pr.yml
```diff
@@ -18,7 +19,11 @@ jobs:
               exit 1
           fi
   run_tests:
-    runs-on: ubuntu-latest
+    if: github.event.action == 'opened' || github.event.action == 'reopened' || github.event.action == 'synchronize'
```
jfeingold35: Is this just so the tests don't re-run every time you change the title?

packages/code-analyzer-core/src/code-analyzer.ts
stephen-carter-at-sf: The name of the field in the config file is engine_settings.(engine_name) and so it makes sense. But this getEngineSettingsFor is only to be used by core anyway... so the name doesn't matter. Client's shouldn't need to ever user this (although they can if they want to debug).

packages/code-analyzer-core/src/config.ts
stephen-carter-at-sf: Supporting json was super easy for core to do as you see in the code. The client can choose whatever they want. Our CLI will probably only support yaml files but I don't want to block other clients from passing in config however they wish. For example, vscode may have better json utilities... or it might want to construct the object itself.

packages/code-analyzer-core/src/config.ts
stephen-carter-at-sf: I actually tried... but as soon as I make 'log_folder' a variable or a constant, then the typescript compiler complains on things like data[FIELDS.log_folder] that it hasn't been verified that it even exists. I can show you in a video chat if you want to see what I'm talking about.

packages/code-analyzer-core/src/config.ts
stephen-carter-at-sf: If I use type guards then it'll throw its own error message which doesn't point to the place in the config file that needs updating. The whole reason for all this validation code (which is tedious) is to make for a good user experience when something is incorrect. Validating the output of parsed code and helping the user figure out what went wrong... is different than simply erroring when something doesn't match.

.github/workflows/verify-pr.yml
stephen-carter-at-sf: Sorry this was a temporary experiment that I did that I've since removed.

.github/workflows/verify-pr.yml
stephen-carter-at-sf: Sorry this was a temporary experiment that I did that I've since removed.

packages/code-analyzer-core/src/config.ts
stephen-carter-at-sf: Fixed. Thanks.

packages/code-analyzer-core/src/config.ts
stephen-carter-at-sf: Ah yes... I figured it out. Updated the code to make these constants.

When I originally tried

const FIELDS = {
    LOG_FOLDER: 'log_folder',
    RULE_SETTINGS: 'rule_settings',
    ENGINE_SETTINGS: 'engine_settings',
    SEVERITY: 'severity',
    TAGS: 'tags'
}

it didn't work. Apparently I needed add in the "as const" to the end and the compiler then knew it wasn't dynamic anymore:

const FIELDS = {
    LOG_FOLDER: 'log_folder',
    RULE_SETTINGS: 'rule_settings',
    ENGINE_SETTINGS: 'engine_settings',
    SEVERITY: 'severity',
    TAGS: 'tags'
} as const

packages/code-analyzer-core/src/config.ts
jfeingold35: Fair enough.

packages/code-analyzer-core/src/config.ts
jfeingold35: I think that makes sense, though you may want to look at the scanner's `Config.ts` file. We're doing some similar  stuff in there, and it might be more efficient than what you're doing here.

jfeingold35 (approved):

packages/code-analyzer-core/src/config.ts
stephen-carter-at-sf: No inside of Config.ts all we do is 

this.configContent = JSON.parse(configFileContent) as ConfigContent;

which again uses the type casting as a means of validation. This doesn't give the best error messages.

jfeingold35 (approved):

PR #14, opened 2024-05-17
NEW: @W-15652775@ Final updates config parsing and rule selection

stephen-carter-at-sf:
* I improved the code readability and quality of config parsing
* I applied rule override information from config to rule selection
* I made rule selection more tolerant with case insensitivity and the ability to choose rules of severity based on number or name

### DIFF
--- a/packages/code-analyzer-core/src/code-analyzer.ts
+++ b/packages/code-analyzer-core/src/code-analyzer.ts
@@ -4,7 +4,7 @@ import {Event, EventType, LogLevel} from "./events"
 import {getMessage} from "./messages";
 import * as engApi from "@salesforce/code-analyzer-engine-api"
 import {EventEmitter} from "node:events";
-import {CodeAnalyzerConfig} from "./config";
+import {CodeAnalyzerConfig, RuleOverride} from "./config";
 
 export type RunOptions = {
     filesToInclude: string[]
@@ -28,7 +28,7 @@ export class CodeAnalyzer {
         const enginePluginV1: engApi.EnginePluginV1 = enginePlugin as engApi.EnginePluginV1;
 
         for (const engineName of getAvailableEngineNamesFromPlugin(enginePluginV1)) {
-            const engConf: engApi.ConfigObject = this.config.getEngineSettingsFor(engineName);
+            const engConf: engApi.ConfigObject = this.config.getEngineConfigFor(engineName);
             const engine: engApi.Engine = createEngineFromPlugin(enginePluginV1, engineName, engConf);
             this.addEngineIfValid(engineName, engine);
         }
@@ -93,11 +93,19 @@ export class CodeAnalyzer {
     private getAllRules(): RuleImpl[] {
         const allRules: RuleImpl[] = [];
         for (const [engineName, engine] of this.engines) {
-            for (const ruleDescription of engine.describeRules()) {
+            for (let ruleDescription of engine.describeRules()) {
+                ruleDescription = this.updateRuleDescriptionWithOverrides(engineName, ruleDescription);
                 allRules.push(new RuleImpl(engineName, ruleDescription))
             }
         }
         return allRules;
+    }
+
+    private updateRuleDescriptionWithOverrides(engineName: string, ruleDescription: engApi.RuleDescription): engApi.RuleDescription {
+        const ruleOverride: RuleOverride = this.config.getRuleOverrideFor(engineName, ruleDescription.name);
+        ruleDescription.severityLevel = ruleOverride.severity || ruleDescription.severityLevel;
+        ruleDescription.tags = ruleOverride.tags || ruleDescription.tags;
+        return ruleDescription;
     }
 }
 
--- a/packages/code-analyzer-core/src/code-analyzer.ts
+++ b/packages/code-analyzer-core/src/code-analyzer.ts
@@ -4,7 +4,7 @@ import {Event, EventType, LogLevel} from "./events"
 import {getMessage} from "./messages";
 import * as engApi from "@salesforce/code-analyzer-engine-api"
 import {EventEmitter} from "node:events";
-import {CodeAnalyzerConfig, RuleOverride} from "./config";
+import {CodeAnalyzerConfig, FIELDS, RuleOverride} from "./config";
 
 export type RunOptions = {
     filesToInclude: string[]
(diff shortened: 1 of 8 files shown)


packages/code-analyzer-core/src/config.ts
```diff
@@ -83,16 +83,20 @@ export class CodeAnalyzerConfig {
...
+    public getRuleOverridesFor(engineName: string): Record<string, RuleOverride> {
         return this.config.rules[engineName] || {};
     }
 
-    public getEngineSettingsFor(engineName: string): engApi.ConfigObject {
+    public getRuleOverrideFor(engineName: string, ruleName: string): RuleOverride {
+        return this.getRuleOverridesFor(engineName)[ruleName] || {};
+    }
+
+    public getEngineConfigFor(engineName: string): engApi.ConfigObject {
```
stephen-carter-at-sf, line 94: After renaming the config to no longer have the word "settings" anywhere... I renamed this yet again which Josh called out before. So I think this is good and things are starting to stabilize and make sense here.

packages/code-analyzer-core/src/rules.ts
```diff
@@ -75,12 +75,17 @@ export class RuleImpl implements Rule {
     matchesRuleSelector(ruleSelector: string): boolean {
         const partsToMatch: string[] = ruleSelector.split(':');
         for (const selectorPart of partsToMatch) {
+            const sevNumber: number = this.getSeverityLevel().valueOf();
+            const sevName: string = SeverityLevel[sevNumber];
+
+            const selectorPartLower = selectorPart.toLowerCase();
```
jfeingold35: Nitpick: You could `toLowerCase()` the selector as a whole and call `split` on that, or do `partsToMatch: string[] = ruleSelector.split(':').map(part => part.toLowerCase())` to do it all in one go.

packages/code-analyzer-core/src/rules.ts
```diff
@@ -75,12 +75,17 @@ export class RuleImpl implements Rule {
...
-                || this.getEngineName() == selectorPart
-                || this.getName() == selectorPart
-                || this.getTags().includes(selectorPart)
-                || String(this.getSeverityLevel().valueOf()) == selectorPart;
+                selectorPartLower == "all"
+                || this.getEngineName().toLowerCase() == selectorPartLower
+                || this.getName().toLowerCase() == selectorPartLower
+                || this.getTags().map(t => t.toLowerCase()).includes(selectorPartLower)
+                || String(sevNumber) == selectorPart
+                || sevName.toLowerCase() == selectorPartLower;
```
jfeingold35: It feels like this could be simplified by putting the things you're going to `== selectorPartLower` into an array and then doing `arr.some(p => p == selectorPartLower)`.

packages/code-analyzer-core/src/rules.ts
```diff
@@ -75,12 +75,17 @@ export class RuleImpl implements Rule {
...
             const partMatched =
-                selectorPart == "all"
-                || this.getEngineName() == selectorPart
-                || this.getName() == selectorPart
-                || this.getTags().includes(selectorPart)
-                || String(this.getSeverityLevel().valueOf()) == selectorPart;
+                selectorPartLower == "all"
+                || this.getEngineName().toLowerCase() == selectorPartLower
+                || this.getName().toLowerCase() == selectorPartLower
+                || this.getTags().map(t => t.toLowerCase()).includes(selectorPartLower)
```
jfeingold35: You could simplify this by `toLowerCase`ing the tags outside of the loop.

packages/code-analyzer-core/src/code-analyzer.ts
```diff
@@ -93,12 +93,20 @@ export class CodeAnalyzer {
...
+            for (let ruleDescription of engine.describeRules()) {
+                ruleDescription = this.updateRuleDescriptionWithOverrides(engineName, ruleDescription);
                 allRules.push(new RuleImpl(engineName, ruleDescription))
             }
         }
         return allRules;
     }
+
+    private updateRuleDescriptionWithOverrides(engineName: string, ruleDescription: engApi.RuleDescription): engApi.RuleDescription {
+        const ruleOverride: RuleOverride = this.config.getRuleOverrideFor(engineName, ruleDescription.name);
```
jfeingold35, line 105: I know we don't have a logging solution in place, but as a note to myself, rule properties being overridden seems like something we'll want to log once we have one.

packages/code-analyzer-core/src/code-analyzer.ts
```diff
@@ -93,12 +93,20 @@ export class CodeAnalyzer {
...
+        ruleDescription.severityLevel = ruleOverride.severity || ruleDescription.severityLevel;
+        ruleDescription.tags = ruleOverride.tags || ruleDescription.tags;
```
jfeingold35: `[] || ['a'] == []`, so this is fine as long as we're correctly treating an absent `tags` property in the override file as equivalent to `null` instead of `[]`.

jfeingold35 (approved):

packages/code-analyzer-core/src/code-analyzer.ts
stephen-carter-at-sf: Notice the config.test.ts has a test for tags: [] to make it so that we can indeed allow users to wipe out all the tags if they want.

packages/code-analyzer-core/src/rules.ts
stephen-carter-at-sf: Updated implementation per your suggestion.

packages/code-analyzer-core/src/rules.ts
stephen-carter-at-sf: Updated implementation per your suggestion.

packages/code-analyzer-core/src/rules.ts
stephen-carter-at-sf: Updated implementation per your suggestion.

packages/code-analyzer-core/src/code-analyzer.ts
stephen-carter-at-sf: Added "Debug" level log events here. Good idea!

jfeingold35: Having one debug per property might end up being noisy in heavily customized configs, but it works for now. If it's determined to be too noisy later we can change it.

jfeingold35 (approved):

PR #15, opened 2024-05-22
NEW: @W-15806448@ Implement core run behavior

stephen-carter-at-sf:
I split this PR into 2 commits. 
* The first commit takes care of the validation of the client input to the run method and passing this off to each of the engines
* The second commit takes care of validating and transforming the results from each engine into the overall RunResults

### DIFF
--- a/packages/code-analyzer-core/src/code-analyzer.ts
+++ b/packages/code-analyzer-core/src/code-analyzer.ts
@@ -108,7 +108,9 @@ export class CodeAnalyzer {
     private getAllRules(): RuleImpl[] {
         const allRules: RuleImpl[] = [];
         for (const [engineName, engine] of this.engines) {
-            for (let ruleDescription of engine.describeRules()) {
+            const ruleDescriptions: engApi.RuleDescription[] = engine.describeRules();
+            validateRuleDescriptions(ruleDescriptions, engineName);
+            for (let ruleDescription of ruleDescriptions) {
                 ruleDescription = this.updateRuleDescriptionWithOverrides(engineName, ruleDescription);
                 allRules.push(new RuleImpl(engineName, ruleDescription))
             }
@@ -150,6 +152,16 @@ function createEngineFromPlugin(enginePlugin: engApi.EnginePluginV1, engineName:
         return enginePlugin.createEngine(engineName, config);
     } catch (err) {
         throw new Error(getMessage('PluginErrorFromCreateEngine', engineName, (err as Error).message), {cause: err})
+    }
+}
+
+function validateRuleDescriptions(ruleDescriptions: engApi.RuleDescription[], engineName: string): void {
+    const ruleNamesSeen: Set<string> = new Set();
+    for (const ruleDescription of ruleDescriptions) {
+        if (ruleNamesSeen.has(ruleDescription.name)) {
+            throw new Error(getMessage('EngineReturnedMultipleRulesWithSameName', engineName, ruleDescription.name));
+        }
+        ruleNamesSeen.add(ruleDescription.name);
     }
 }
 
--- a/packages/code-analyzer-core/src/code-analyzer.ts
+++ b/packages/code-analyzer-core/src/code-analyzer.ts
@@ -1,10 +1,12 @@
 import {RuleImpl, RuleSelection, RuleSelectionImpl} from "./rules"
-import {RunResults} from "./results"
+import {RunResults, RunResultsImpl} from "./results"
 import {Event, EventType, LogLevel} from "./events"
 import {getMessage} from "./messages";
 import * as engApi from "@salesforce/code-analyzer-engine-api"
 import {EventEmitter} from "node:events";
 import {CodeAnalyzerConfig, FIELDS, RuleOverride} from "./config";
+import {toAbsolutePath} from "./utils";
+import fs from "node:fs";
 
 export type RunOptions = {
     filesToInclude: string[]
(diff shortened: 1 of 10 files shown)


packages/code-analyzer-core/src/code-analyzer.ts
```diff
@@ -50,8 +52,21 @@ export class CodeAnalyzer {
...
+        const engineRunOptions: engApi.RunOptions = extractEngineRunOptions(runOptions);
+        this.emitLogEvent(LogLevel.Debug, getMessage('RunningWithRunOptions', JSON.stringify(engineRunOptions)));
+
+        const runResults: RunResultsImpl = new RunResultsImpl();
+        for (const engineName of ruleSelection.getEngineNames()) {
+            const rulesToRun: string[] = ruleSelection.getRulesFor(engineName).map(r => r.getName());
+            this.emitLogEvent(LogLevel.Debug, getMessage('RunningEngineWithRules', engineName, JSON.stringify(rulesToRun)));
+            const engine: engApi.Engine = this.getEngine(engineName);
+            const apiEngineRunResults: engApi.EngineRunResults = engine.runRules(rulesToRun, engineRunOptions);
+            validateEngineRunResults(engineName, apiEngineRunResults, ruleSelection);
```
stephen-carter-at-sf, line 65: Note that I currently have this implemented such it throws an error if something is wrong with what the engine returns. But we could if we wanted to, just emit a log event here with "Error" severity... and then have clients simply inform the user that something bad happened... but continue returning the results even if they are wrong.

But i'm leaning towards just erroring here so that users actually get upset and inform us if we have an engine giving back bad results. That is, I think I'd rather immediately fix an engine and deploy a fix asap (because it really shouldn't happen if we code up the engines well). But i'm open to change this if others disagree. It make sense to change it once we open up the engine api... but right now I propose we keep it as is.

packages/code-analyzer-core/src/utils.ts
```diff
@@ -0,0 +1,6 @@
+import path from "node:path";
+
+export function toAbsolutePath(fileOrFolder: string): string {
```
jfeingold35, line 3: In the current scanner, we use a library called `normalize-path` for this. If it's really just a simple one-liner then we can do it ourselves, but if not we could consider continuing to delegate it.

packages/code-analyzer-core/src/rules.ts
```diff
@@ -117,4 +119,13 @@ export class RuleSelectionImpl implements RuleSelection {
     getRulesFor(engineName: string): Rule[] {
         return this.ruleMap.get(engineName) || [];
     }
+
+    getRule(engineName: string, ruleName: string): Rule {
+        for (const rule of this.getRulesFor(engineName)) {
```
jfeingold35, line 124: I'm not sure how it would happen, but would the existence of two rules with the same name be an error case as well?

packages/code-analyzer-core/src/results.ts
```diff
@@ -1,22 +1,26 @@
...
-    getEndLine(): number | null
-    getEndColumn(): number | null
+    getEndLine(): number | undefined
+    getEndColumn(): number | undefined
 }
 
 export interface Violation {
-    getRule(): Rule,
-    getMessage(): string,
+    getRule(): Rule
```
jfeingold35, line 15: What was the impact of the commas?

packages/code-analyzer-core/src/code-analyzer.ts
```diff
@@ -131,4 +151,154 @@ function createEngineFromPlugin(enginePlugin: engApi.EnginePluginV1, engineName:
     } catch (err) {
         throw new Error(getMessage('PluginErrorFromCreateEngine', engineName, (err as Error).message), {cause: err})
     }
+}
+
+function extractEngineRunOptions(runOptions: RunOptions): engApi.RunOptions {
```
jfeingold35, line 168: I'd like to see some documentation for this method and the other validation methods. At the very least, a JSDoc header would go a long way.

packages/code-analyzer-core/src/rules.ts
stephen-carter-at-sf: Yeah good question. I'll add in a validation step when we retrieve the rules from the engine via the describeRules method to ensure that they are unique.

packages/code-analyzer-core/src/results.ts
stephen-carter-at-sf: No impact... this was just stylistic and inconsistent with the rest of the code base so I was just cleaning this up.

packages/code-analyzer-core/src/code-analyzer.ts
stephen-carter-at-sf: I have scheduled to add in doc comments soon as a separate polishing work item. I mentioned this in one of my earlier reviews... but i'm waiting for things to stabilize and finalize before doing so to avoid having to maintain things while things are still changing a bit.

packages/code-analyzer-core/src/utils.ts
stephen-carter-at-sf: As you've probably noticed... I'm trying my best to avoid bringing in unnecessary dependencies that we then need to maintain. Also, normalize-path (https://github.com/jonschlinkert/normalize-path/blob/master/index.js) actually does something different from what I'm doing here.

First, normalize-path always makes the path have forward slashes... which doesn't make sense for windows operating systems that have things like C:\Users\Some\Folder (causing the path to say something like C:/Users/Some/Folder which is invalid). 

Second: what my code does is actually fix relative paths to use the correct slash according to the operating system in which it runs and then converts it to an absolute path if it isn't already one.

You'll notice that in some of my test examples I use different slashes to cover it fixing the slashes as needed and that I run my tests both on windows and linux to confirm that things are working.

packages/code-analyzer-core/src/rules.ts
stephen-carter-at-sf: Added a commit to this PR to add in this validation.

jfeingold35 (approved):

PR #16, opened 2024-05-28
NEW: @W-15806455@: Implement listeners for the core run method

stephen-carter-at-sf:


packages/code-analyzer-core/src/index.ts
```diff
@@ -8,6 +8,7 @@ export {
 } from "./code-analyzer"
 
 export {
+    EngineLogEvent,
```
stephen-carter-at-sf, line 11: I decided to add in a separate EngineLogEvent instead of just reusing the LogEvent. This way the client can differentiate where the logs came from: Core (via LogEvent) or a specific engine (via EngineLogEvent).

packages/code-analyzer-core/src/events.ts
```diff
@@ -21,6 +22,14 @@ export type LogEvent = {
...
 
+export type EngineLogEvent = {
+    type: EventType.EngineLogEvent,
+    timestamp: Date,
+    engineName: string
+    logLevel: LogLevel,
+    message: string
+}
+
 export type EngineProgressEvent = {
```
stephen-carter-at-sf, line 33: We might decide to add in an updateMessage field. But I don't know if we need it yet, so I'm waiting to see how things evolve before doing so.

jag-j: How is it different from message?

stephen-carter-at-sf: There is no message field on EngineProgressEvent right now. That's my point. We may or may not need any message sent when updating progress information from an engine. The client might just use the Info level EngineLogEvent and use that only. We'll see.

stephen-carter-at-sf:
There is also a question of whether or not we want core to populate the 0 and 100 percentComplete events just in case the engine fails to do so. What do you all think?

Doing so would mean that the client can always rely on these events existing (sort of like the EngineResultsEvent). But it may also mean that engines may also provide their own 0 and 100 thus doubling these events. I'm on the fence since we will be the ones who currently own the engines and have control over which events will fire.

Update: After chatting with some folks, decided to update to have core always push the 0 and 100 percentComplete values even though engines might also do so.

packages/code-analyzer-core/test/stubs.ts
```diff
@@ -99,7 +100,24 @@ export class StubEngine1 extends engApi.Engine {
     }
 
     runRules(ruleNames: string[], runOptions: engApi.RunOptions): engApi.EngineRunResults {
+        this.emitEvent<engApi.ProgressEvent>({
```
jfeingold35, line 104: Question: Are emitted events guaranteed to arrive in the same order as they were emitted? If not, then do we need handling for when the 50% event arrives before the 25% event? Do we need that handling anyway to account for the possibility of a developer error?

stephen-carter-at-sf: Yes listeners fire synchronously when the event happens. So as long as an engine reports things in the correct order then all is well. But if the client is worried about this (when we allow for external engine authors), then the client can just keep track of the last known percentage and throw away any event with a percentage less than the largest known one.

stephen-carter-at-sf: In the run.test.ts file I am now no longer using toContainEqual, but I am doing the toEqual to assert order of events now.

jfeingold35 (approved):

PR #18, opened 2024-06-03
NEW: @W-15806472@: Add in CSV, JSON, and XML result formats

stephen-carter-at-sf:
I did a few things in this PR:
* Removed eslint 9 in favor of eslint 8.57.0
* Moved to use "Recommended" instead of default for the default tag to be selected.
* Updated the RuleTypes to be Standard, DataFlow, Flow, and UnexpectedError for now
* Added in engine run exception handling to report exception as UnexpectedError violation.
* Added in CSV, JSON, and XML output formats.

### DIFF
--- a/packages/code-analyzer-core/package.json
+++ b/packages/code-analyzer-core/package.json
@@ -16,7 +16,9 @@
     "@salesforce/code-analyzer-engine-api": "0.1.2",
     "@types/js-yaml": "^4.0.9",
     "@types/node": "^20.0.0",
-    "js-yaml": "^4.1.0"
+    "csv-stringify": "^6.5.0",
+    "js-yaml": "^4.1.0",
+    "xmlbuilder": "^15.1.1"
   },
   "devDependencies": {
     "@eslint/js": "^8.57.0",
--- a/packages/code-analyzer-core/package.json
+++ b/packages/code-analyzer-core/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@salesforce/code-analyzer-core",
   "description": "Core Package for the Salesforce Code Analyzer",
-  "version": "0.1.2",
+  "version": "0.2.0",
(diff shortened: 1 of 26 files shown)


packages/code-analyzer-core/package.json
```diff
@@ -13,13 +13,15 @@
   "main": "dist/index.js",
   "types": "dist/index.d.ts",
   "dependencies": {
-    "@salesforce/code-analyzer-engine-api": "0.1.2",
+    "@salesforce/code-analyzer-engine-api": "0.2.0",
     "@types/js-yaml": "^4.0.9",
     "@types/node": "^20.0.0",
-    "js-yaml": "^4.1.0"
+    "csv-stringify": "^6.5.0",
```
stephen-carter-at-sf, line 19: I did confirm that these third party libraries have already been approved for other teams. So we are safe to use them.

packages/code-analyzer-core/package.json
```diff
@@ -13,13 +13,15 @@
...
+    "js-yaml": "^4.1.0",
+    "xmlbuilder": "^15.1.1"
   },
   "devDependencies": {
-    "@eslint/js": "^9.2.0",
+    "@eslint/js": "^8.57.0",
```
jfeingold35, line 24: Question: If you're downgrading ESLint, do you need to switch from `typescript-eslint` to `@typescript-eslint/eslint-plugin` and `@typescript-eslint/parser`?

jfeingold35 (approved):
Looks good to me.

packages/code-analyzer-core/package.json
stephen-carter-at-sf: No... they are rather independent. In fact the 9.2 wasn't actually being used. 8.57 was being used anyway since it was pulled in from another dependency.

PR #20, opened 2024-06-06
NEW: @W-15951478@: Allow resourceUrls to be provided by engine from the violation and the rule

stephen-carter-at-sf:


packages/code-analyzer-core/src/output-format.ts
```diff
@@ -191,7 +191,7 @@ function createViolationOutput(id: number, violation: Violation, runDir: string)
         endColumn: primaryLocation.getEndColumn(),
         pathLocations: [RuleType.DataFlow, RuleType.Flow].includes(rule.getType()) ? createPathLocations(codeLocations, runDir) : undefined,
         message: violation.getMessage(),
-        resources: violation.getRule().getResourceUrls()
+        resources: violation.getResourceUrls()
```
jag-j, line 194: Question - I understand you are starting off with retirejs. But other engines have resource urls at the rule level right? Would it make sense to make this a either or operation right from the start? (i.e. pick urls from rule or violations, wherever they are available)

packages/code-analyzer-core/src/results.ts
```diff
@@ -143,6 +150,10 @@ export class UnexpectedEngineErrorViolation implements Violation {
     getPrimaryLocationIndex(): number {
         return 0;
     }
+
+    getResourceUrls(): string[] {
+        return [];
```
jfeingold35, line 156: Maybe it makes sense for this to be one our help pages?

packages/code-analyzer-core/src/results.ts
```diff
@@ -115,6 +116,12 @@ export class ViolationImpl implements Violation {
     getPrimaryLocationIndex(): number {
         return this.apiViolation.primaryLocationIndex;
     }
+
+    getResourceUrls(): string[] {
+        const urls: string[] = this.rule.getResourceUrls();
+        return !this.apiViolation.resourceUrls ? urls :
```
jfeingold35, line 123: I had to turn this ternary over in my head a few times before I understood what it was doing. Could you add a 1-line comment explaining that this is combining the urls at the violation- and rule-level while skipping duplicates?

packages/code-analyzer-core/src/results.ts
stephen-carter-at-sf: It depends on the engine and the situation. But core is agnostic to engines. So i'm not sure if we'll be using this resource urls right now... but maybe in the future we might.

packages/code-analyzer-core/src/output-format.ts
stephen-carter-at-sf: On the core side, a violation object contains a rule. So the getResourceUrls method on our ViolationImpl actually is indeed putting the rule urls first. See line 120 of packages/code-analyzer-core/src/results.ts.

packages/code-analyzer-core/src/results.ts
stephen-carter-at-sf: Sure. Done.

jag-j (approved):

PR #22, opened 2024-06-11
NEW: @W-15983430@ : Update engine api's describeRules and runRules to be async. Also a few enhancements.

stephen-carter-at-sf:
In this PR:
* I updated the engine api's describeRules and runRules to be async
* I updated core to use async as well where needed
* I added in some additional file processing to ensure that we have a reduced non-redundant set of files.
* Lastly, I added in a custom_engine_plugin_modules field to our config that we can use internally for demo purposes.

### DIFF
--- a/packages/code-analyzer-core/src/code-analyzer.ts
+++ b/packages/code-analyzer-core/src/code-analyzer.ts
@@ -254,7 +254,7 @@ function extractEngineRunOptions(runOptions: RunOptions): engApi.RunOptions {
 
 function removeRedundantPaths(absolutePaths: string[]): string[] {
     // If a user supplies a parent folder and subfolder of file underneath the parent folder, then we can safely
-    // remove that subfolder or file. So
+    // remove that subfolder or file. Also, if we find duplicate entries, we remove those as well.
     const pathsSortedByLength: string[] = absolutePaths.sort((a, b) => a.length - b.length);
     const filteredPaths: string[] = [];
     for (const currentPath of pathsSortedByLength) {
--- a/packages/code-analyzer-core/src/code-analyzer.ts
+++ b/packages/code-analyzer-core/src/code-analyzer.ts
@@ -13,6 +13,7 @@ import {EventEmitter} from "node:events";
 import {CodeAnalyzerConfig, FIELDS, RuleOverride} from "./config";
 import {Clock, RealClock, toAbsolutePath} from "./utils";
 import fs from "node:fs";
+import path from "node:path";
 
 export type RunOptions = {
     filesToInclude: string[]
@@ -24,6 +25,7 @@ export class CodeAnalyzer {
     private clock: Clock = new RealClock();
     private readonly eventEmitter: EventEmitter = new EventEmitter();
     private readonly engines: Map<string, engApi.Engine> = new Map();
+    private readonly allRules: RuleImpl[] = [];
 
     constructor(config: CodeAnalyzerConfig) {
         this.config = config;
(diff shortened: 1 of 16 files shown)


packages/code-analyzer-core/src/code-analyzer.ts
```diff
@@ -44,7 +46,14 @@ export class CodeAnalyzer {
         for (const engineName of getAvailableEngineNamesFromPlugin(enginePluginV1)) {
             const engConf: engApi.ConfigObject = this.config.getEngineConfigFor(engineName);
             const engine: engApi.Engine = createEngineFromPlugin(enginePluginV1, engineName, engConf);
-            this.addEngineIfValid(engineName, engine);
+            await this.addEngineIfValid(engineName, engine);
+
+            const ruleDescriptions: engApi.RuleDescription[] = await engine.describeRules();
```
stephen-carter-at-sf, line 51: Adding an engine's rules immedialy whent he engine is added instead of waiting for rule selection. This makes it so that the selectRules doesn't have to be async. Only the addEnginePlugin (which already needs to be async for validation purposes) and the run methods need to be async.

packages/code-analyzer-core/src/code-analyzer.ts
```diff
@@ -256,6 +252,22 @@ function extractEngineRunOptions(runOptions: RunOptions): engApi.RunOptions {
     return engineRunOptions;
 }
 
+function removeRedundantPaths(absolutePaths: string[]): string[] {
+    // If a user supplies a parent folder and subfolder of file underneath the parent folder, then we can safely
+    // remove that subfolder or file. So
```
jfeingold35: Just double-checking: was there supposed to be more to this comment?

packages/code-analyzer-core/src/code-analyzer.ts
```diff
@@ -256,6 +252,22 @@ function extractEngineRunOptions(runOptions: RunOptions): engApi.RunOptions {
...
+    const pathsSortedByLength: string[] = absolutePaths.sort((a, b) => a.length - b.length);
+    const filteredPaths: string[] = [];
+    for (const currentPath of pathsSortedByLength) {
+        const isAlreadyContained = filteredPaths.some(existingPath =>
+            currentPath.startsWith(existingPath + path.sep) || existingPath === currentPath
```
jfeingold35, line 262: Could casing cause unexpected results here? E.g., does `./src/beep/boop` count as redundant with `./src/BEEP`?

packages/code-analyzer-core/src/code-analyzer.ts
```diff
@@ -256,6 +252,22 @@ function extractEngineRunOptions(runOptions: RunOptions): engApi.RunOptions {
     return engineRunOptions;
 }
 
+function removeRedundantPaths(absolutePaths: string[]): string[] {
+    // If a user supplies a parent folder and subfolder of file underneath the parent folder, then we can safely
+    // remove that subfolder or file. So
+    const pathsSortedByLength: string[] = absolutePaths.sort((a, b) => a.length - b.length);
```
jfeingold35, line 258: Recommendation: Consider checking node's `path` module to see if there's already a built-in method for comparing paths. I have to imagine there is.

packages/code-analyzer-core/src/code-analyzer.ts
stephen-carter-at-sf: Linux has case sensitivity on paths in many cases. Just to be safe, I think i’ll leave things as is. Note that this is just a nice to have anyway to prevent engines from running folders/files multiple times if the user accidentally provides them multiple times.

packages/code-analyzer-core/src/code-analyzer.ts
stephen-carter-at-sf: Fixed. Thanks.

packages/code-analyzer-core/src/code-analyzer.ts
stephen-carter-at-sf: Since the paths have already been converted and resolved to absolute paths... they are already for comparison. Node only provides those transformation functions (which we are using) - but not comparison functions. So this I believe is the way to go. I tested this like crazy... so I think we are good to go.

jfeingold35: Fair enough.

jfeingold35 (approved):

PR #23, opened 2024-06-12
NEW: @W-15884203@: Implement the retire-js engine

stephen-carter-at-sf:


packages/code-analyzer-retirejs-engine/src/engine.ts
```diff
@@ -0,0 +1,131 @@
...
+    return violations;
+}
+
+function toViolation(vulnerability: Vulnerability, library: string, fileOrZipArchive: string, fileInsideZipArchive?: string) {
+    const vulnerabilityDetails: string = JSON.stringify(vulnerability.identifiers, null, 2);
+    const message: string = (fileInsideZipArchive ?
+                `'${library}' was found inside of the zipped archive in '${fileInsideZipArchive}' which contains a known vulnerability. `
+                : `'${library}' contains a known vulnerability. `
+        ) +
+        `Please upgrade to latest version.\nVulnerability details: ${vulnerabilityDetails}`;
```
stephen-carter-at-sf, line 122: As a separate work item, i'll put all of our messages in a separate message file so that we can eventually do a doc review on it.

packages/code-analyzer-retirejs-engine/package.json
```diff
@@ -14,7 +14,12 @@
   "types": "dist/index.d.ts",
   "dependencies": {
     "@salesforce/code-analyzer-engine-api": "0.3.0",
-    "@types/node": "^20.0.0"
+    "@types/node": "^20.0.0",
+    "@types/tmp": "^0.2.6",
+    "isbinaryfile": "^5.0.2",
+    "node-stream-zip": "^1.15.0",
+    "retire": "^5.0.0",
+    "tmp": "^0.2.3"
```
stephen-carter-at-sf: I confirmed that all these third party libraries (which we used before) have all already been approved.

packages/code-analyzer-retirejs-engine/src/engine.ts
```diff
@@ -0,0 +1,135 @@
...
+    private readonly retireJsExecutor: RetireJsExecutor;
+
+    constructor(retireJsExecutor?: RetireJsExecutor) {
+        super();
+        const emitLogEventFcn: EmitLogEventFcn = (logLevel: LogLevel, msg: string) => this.emitEvent<LogEvent>(
+            {type: EventType.LogEvent, logLevel: logLevel, message: msg});
+        this.retireJsExecutor = retireJsExecutor ? retireJsExecutor : new AdvancedRetireJsExecutor(emitLogEventFcn);
+    }
+
+    getName(): string {
```
jag-j, line 57: Some places you call this method and in other places, you directly use the RetireJsEngine.NAME. Is that intentional?

packages/code-analyzer-retirejs-engine/src/engine.ts
```diff
@@ -0,0 +1,135 @@
...
+    LogEvent,
+    LogLevel,
+    RuleDescription,
+    RuleType,
+    RunOptions,
+    SeverityLevel,
+    Violation
+} from "@salesforce/code-analyzer-engine-api";
+import {RetireJsExecutor, AdvancedRetireJsExecutor, ZIPPED_FILE_MARKER, EmitLogEventFcn} from "./executor";
+import {Finding, Vulnerability} from "retire/lib/types";
```
jfeingold35, line 16: I have reservations about importing from within a module like this. RetireJS has actually burnt us once before by relocating a file. This one seems reasonable, though, I think.

packages/code-analyzer-retirejs-engine/src/utils.ts
```diff
@@ -0,0 +1,79 @@
...
+ * @param parentTempDir - if supplied, then a temporary folder is placed directly underneath this parent folder.
+ */
+export async function createTempDir(parentTempDir?: string) : Promise<string> {
+    return tmpDirAsync({dir: parentTempDir, keep: false, unsafeCleanup: true});
+}
+
+/**
+ * Expands a list of files and/or folders to be a list of all contained files, including the files found in subfolders
+ */
+export function expandToListAllFiles(absoluteFileOrFolderPaths: string[]): string[] {
```
jfeingold35, line 22: Question: What does this method do that using `globby` and `glob('**/**')` wouldn't do?

packages/code-analyzer-retirejs-engine/src/utils.ts
```diff
@@ -0,0 +1,79 @@
...
+/**
+ * Determines if the provided file is a compressed zip file based on its contents
+ *
+ * Note that this code is directly from https://github.com/kevva/is-zip/blob/master/index.js
+ * The is-zip module does not come with any typescript type information and thus cannot be used without "require".
+ * But since it literally just contains the following code, it is simple enough to just copy the function here.
+ */
+export function isZipFile(file: string) {
+    const buf: Buffer = fs.readFileSync(file);
+    /* istanbul ignore next */
```
jfeingold35, line 65: Why is this being excluded?

packages/code-analyzer-retirejs-engine/test/engine.test.ts
```diff
@@ -0,0 +1,214 @@
...
+    beforeEach(() => {
+        engine = new RetireJsEngine(new StubRetireJsExecutor());
+    });
+
+    it('When getName is called, then retire-js is returned', () => {
+        expect(engine.getName()).toEqual('retire-js');
+    });
+
+    it('When validate is called, then nothing happens since it currently is a no-op', () => {
+        engine.validate();
```
jfeingold35: Is this test actually verifying that nothing is happening? Or is it just for coverage?

packages/code-analyzer-retirejs-engine/test/engine.test.ts
```diff
@@ -0,0 +1,214 @@
...
+            severityLevel: SeverityLevel.Low,
+            type: RuleType.Standard,
+            tags: ['Recommended'],
+            description: getMessage('RetireJsRuleDescription', 'low'),
+            resourceUrls: ['https://retirejs.github.io/retire.js/']
+        });
+    });
+
+    it('When runRules is called, then the RetireJsExecutor is called with the correct inputs', async () => {
+        const spyExecutor: SpyRetireJsExecutor = new SpyRetireJsExecutor();
```
jfeingold35, line 154: Nitpick: A comment clarifying that this is replacing the engine that's initialized in the `before` with one that uses a spy instead of a stub would be helpful.

packages/code-analyzer-retirejs-engine/test/executor.test.ts
```diff
@@ -0,0 +1,241 @@
...
+            file: path.resolve('test', 'test-data', 'scenarios', '7_hasZipFolderWithVulnFileInChildFolders', 'ZipWithDirectories.zip') + ZIPPED_FILE_MARKER + 'FilledParentFolder/ChildFolderWithText/JsFileWithoutExt',
+            results: getExpectedJQueryResults("filecontent")
+        });
+        expect(findings).toContainEqual({
+            file: path.resolve('test', 'test-data', 'scenarios', '7_hasZipFolderWithVulnFileInChildFolders', 'ZipWithDirectories.zip') + ZIPPED_FILE_MARKER + 'FilledParentFolder/ChildFolderWithText/JsFileWithOddExt.foo',
+            results: getExpectedJQueryResults("filecontent")
+        });
+    });
+
+    it('When files are specified among input paths, then AdvancedRetireJsExecutor knows to target that scan those files without its peers', async () => {
```
jfeingold35: This sentence is malformed.

packages/code-analyzer-retirejs-engine/src/executor.ts
```diff
@@ -0,0 +1,171 @@
...
+            throw new Error(getMessage('UnexpectedErrorWhenProcessingOutputFile', tempOutputFile, (err as Error).message),
+                {cause: err});
+        }
+    }
+}
+
+/**
+ * AdvancedRetireJsExecutor - An advanced strategy of calling retire that does a lot more work to catch more vulnerabilities
+ *
+ * This executor examines all specifies files and makes a copy of all text based files into a temporary folder, renaming
```
jag-j, line 95: Is this something we introduced in 5.x or is this the same strategy currently followed by sfdx-scanner?

packages/code-analyzer-retirejs-engine/src/engine.ts
stephen-carter-at-sf: Yes I saw that history. But I'd prefer not to reimplement all their types... so I think I'm good with this. If they move things around on us, then our build will fail (only if we grab a later minor change from them). But at that point... we can simply not ship it and revert to the older version if needed... at which point we will wag our finger at them saying they made a breaking change in a minor release.

packages/code-analyzer-retirejs-engine/src/engine.ts
stephen-carter-at-sf: getName is the only way the core module gets the engine name. The only other place I call getName here in this project is when testing that getName returns the expected name. Otherwise, I only use the constant. This indeed is intentional. In those other instances where I use the constant, I don't have an instance of the engine in order to call getName anyway.

packages/code-analyzer-retirejs-engine/src/executor.ts
stephen-carter-at-sf: No this is exactly what our current strategy with 4.x is. This is what makes our retire-js engine more complicated... because we have to do all this advanced stuff.

packages/code-analyzer-retirejs-engine/src/utils.ts
stephen-carter-at-sf: This function receives the list of files and folders that it gets from core. For example:
if the user passes to the cli ['some/relativeFile.js', 'some/relative/folder', 'some/relative/folder/someFileInTheFolder.js'] then core automatically removes dudundant entries, and makes the paths absolute:
['/full/path/some/relativeFile.js', '/full/path/some/relative/folder']

and then the engine gets this list.

Now retire js needs to walk through the list of all files... so this expandToListAllFiles will then convert things to

['/full/path/some/relativeFile.js', '/full/path/some/relative/folder/subFolder/someFile.js', '/full/path/some/relative/folder/anotherFile.html', ...., '/full/path/some/relative/folder/yetAnotherFile.zip']

for example.

globby just expands globs. Also, i'd like to not bring in unnecessary tools to do operations that are simple. Using a hammer all the time can get messy.

The algorithm I'm using here is a very common path algorithm for finding all files under a directory.

packages/code-analyzer-retirejs-engine/src/utils.ts
stephen-carter-at-sf: Any time we have literally no way to hit a branch of code because of some crazy edge case, it's nice to make this known to the code coverage utility.

/* istanbul ignore next */
is how you basically say, ignore the next line for code coverage checking here.

It seems there are many edge cases that is-zip covers... but our zip files only hit a few of these branches. But it leaving code as is since our users might have some random file that hits these other branches.

packages/code-analyzer-retirejs-engine/test/engine.test.ts
stephen-carter-at-sf: Our validate method here is a no-op. Making sure that at least the core module can call it without things blowing up. So it is both for coverage, but also as a sanity check that when the core module calls it - it won't break anything.

packages/code-analyzer-retirejs-engine/test/engine.test.ts
stephen-carter-at-sf: Sure - I added a comment per your suggestion.

packages/code-analyzer-retirejs-engine/test/executor.test.ts
stephen-carter-at-sf: Fixed. Thanks.

packages/code-analyzer-retirejs-engine/src/engine.ts
jfeingold35: Fair enough.

jfeingold35 (approved):

PR #27, opened 2024-06-15
CHANGE(retire-js): @W-15884203@: Restrict files to target with retire-js engine and improve performance

stephen-carter-at-sf:
This change brings us in line with exactly how our old engine worked over athttps://github.com/forcedotcom/sfdx-scanner/blob/dev/src/lib/retire-js/RetireJsEngine.ts by:
* restricting target files to be scanned the same way we were doing so before
* more efficiently creating copy of files

This dramatically improved the performance... in fact it is now slightly faster than our old engine.

### DIFF
--- a/packages/code-analyzer-retirejs-engine/src/executor.ts
+++ b/packages/code-analyzer-retirejs-engine/src/executor.ts
@@ -18,6 +18,8 @@ const RETIRE_JS_VULN_REPO_FILE: string = path.resolve(__dirname, '..', 'vulnerab
 
 const RETIRE_COMMAND: string = utils.findCommand('retire');
 
+const JS_EXTENSIONS = ['.js', '.mjs', '.cjs'];
+
 export interface RetireJsExecutor {
     execute(filesAndFoldersToScan: string[]): Promise<Finding[]>
 }
@@ -63,8 +65,14 @@ export class SimpleRetireJsExecutor implements RetireJsExecutor {
     }
 
     private async scanFolder(folder: string): Promise<Finding[]> {
-        const tempOutputFile: string = path.resolve(await utils.createTempDir(), "output.json");
-        const command: string = `${RETIRE_COMMAND} --path "${folder}" --exitwith 13 --outputformat jsonsimple --outputpath "${tempOutputFile}" --jsrepo "${RETIRE_JS_VULN_REPO_FILE}"`;
+        const tempOutputFile: string = (await utils.createTempDir()) + path.sep + 'output.json';
+        const command: string = RETIRE_COMMAND +
+            ` --path "${folder}"` +
+            ` --exitwith 13` +
+            ` --outputformat jsonsimple` +
+            ` --outputpath "${tempOutputFile}"` +
+            ` --jsrepo "${RETIRE_JS_VULN_REPO_FILE}"` +
+            ` --ext "${JS_EXTENSIONS.map(ext => ext.replace('.','')).join(',')}"`;
 
         this.emitLogEvent(LogLevel.Fine, `Executing command: ${command}`);
         try {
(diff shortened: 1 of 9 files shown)


packages/code-analyzer-retirejs-engine/src/executor.ts
```diff
@@ -106,68 +106,140 @@ export class AdvancedRetireJsExecutor implements RetireJsExecutor {
...
     // <zip_file>::[ZIPPED_FILE]::<embedded_file>
-    const zip: StreamZip.StreamZipAsync = new StreamZip.async({file: zipFile, storeEntries: true});
-    const entries: { [name: string]: StreamZip.ZipEntry } = await zip.entries();
+    private async processZipFile(zipFile: string): Promise<void> {
+        const zip: StreamZip.StreamZipAsync = new StreamZip.async({file: zipFile, storeEntries: true});
+        const entries: { [name: string]: StreamZip.ZipEntry } = await zip.entries();
+        for (const entry of Object.values(entries)) {
+            if (entry.isDirectory || !utils.isTextFile(await zip.entryData(entry.name))) {
+                continue; // Skip directories and non-text files.
+            }
```
stephen-carter-at-sf, line 203: I'm doing exactly what the old engine did. But I wonder, why aren't we filtering down the entry list of zip folders further the same way we target non-zipped files. I noticed we don't exclude files underneath node_modules folders for example... or non-js files that do not have a corresponding *.resource-meta.xml file? My only concern here is that if a project has a lot of zip folders, then we are extracting and scanning all the text files from inside all the zip folders. But if this is the way it was before, then we'll keep it as is for now.

packages/code-analyzer-retirejs-engine/src/executor.ts
```diff
@@ -63,7 +63,7 @@ export class SimpleRetireJsExecutor implements RetireJsExecutor {
     }
 
     private async scanFolder(folder: string): Promise<Finding[]> {
-        const tempOutputFile: string = path.resolve(await utils.createTempDir(), "output.json");
+        const tempOutputFile: string = (await utils.createTempDir()) + path.sep + 'output.json';
```
stephen-carter-at-sf: We get little performance boosts by just concatenating instead of doing a full resolve unnecessarily. Note the tempDir is already resolved.

packages/code-analyzer-retirejs-engine/src/executor.ts
```diff
@@ -106,68 +106,140 @@ export class AdvancedRetireJsExecutor implements RetireJsExecutor {
...
+    return files.filter(file => shouldTarget(file, filesSet));
+}
+function shouldTarget(file: string, filesSet: Set<string>): boolean {
+    const fileInfo: path.ParsedPath = path.parse(file);
+    if (fileIsInFolderToSkip(file)) {
+        return false;
+    } else if (EXTENSIONS_TO_TARGET.includes(fileInfo.ext.toLowerCase())) {
+        return true;
+    }
+    return filesSet.has(`${fileInfo.dir}${path.sep}${fileInfo.name}.resource-meta.xml`);
```
stephen-carter-at-sf, line 244: Since we have the original file list, we don't need to do the expensive operating system file existence checks. Instead we get to simply look at the list of files in the workspace. Another big performance boost!

packages/code-analyzer-retirejs-engine/src/executor.ts
```diff
@@ -106,68 +114,140 @@ export class AdvancedRetireJsExecutor implements RetireJsExecutor {
...
         return findings;
     }
-}
 
-async function processFile(file: string, tmpDir: string, fileMap: Map<string, string>): Promise<void> {
-    if (file.toLowerCase().endsWith(".js") || utils.isTextFile(file)) {
-        return processTextFile(file, tmpDir, fileMap);
-    } else if (utils.isZipFile(file)) {
-        return processZipFile(file, tmpDir, fileMap);
+    // Create parent temporary directory (that cleans up after itself when process exits) and add subdirectories under
```
jfeingold35: This should be JSDoc style.

packages/code-analyzer-retirejs-engine/src/executor.ts
```diff
@@ -106,68 +114,140 @@ export class AdvancedRetireJsExecutor implements RetireJsExecutor {
...
-    // whenever possible we preserve the original name of the file by placing the file with the same name
-    // (but with a .js) extension inside a sub folder in the temporary directory. The sub folder helps avoid with
-    // possible name collisions. For performance, we avoid synchronous calls which block the main thread.
-    const fileNameWithJsExt: string = path.basename(file, path.extname(file)) + '.js';
-    const tmpSubFolder: string = await utils.createTempDir(tmpDir);
-    const linkFile: string = path.resolve(tmpSubFolder, fileNameWithJsExt);
-    fileMap.set(linkFile, file);
-    return utils.linkOrCopy(file, linkFile);
-}
+    // Make a symlink of the text file into its associated temporary directory with a .js file extension and update
```
jfeingold35: JSDoc.

packages/code-analyzer-retirejs-engine/src/executor.ts
```diff
@@ -106,68 +114,140 @@ export class AdvancedRetireJsExecutor implements RetireJsExecutor {
...
+}
 
-    for (const entry of Object.values(entries)) {
-        if (entry.isDirectory || !utils.isTextFile(await zip.entryData(entry.name))) {
-            continue; // Skip directories and non-text files.
+// If we do not reduce the file list down, then there could be a massive amount of files to process resulting
+// in a rather slow engine. Therefore, we reduce the files with a targeted extension or files that have an associated
+// salesforce metadata file while skipping files that live in certain folders we want to skip.
+const FOLDERS_TO_SKIP = ['node_modules', 'bower_components'];
+const EXTENSIONS_TO_TARGET = [...JS_EXTENSIONS, '.resource', '.zip'];
```
jfeingold35: I feel like these should be defined higher up, with `JS_EXTENSIONS`.

packages/code-analyzer-retirejs-engine/src/executor.ts
```diff
@@ -18,6 +18,8 @@ const RETIRE_JS_VULN_REPO_FILE: string = path.resolve(__dirname, '..', 'vulnerab
 
 const RETIRE_COMMAND: string = utils.findCommand('retire');
 
+const JS_EXTENSIONS = ['.js', '.mjs', '.cjs'];
```
jfeingold35, line 19: IMO, this, `FOLDERS_TO_SKIP`, and `EXTENSIONS_TO_TARGET` should be defined together, and each should have a JSDoc comment explaining why its values are what they are.

packages/code-analyzer-retirejs-engine/src/executor.ts
```diff
@@ -106,68 +114,140 @@ export class AdvancedRetireJsExecutor implements RetireJsExecutor {
...
+    const fileInfo: path.ParsedPath = path.parse(file);
+    if (fileIsInFolderToSkip(file)) {
+        return false;
+    } else if (EXTENSIONS_TO_TARGET.includes(fileInfo.ext.toLowerCase())) {
+        return true;
+    }
+    return filesSet.has(`${fileInfo.dir}${path.sep}${fileInfo.name}.resource-meta.xml`);
+}
+function fileIsInFolderToSkip(file: string): boolean {
+    for (const folderToSkip of FOLDERS_TO_SKIP) {
```
jfeingold35: I think this could be simplified to:
```
file.split(path.sep).some(segment => FOLDERS_TO_SKIP.includes(segment))
```

packages/code-analyzer-retirejs-engine/src/utils.ts
```diff
@@ -23,7 +23,7 @@ export function expandToListAllFiles(absoluteFileOrFolderPaths: string[]): strin
     let allFiles: string[] = [];
     for (const fileOrFolder of absoluteFileOrFolderPaths) {
         if (fs.statSync(fileOrFolder).isDirectory()) {
-            const absSubPaths: string[] = fs.readdirSync(fileOrFolder).map(f => path.resolve(fileOrFolder, f));
+            const absSubPaths: string[] = fs.readdirSync(fileOrFolder).map(f => fileOrFolder + path.sep + f);
```
jfeingold35, line 26: What motivated this change?

packages/code-analyzer-retirejs-engine/src/executor.ts
jfeingold35: This is assuming that the workspace is always a whole folder, right? In the future, we might re-add glob support, which could violate that assumption. Therefore we should probably document that assumption in the code here.

packages/code-analyzer-retirejs-engine/src/executor.ts
```diff
@@ -106,68 +114,140 @@ export class AdvancedRetireJsExecutor implements RetireJsExecutor {
     private readonly simpleExecutor: RetireJsExecutor;
     private readonly emitLogEvent: EmitLogEventFcn;
 
+    private readonly origToTempDirMap: Map<string, string> = new Map();
```
jfeingold35: I know what these maps are for, but you might consider adding JSDoc comments to them to describe their purpose.

packages/code-analyzer-retirejs-engine/src/executor.ts
```diff
@@ -106,68 +114,140 @@ export class AdvancedRetireJsExecutor implements RetireJsExecutor {
...
-    if (file.toLowerCase().endsWith(".js") || utils.isTextFile(file)) {
-        return processTextFile(file, tmpDir, fileMap);
-    } else if (utils.isZipFile(file)) {
-        return processZipFile(file, tmpDir, fileMap);
+    // Create parent temporary directory (that cleans up after itself when process exits) and add subdirectories under
+    // the parent for each of the unique folders containing text files
+    private async prepareTempDirs(textFiles: string[]): Promise<void[]> {
+        this.origToTempDirMap.clear();
+        this.tempToOrigFileMap.clear();
+        this.uniqNameCounter = 0;
```
jfeingold35, line 165: Since this counter is shared for both files and folders, there will be situations where the folders are number 1, 3, and 7, and the files are 2, 4, 5, and 6. If we ever have to manually debug, that could cause confusion. So maybe add a comment clarifying that?

packages/code-analyzer-retirejs-engine/src/executor.ts
stephen-carter-at-sf: The contract is that core will hand off to engines only absolute paths. The client must handle globs and core will resolve to absolute.

packages/code-analyzer-retirejs-engine/src/utils.ts
stephen-carter-at-sf: path.resolve actually does a lot more than we need. Concatenating is actually much faster when we know the folder is already resolved. So to answer your question - it is a simple optimization to gain yet more speed.

packages/code-analyzer-retirejs-engine/src/executor.ts
stephen-carter-at-sf: True... the counter isn't for humans... it is just to ensure uniqueness. But I'll  add a comment.

packages/code-analyzer-retirejs-engine/src/executor.ts
stephen-carter-at-sf: Done.

packages/code-analyzer-retirejs-engine/src/executor.ts
stephen-carter-at-sf: Done.

packages/code-analyzer-retirejs-engine/src/executor.ts
stephen-carter-at-sf: Done.

packages/code-analyzer-retirejs-engine/src/executor.ts
stephen-carter-at-sf: Done.

packages/code-analyzer-retirejs-engine/src/executor.ts
stephen-carter-at-sf: Done.

packages/code-analyzer-retirejs-engine/src/executor.ts
stephen-carter-at-sf: Done.

jfeingold35 (approved):

PR #29, opened 2024-06-17
NEW (RegexEngine): @W-15973916@: Scaffolding for new regex engine

ravipanguluri:
In this PR:

* Created project scaffolding to set up the new code-analyzer-regex-engine package
* Created a new RegexPluginEngine that returns a RegexEngine
* runRules() is a no-op


package.json
```diff
@@ -37,5 +38,8 @@
       "src/**/*.ts",
       "!src/index.ts"
     ]
+  },
+  "dependencies": {
+    "@salesforce/code-analyzer-engine-api": "^0.2.0"
   }
-}
\ No newline at end of file
+}
```
stephen-carter-at-sf: This is the top level package.json file... which doesn't need to be modified.

packages/code-analyzer-regex-engine/package.json
```diff
@@ -0,0 +1,63 @@
...
+  "homepage": "https://developer.salesforce.com/docs/platform/salesforce-code-analyzer/overview",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/forcedotcom/code-analyzer-core.git",
+    "directory": "packages/code-analyzer-core"
+  },
+  "main": "dist/index.js",
+  "types": "dist/index.d.ts",
+  "dependencies": {
+    "@salesforce/code-analyzer-engine-api": "^0.1.0",
```
stephen-carter-at-sf: This is the wrong version. You should be matching the version of the engine-api that is in the repo. This is why your package-lock.json file also got updated. Please revert your package-lock.json file changes first. And then Have this package.json file match that as shown over at The version here needs to match the version in the monorepo. The code-analyzer-engine-api's package.json file shows it is at version 0.4.0.

Also I think you are missing @types/node dependency.

Please make this file match that over at https://github.com/forcedotcom/code-analyzer-core/blob/dev/packages/T-E-M-P-L-A-T-E/package.json

Specifically, the engine-api version you should be using here is "^0.4.0".

packages/code-analyzer-regex-engine/src/RegexEnginePlugin.ts
```diff
@@ -0,0 +1,69 @@
+import * as EngineApi from '@salesforce/code-analyzer-engine-api/';
+import { EngineRunResults } from '@salesforce/code-analyzer-engine-api/';
```
stephen-carter-at-sf: Either use one or the other.

That is, either import each thing you want individually from the engine-api (which I prefer for readability sakes... since there are no conflicts) or just have import * as engineApi only and use engineApi.<...> everywhere.

packages/code-analyzer-regex-engine/src/RegexEnginePlugin.ts
```diff
@@ -0,0 +1,69 @@
...
+
+
+
+export class RegexEnginePlugin extends EngineApi.EnginePluginV1 {
+	private readonly createdEngines: Map<string, EngineApi.Engine> = new Map();
```
stephen-carter-at-sf: This is a premature optimization. Since the core module only calls createEngine ones in your case (because 'regex' is all that is supplied)... catching is unnecessary.

If you are caching for a different purpose - like just to help with testing. Then I would recommend you add caching on the test side only and not on the source side. No need for unnecessary code here.

packages/code-analyzer-regex-engine/src/RegexEnginePlugin.ts
```diff
@@ -0,0 +1,69 @@
...
+		}
+		return this.getCreatedEngine(engineName);
+	}
+
+	getCreatedEngine(engineName: string): EngineApi.Engine {
+		if (this.createdEngines.has(engineName)) {
+			return this.createdEngines.get(engineName) as EngineApi.Engine;
+		}
+		throw new Error(`Engine with name ${engineName} has not yet been created`);
+	}
```
stephen-carter-at-sf: What is this for? If it is just for testing... then move this to the testing side. I have a few ideas how you can do this on the test side only if you need help.

packages/code-analyzer-regex-engine/src/RegexEnginePlugin.ts
```diff
@@ -0,0 +1,69 @@
...
+}
+
+export class RegexEngine extends EngineApi.Engine {
+	readonly config: EngineApi.ConfigObject;
+	readonly runRulesCallHistory: {ruleNames: string[], runOptions: EngineApi.RunOptions}[] = [];
```
stephen-carter-at-sf: This isn't something that you should see on the source side. Maybe you were looking at the core tests with the dummy plugins and got this idea. This call history is a useful trick when creating stubs and spys for testing purposes. But you aren't creating a test double (mock, stub, or spy) here... you are creating a real engine. So no need for call history on the source side of things.

packages/code-analyzer-regex-engine/src/RegexEnginePlugin.ts
```diff
@@ -0,0 +1,69 @@
...
+	resultsToReturn: EngineApi.EngineRunResults = {
+		violations: []
```
stephen-carter-at-sf: This repo uses whilespaces instead of tabs. I currently don't have that locked down yet... but can you please use 4 spaces instead of tab characters so we can keep the repo consistent (until I lock it down at some point).

packages/code-analyzer-regex-engine/src/RegexEnginePlugin.ts
```diff
@@ -0,0 +1,69 @@
...
+
+export class RegexEnginePlugin extends EngineApi.EnginePluginV1 {
+	private readonly createdEngines: Map<string, EngineApi.Engine> = new Map();
+
+	getAvailableEngineNames(): string[] {
+		return ['regex'];
+	}
+
+	createEngine(engineName: string, config: EngineApi.ConfigObject): EngineApi.Engine {
+		if (engineName === 'regex') {
```
stephen-carter-at-sf: I see the word 'regex' here is an engine name. I would recommend making this a constant on the RegexpEngine and then reference that constant. That way, if we ever want to change the name of the engine, we can do it in one place instead of having to change it in multiple places. See the retire-js engine for example.

packages/code-analyzer-regex-engine/src/RegexEnginePlugin.ts
```diff
@@ -0,0 +1,69 @@
...
+			return this.createdEngines.get(engineName) as EngineApi.Engine;
+		}
+		throw new Error(`Engine with name ${engineName} has not yet been created`);
+	}
+}
+
+export class RegexEngine extends EngineApi.Engine {
+	readonly config: EngineApi.ConfigObject;
+	readonly runRulesCallHistory: {ruleNames: string[], runOptions: EngineApi.RunOptions}[] = [];
+	resultsToReturn: EngineApi.EngineRunResults = {
```
stephen-carter-at-sf: What is this public field for? It isn't needed. For now, since you haven't implemented the runRules method... then just have the runRules method return {violations:[]} locally instead of adding this field.

packages/code-analyzer-regex-engine/src/RegexEnginePlugin.ts
```diff
@@ -0,0 +1,69 @@
...
+
+	describeRules(): EngineApi.RuleDescription[] {
+		return [
+			{
+				name: "TrailingWhitespaceRule",
+				severityLevel: EngineApi.SeverityLevel.Low,
+				type: EngineApi.RuleType.Standard,
+				tags: ["Recommended", "CodeStyle"],
+				description: "Some description for TrailingWhiteSpaceRuleStub",
+				resourceUrls: ["https://example.com/TrailingWhiteSpaceRuleStub"]
```
stephen-carter-at-sf: I'm guessing these are just dummy values for now? If so - then it is best to leave them blank for now with a // TODO comment so that we don't accidentally leave them this way.

packages/code-analyzer-regex-engine/test/engine.test.ts
```diff
@@ -0,0 +1,99 @@
...
+            ]
+        )
+        
+    });
+
+    it('Check that any settings configurations are propogated to the engine when creating the new engine', () => {
+        const engConfig: EngineApi.ConfigObject = engine.config;
+        expect(engConfig).toEqual(config);
+        
+    });
```
stephen-carter-at-sf: Do you even need config right now? If not, then why lock this in with a test? Wait until you actually have a use case and requirement before locking something in with a test. Specifically I wouldn't want to lock in that the config field is even public.

packages/code-analyzer-regex-engine/src/RegexEnginePlugin.ts
```diff
@@ -0,0 +1,69 @@
...
+	getCreatedEngine(engineName: string): EngineApi.Engine {
+		if (this.createdEngines.has(engineName)) {
+			return this.createdEngines.get(engineName) as EngineApi.Engine;
+		}
+		throw new Error(`Engine with name ${engineName} has not yet been created`);
+	}
+}
+
+export class RegexEngine extends EngineApi.Engine {
+	readonly config: EngineApi.ConfigObject;
```
stephen-carter-at-sf: I don't think you need a config object yet since you aren't using it for any reason. So I recommend you wait until you need it before you pass it in to the constructor from the plugin. Also if you do end up needing it - I wouldn't expose the readonly field as public at that time.

packages/code-analyzer-regex-engine/test/engine.test.ts
```diff
@@ -0,0 +1,99 @@
+import path from "node:path"
+import { RegexEngine, RegexEnginePlugin} from "../src/RegexEnginePlugin";
+import * as EngineApi from '@salesforce/code-analyzer-engine-api';
+import { EngineRunResults } from "@salesforce/code-analyzer-engine-api";
```
stephen-carter-at-sf: ditto.

packages/code-analyzer-regex-engine/test/engine.test.ts
```diff
@@ -0,0 +1,99 @@
...
+        const runOptions: EngineApi.RunOptions = {"filesToInclude": ["path/to/dir"] }
+        engine.runRules(ruleNames, runOptions);
+
+    })
+});
+
+describe('RegexEnginePlugin Tests' , () => {
+    
+    let config: EngineApi.ConfigObject = {"config": 5};
+    let engine: RegexEngine = new RegexEngine(config);
```
stephen-carter-at-sf: What is this for?

packages/code-analyzer-regex-engine/test/engine.test.ts
```diff
@@ -0,0 +1,99 @@
...
+
+    it('Check that I can get all available engine names', () => {
+        const availableEngines: string[] = ['regex'] 
+        expect(enginePlugin.getAvailableEngineNames()).toStrictEqual(availableEngines)
+    })
+   
+    it('Check that I can create engine using createEngine from the plugin (while retaining passed-in configuration information) ', () => {
+        enginePlugin.createEngine("regex", config);
+        const pluginEngine = enginePlugin.getCreatedEngine("regex") as RegexEngine;
+        expect(pluginEngine).toStrictEqual(engine)
```
stephen-carter-at-sf: This uses source implementation detail to help test. You don't want that typically. I would remove getCreatedEngine ... and just validate the engine via its properties instead of doing a strict equals check. Basically its best to ask yourself what you would like do expect with the returned value. Maybe just check that the return type is RegexEngine.

I don't think we need config right now... so maybe don't lock that in yet either.

packages/code-analyzer-regex-engine/test/engine.test.ts
```diff
@@ -0,0 +1,99 @@
...
+        let engine2: RegexEngine = new RegexEngine(config2);
+        enginePlugin.createEngine("regex", config);
+        enginePlugin.createEngine("regex", config2);
+        const pluginEngine = enginePlugin.getCreatedEngine("regex") as RegexEngine;
+
+        expect(pluginEngine).toStrictEqual(engine2)
+
+
+
+    });
```
stephen-carter-at-sf: I think this test isn't needed for 2 reasons:
* It is a premature optimization
* It may not actually be what we would want. In fact, I would say that create* methods... should do just that: create. So in anything, i would expect the createEngine to return a new instance of the engine each time I call it. But it's best to simply remove this test.

packages/code-analyzer-regex-engine/test/engine.test.ts
```diff
@@ -0,0 +1,99 @@
...
+    it('If I make an engine with an invalid name, it should throw an error with the proper error message', () => { 
+        expect(() => {enginePlugin.createEngine('OtherEngine', config)}).toThrow("Unsupported engine name: OtherEngine");
+        
+
+    });
+    it('If I try to access a plugin engine before it has been created, it should throw an error with the correct message', () => { 
+        expect(() => {enginePlugin.getCreatedEngine('OtherEngine')}).toThrow("Engine with name OtherEngine has not yet been created");
+        
+
+    });
```
stephen-carter-at-sf: I would remove this test. This getCreatedEngine method is not a part of the engine contract and isn't needed anywhere else in the source code... therefore we don't want to lock it in with a test.

packages/code-analyzer-regex-engine/package.json
```diff
@@ -0,0 +1,63 @@
...
+  "version": "0.1.0",
+  "author": "The Salesforce Code Analyzer Team",
+  "license": "BSD-3-Clause license",
+  "homepage": "https://developer.salesforce.com/docs/platform/salesforce-code-analyzer/overview",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/forcedotcom/code-analyzer-core.git",
+    "directory": "packages/code-analyzer-core"
+  },
+  "main": "dist/index.js",
```
stephen-carter-at-sf, line 13: Note that this designates the main module for the package (post typescript compilation step). So you need a index.ts file and in the index.ts file you should be exporting your plugin. In fact, you should follow the example that I have over at: https://github.com/forcedotcom/code-analyzer-core/blob/dev/packages/code-analyzer-retirejs-engine/src/index.ts

stephen-carter-at-sf (commented):
I know this is your first PR... so don't be taken back by all the comments I made. Think of it is a learning experience and take the time to understand each comment that I gave. Please meet with me if anything doesn't make sense.

Lastly, please make your PR title the following:

NEW (RegexEngine): @W-15973916@: Scaffolding for new regex engine

jag-j:
@ravipanguluri - Can you change the title to match the other PRs. The work item should be in this format - "@W-15973916@".

jag-j:
> Thanks for the contribution! Unfortunately we can't verify the commit author(s): Ravi Panguluri <r***@r***.i***.s***.com>. One possible solution is to [add that email to your GitHub account](https://github.com/settings/emails). Alternatively you can [change your commits to another email](https://help.github.com/articles/why-are-my-commits-linked-to-the-wrong-user/#commits-are-not-linked-to-any-user) and force push the change. After getting your commits associated with your GitHub account, [sign the Salesforce Inc. Contributor License Agreement](https://cla.salesforce.com/sign-cla?prUrl=https%3A%2F%2Fgithub.com%2Fforcedotcom%2Fcode-analyzer-core%2Fpull%2F29) and this Pull Request will be revalidated.

Please follow the steps in this doc: https://salesforce.quip.com/jBGaAuObdV3E to fix this issue.

packages/code-analyzer-regex-engine/src/RegexEnginePlugin.ts
jag-j: I also think we should think of a different name than simply "regex".

packages/code-analyzer-regex-engine/src/RegexEnginePlugin.ts
jfeingold35: Note: If you're using IntelliJ, it sometimes has a tendency to create weird redundant import statements like this. So just be sure to double-check every once in a while (or write your imports manually).

packages/code-analyzer-regex-engine/test/engine.test.ts
```diff
@@ -0,0 +1,99 @@
...
+import { EngineRunResults } from "@salesforce/code-analyzer-engine-api";
+
+
+describe('Regex Engine Tests', () => {
+    let engine: RegexEngine;
+    let config: EngineApi.ConfigObject = {"config" : 5}
+    beforeAll(() => {
+        engine = new RegexEngine(config);
+    });
+    it('Check that when I try to access an engine name, it is retrieved and the name is correct', () => {
```
jfeingold35, line 12: Minor nitpick: I personally prefer concise test names. E.g., "Engine name is accessible and correct."

packages/code-analyzer-regex-engine/test/engine.test.ts
jfeingold35: Agree. If the `config` property is necessary, it should be private.
Pro-tip: If, for some reason, you absolutely _need_ to access a private property for testing, you could add a public getter method that starts with an underscore (e.g., `_getConfig()`), which is the common convention for "this method is technically public but shouldn't be called normally". Alternatively, you can cast the object to an `any` and access it that way, e.g., `(engine as any).config`.

packages/code-analyzer-regex-engine/src/RegexEnginePlugin.ts
```diff
@@ -0,0 +1,69 @@
...
+
+	getCreatedEngine(engineName: string): EngineApi.Engine {
+		if (this.createdEngines.has(engineName)) {
+			return this.createdEngines.get(engineName) as EngineApi.Engine;
+		}
+		throw new Error(`Engine with name ${engineName} has not yet been created`);
+	}
+}
+
+export class RegexEngine extends EngineApi.Engine {
```
jfeingold35, line 20: Personal preference: Since these two classes aren't tightly related (e.g., an abstract class/interface and its primary extension/implementation), it might make sense to have this be in a separate file, for clarity? It's not a big deal here because both classes are relatively small, but it's something that could get hairy.

packages/code-analyzer-regex-engine/src/RegexEnginePlugin.ts
jfeingold35: Agree, this could just be `private readonly engine: EngineApi.Engine`.

packages/code-analyzer-regex-engine/src/RegexEnginePlugin.ts
jfeingold35: `RegEngine`? `RegExgine`? Are those anything?

packages/code-analyzer-regex-engine/src/RegexEnginePlugin.ts
jfeingold35: Recommendation: We should add a `.editorconfig` file containing this data.

packages/code-analyzer-regex-engine/src/RegexEnginePlugin.ts
jfeingold35: Alternatively/additionally, make the value itself include "TODO" so anyone who sees it knows.

packages/code-analyzer-regex-engine/test/engine.test.ts
```diff
@@ -0,0 +1,99 @@
...
+
+    it('When I make a call to runRules(), nothing should happen since it is currently a no-op', () => {
+        const ruleNames: string[] = ['TrailingWhitespaceRule']
+        const runOptions: EngineApi.RunOptions = {"filesToInclude": ["path/to/dir"] }
+        engine.runRules(ruleNames, runOptions);
+
+    })
+});
+
+describe('RegexEnginePlugin Tests' , () => {
```
jfeingold35, line 41: Personal preference: I prefer my test files to have only one top-level `describe`. It just makes the organization a little cleaner.

jfeingold35 (changes requested):

packages/code-analyzer-regex-engine/package.json
ravipanguluri: Just wondering because I am not intimately familiar with how this all works, but would it make more sense to revert the package.json file to the previous version then regenerate the new package-lock file to correct this mistake? The only reason is I don't know of a way to regenerate the package.json file from just the package-lock.json file, but maybe an npm run build call would do that.

packages/code-analyzer-regex-engine/src/RegexEnginePlugin.ts
ravipanguluri: How does "regexEngine" sound?

packages/code-analyzer-regex-engine/test/engine.test.ts
ravipanguluri: Would it be best to make all of the property checks into separate tests? Meaning one test checks that getName() is correct, another checks that describeRules() is correct, etc.

PR #31, opened 2024-06-19
NEW: @W-15884156@: Implement eslint engine's describeRules method

stephen-carter-at-sf:
I have only implemented the describeRules method so far. The runRules method will be implemented in the next PR.


packages/code-analyzer-eslint-engine/src/engine.ts
```diff
@@ -0,0 +1,136 @@
...
+    // update their code analyzer config file to manually add in the 'Recommended' tag for these special cases.
+    // Alternatively if users complain, then in the future we can add to the eslint engine a boolean field,
+    // like perform-full-rule-analysis, to the config which would trigger using all their workspace files here instead.
+    const cwd: string = process.cwd();
+    const extensionsToScan: string[] = ['.js', '.cjs', '.mjs', '.ts'];
+    return extensionsToScan.map(ext => `${cwd}${path.sep}dummy${ext}`);
+}
+
+function findLegacyConfigFile(): string | undefined {
+    const possibleUserConfigFiles = [".eslintrc.js", ".eslintrc.cjs", ".eslintrc.yaml", ".eslintrc.yml", ".eslintrc.json"];
```
jfeingold35, line 116: Can we add a citation for the page where it indicates that this is the order in which ESLint itself will check for files?

packages/code-analyzer-eslint-engine/src/strategies.ts
```diff
@@ -0,0 +1,78 @@
...
+            }
+        }
+        return allRulesMetadata;
+    }
+
+    private getAllRuleModules(): Map<string, Rule.RuleModule> {
+        // See https://github.com/eslint/eslint/discussions/18546 to see how we arrived at this implementation.
+        const legacyESLintModule: string = path.resolve(path.dirname(require.resolve('eslint')),'eslint','eslint.js');
+        // eslint-disable-next-line @typescript-eslint/no-var-requires
+        return require(legacyESLintModule).getESLintPrivateMembers(this.legacyESLint).cliEngine.getRules();
```
jfeingold35: A name like `getESLintPrivateMembers` immediately makes me nervous. Could we get a comment in the code explaining a little bit about this method and what expectations we can have for it?

packages/code-analyzer-eslint-engine/src/engine.ts
```diff
@@ -0,0 +1,136 @@
...
+    return tags;
+}
+
+function getFilesForRuleStatusCalculation(): string[] {
+    // To get the user selected status ("error", "warn" , or "off") of every single rule, we would theoretically have to
+    // ask eslint to calculate the configuration (which contains the rule status) for every single file that will be
+    // scanned. This is because users can add to their config that they only want certain rules to run for a particular
+    // file deep in their project. This is why ESLint only offers a calculateConfigForFile(filePath) method and not a
+    // getConfig method. See https://eslint.org/docs/v8.x/integrate/nodejs-api#-eslintcalculateconfigforfilefilepath
+    // If we were to do a full analysis using all the workspace files, it would be rather expensive. Instead, we will
```
jfeingold35: Question: How expensive are we talking? If we're talking about adding just another few seconds, that's honestly a sacrifice we might consider.
Another question: Does doing it this way introduce behavioral differences between `rules` and `run`?

packages/code-analyzer-eslint-engine/src/strategies.ts
```diff
@@ -0,0 +1,137 @@
...
+        }
+        this.allRulesMetadataCache = allRulesMetadata;
+        return allRulesMetadata;
+    }
+
+    private getAllRuleModules(legacyESLint: ESLint): Map<string, Rule.RuleModule> {
+        // See https://github.com/eslint/eslint/discussions/18546 to see how we arrived at this implementation.
+        const legacyESLintModule: string = path.resolve(path.dirname(require.resolve('eslint')),'eslint','eslint.js');
+        // eslint-disable-next-line @typescript-eslint/no-var-requires
+        return require(legacyESLintModule).getESLintPrivateMembers(legacyESLint).cliEngine.getRules();
```
jfeingold35, line 56: A name like `getESLIntPrivateMembers` immediately makes me nervous. Could we get a comment in the code explaining a little bit about this method and what expectations we can have for it to change or not change between releases?

packages/code-analyzer-eslint-engine/src/strategies.ts
```diff
@@ -0,0 +1,137 @@
...
+    }
+}
+
+function getSampleFilesForRuleStatusCalculation(): string[] {
+    // To get the user selected status ("error", "warn" , or "off") of every single rule, we would theoretically have to
+    // ask eslint to calculate the configuration (which contains the rule status) for every single file that will be
+    // scanned. This is because users can add to their config that they only want certain rules to run for a particular
+    // file deep in their project. This is why ESLint only offers a calculateConfigForFile(filePath) method and not a
+    // getConfig method. See https://eslint.org/docs/v8.x/integrate/nodejs-api#-eslintcalculateconfigforfilefilepath
+    // If we were to do a full analysis using all the workspace files, it would be rather expensive. Instead, we will
```
jfeingold35, line 125: Question: how expensive are we talking? If it's only the difference between 3 and 5 seconds (or even 5 and 10 seconds), then we might consider doing it anyway, particularly if it could lead to different behavior for `code-analyzer run` and `code-analyzer rules`.

packages/code-analyzer-eslint-engine/src/strategies.ts
```diff
@@ -0,0 +1,137 @@
...
+        const ruleStatuses: Map<string, ESLintRuleStatus> = new Map();
+        for (const config of configs) {
+            for (const ruleName of Object.keys(config["rules"])) {
+                const newStatus: ESLintRuleStatus = config["rules"][ruleName][0];
+                if (!ruleStatuses.has(ruleName)) {
+                    ruleStatuses.set(ruleName, newStatus);
+                    continue;
+                }
+                const existingStatus: ESLintRuleStatus = ruleStatuses.get(ruleName) as ESLintRuleStatus;
+                if (existingStatus === ESLintRuleStatus.ERROR || newStatus == ESLintRuleStatus.ERROR) {
```
jfeingold35, line 105: Optimization: Since a rule's severity can't be worse than `error`, could we take rules out of circulation in this inner loop once we assign them `error`?

packages/code-analyzer-eslint-engine/src/strategies.ts
```diff
@@ -0,0 +1,137 @@
...
+            baseConfig: LEGACY_BASE_CONFIG_ALL,
+            overrideConfigFile: this.userEslintConfigFile
+        }));
+
+        const ruleStatusForRecommended: Map<string, ESLintRuleStatus> = await this.calculateRuleStatusesFor(filesToScan, new ESLint({
+            baseConfig: LEGACY_BASE_CONFIG_RECOMMENDED,
+            overrideConfigFile: this.userEslintConfigFile
+        }));
+
+        for (const ruleInAll of ruleStatusForAll.keys()) {
```
jfeingold35, line 84: A comment indicating that we need to explicitly turn off all rules that aren't given a status by the Recommended config would go well here. (Or put this in a method with a name to that effect.)

packages/code-analyzer-eslint-engine/src/enums.ts
jfeingold35, line 1: It would be great to have some comments here indicating how "type" differs conceptually from "status", and what we use each one for.

packages/code-analyzer-eslint-engine/src/base-config.ts
```diff
@@ -0,0 +1,62 @@
+import {Linter} from "eslint";
+
+export const LEGACY_BASE_CONFIG_RECOMMENDED: Linter.BaseConfig = createLegacyBaseConfig('recommended');
+export const LEGACY_BASE_CONFIG_ALL: Linter.BaseConfig = createLegacyBaseConfig('all');
+
+
+// TODO: We need to review this and decide what we actually want as our base config. There are so many options.
```
jfeingold35, line 7: Maybe it could be configurable by the user in some way?

packages/code-analyzer-eslint-engine/test/engine.test.ts
```diff
@@ -0,0 +1,71 @@
...
+    });
+});
+
+describe('Tests for the ESLintEngine', () => {
+    let engine: Engine;
+    beforeEach(() => {
+        engine = new ESLintEngine();
+    });
+
+    it('When getName is called, then retire-js is returned', () => {
```
jfeingold35, line 37: "retire-js"?

packages/code-analyzer-eslint-engine/src/messages.ts
```diff
@@ -0,0 +1,27 @@
+const messageCatalog : { [key: string]: string } = {
```
jfeingold35, line 1: Since it seems like every engine is going to use this message pattern, could we extract it to the API somehow, so we don't need to copy-paste this file into every engine?

packages/code-analyzer-eslint-engine/test/test-data/cases/1_NoCustomization/placeholder.txt
```diff
@@ -0,0 +1 @@
+Just a placeholder file so that the folder exists.
```
jfeingold35, line 1: Question: Could you maybe make the expectation file go in the same subfolder instead of in a `blah_expectations` file at the `cases` level? Then you wouldn't need this placeholder file.

packages/code-analyzer-eslint-engine/test/test-data/cases/1_NoCustomization_ExpectedRuleDescriptions.json
jfeingold35, line 1: How much do we need to worry about these expectations changing between minor releases of the various plugins?

stephen-carter-at-sf:
Things are evolving a little bit away from this code... so i'm going to close this PR and i'll open up a new one soon.

PR #32, opened 2024-06-25
CHANGE: @W-16092798@: Introduce Workspace and update rule selection to take in a workspace

stephen-carter-at-sf:


packages/code-analyzer-core/src/code-analyzer.ts
```diff
@@ -11,30 +11,49 @@ import {getMessage} from "./messages";
...
     // For testing purposes only
-    setClock(clock: Clock) {
+    _setClock(clock: Clock) {
         this.clock = clock;
     }
+    _setUniqueIdGenerator(uniqueIdGenerator: UniqueIdGenerator) {
+        this.uniqueIdGenerator = uniqueIdGenerator;
+    }
+
+    public async createWorkspace(filesAndFolders: string[]): Promise<Workspace> {
```
stephen-carter-at-sf, line 53: Clients will need to use this to create a workspace first before they can call selectRules or runRules.

packages/code-analyzer-core/src/workspace.ts
```diff
@@ -0,0 +1,72 @@
...
+    getFilesAndFolders(): string[] {
+        return this.filesAndFolders;
+    }
+
+    async getExpandedFiles(): Promise<string[]> {
+        if (!this.expandedFiles) {
+            this.expandedFiles = await expandToListAllFiles(this.filesAndFolders);
+        }
+        return this.expandedFiles as string[];
+    }
```
stephen-carter-at-sf, line 34: This is where we get a big win. Instead of having each engine to do this themselves, the workspace object that they receive will be able to do this for them and cache the information so that other engines won't have to do it yet again.

packages/code-analyzer-core/src/utils.ts
```diff
@@ -13,4 +13,16 @@ export class RealClock implements Clock {
     now(): Date {
         return new Date();
     }
+}
+
+export interface UniqueIdGenerator {
```
jfeingold35, line 18: I've typically used [`uuid`](https://www.npmjs.com/package/uuid) for this sort of thing.

packages/code-analyzer-regex-engine/test/test-helpers.ts
```diff
@@ -16,3 +18,42 @@ export function changeWorkingDirectoryToPackageRoot() {
         process.chdir(original_working_directory);
     });
 }
+
+export class WorkspaceForTesting implements Workspace {
```
stephen-carter-at-sf, line 22: Eventually all these test tools for each of the various engines should probably end up in their own package. At that point maybe we'd call it @salesforce/code-analyzer-engine-test-helpers or something. Until then I'm fine living with the redundant code.

packages/code-analyzer-core/test/rule-selection.test.ts
```diff
@@ -213,19 +220,52 @@ describe('Tests for selecting rules', () => {
...
+    it('When attempting to get a rule that does not exist in the selection, then error', async () => {
+        const selection: RuleSelection = await codeAnalyzer.selectRules([]);
 
         expect(() => selection.getRule('stubEngine1', 'doesNotExist')).toThrow(
             getMessage('RuleDoesNotExistInSelection', 'doesNotExist', 'stubEngine1'));
         expect(() => selection.getRule('oopsEngine', 'stub1RuleD')).toThrow(
             getMessage('RuleDoesNotExistInSelection', 'stub1RuleD', 'oopsEngine'));
     });
+
+    it('When an engine returns multiple rules with the same name, then error', async () => {
```
jfeingold35, line 237: If I'm understanding correctly, this test got moved from `add-engines` to here because adding engines no longer intrinsically fetches all the rules from the engine the way it used to?

packages/code-analyzer-regex-engine/test/test-helpers.ts
jfeingold35: Question: How feasible would it be to move these into the `core` module instead? Yeah, it's not their final destination, but it still cuts down the redundant code and minimizes the risk of divergence.

jfeingold35 (approved):

packages/code-analyzer-regex-engine/test/test-helpers.ts
stephen-carter-at-sf: The engines don't and shouldn't depend on core. So either we move these into the engine-api or we put it into a test package. But yeah maybe the engine-api can have test helpers. I try not to bloat an api module though... so we'll defer this decision for now.

packages/code-analyzer-core/src/utils.ts
stephen-carter-at-sf: The issue with uuid or any other utility is that for testing, I might need to fix the generated ids when doing equality checking. Same issue with clock or random number generator. Have to be able to fix things. Because I needed an interface anyway, a so I went with a simple counter approach.

packages/code-analyzer-core/test/rule-selection.test.ts
stephen-carter-at-sf: That's right. I didn't remove any tests actually - they all got moved. And your understanding is spot on.

PR #33, opened 2024-06-27
NEW: @W-16092798@ Remove unwanted paths from workspace

stephen-carter-at-sf:


packages/code-analyzer-core/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@salesforce/code-analyzer-core",
   "description": "Core Package for the Salesforce Code Analyzer",
-  "version": "0.5.0",
+  "version": "0.5.1",
```
jag-j, line 4: What is the strategy we follow around versioning in the core library? Like when do we increase minor/patch versions?

packages/code-analyzer-core/test/workspace.test.ts
```diff
@@ -35,6 +35,14 @@ describe("Tests for the createWorkspace method of CodeAnalyzer", () => {
         expect(workspace.getFilesAndFolders()).toEqual([path.resolve('test')]);
     });
 
+    it("When including a .gitignore file or a node_modules folder, then it they are removed", async () => {
```
jag-j: typo here. Did you mean "then test they are removed?"

jag-j (approved):

packages/code-analyzer-core/test/workspace.test.ts
stephen-carter-at-sf: Yes that was a typo... fixed. Thanks.

packages/code-analyzer-core/package.json
stephen-carter-at-sf: Right now things are pretty loosy goosy before we get to v1.x. I'm considering this a fundamental missing component and thus I'm treating it sort of like a bug to be patched. So instead of incrementing it up a full minor step... just making this a patch update.

PR #36, opened 2024-06-28
CHANGE: @W-16092798@ Update workspace to be optional and add testtools

stephen-carter-at-sf:


packages/code-analyzer-core/src/code-analyzer.ts
```diff
@@ -133,7 +132,7 @@ export class CodeAnalyzer {
         this.eventEmitter.on(eventType, callback);
     }
 
-    private async getAllRules(workspace: Workspace): Promise<RuleImpl[]> {
+    private async getAllRules(workspace: Workspace | undefined): Promise<RuleImpl[]> {
```
jfeingold35: Why declare it this way instead of as `workspace?: Workspace`?

packages/code-analyzer-core/test/misc.test.ts
```diff
@@ -0,0 +1,25 @@
+import {LogLevel, RuleType, SeverityLevel} from "../src";
+import * as engApi from "@salesforce/code-analyzer-engine-api";
+
+describe('Misc tests', () => {
```
jfeingold35: Categorizing tests as "misc" tends to bug me, but I don't have a better name in mind. Maybe "interoperability"?

packages/code-analyzer-core/src/code-analyzer.ts
stephen-carter-at-sf: Ah good point. will fix.

packages/code-analyzer-core/test/misc.test.ts
stephen-carter-at-sf: Fair point... will update.

jfeingold35 (approved):

PR #37, opened 2024-07-01
NEW: @W-15884468@: Implement eslint engine's describeRules method

stephen-carter-at-sf:


packages/code-analyzer-core/src/code-analyzer.ts
```diff
@@ -68,12 +68,13 @@ export class CodeAnalyzer {
     }
 
     // This method should be called from the client with an absolute path to the module if it isn't available globally.
-    // Basically, clients should call this method after resolving the module using require.resolve. For example:
-    //     codeAnalyzer.dynamicallyAddEnginePlugin(require.resolve('./someRelativePluginModule'));
+    // Basically, clients should call this method after resolving the module using require.resolve with the directory of
+    // the config file. For example:
+    //     codeAnalyzer.dynamicallyAddEnginePlugin(require.resolve(moduleFromConfig, {paths: [<FolderOfConfigFile>]));
```
stephen-carter-at-sf, line 73: This actually makes me feel that if a config is provided, then we really need the config's location provided to core and the engines as well. This way, the client's would not need to resolve because core would resolve correctly.

Also, once we add in other path like values to the config (like the location of eslint config file, tsconfig file, etc)... then we'll need to know how to resolve relative roots. So I'll create a follow up PR after this to try to enhance our config constructors in a way where if a file is provided, I can get the config's directory... and if a file isn't provided, then the config's directory would need to be provided (otherwise cwd is used).

packages/code-analyzer-eslint-engine/src/engine.ts
```diff
@@ -0,0 +1,113 @@
...
+        name: name,
+        severityLevel: toSeverityLevel(metadata, status),
+        type: RuleType.Standard,
+        tags: toTags(metadata, status),
+        description: metadata.docs?.description || '',
+        resourceUrls: metadata.docs?.url ? [metadata.docs.url] : []
+    }
+}
+
+function toSeverityLevel(metadata: Rule.RuleMetaData, status: ESLintRuleStatus | undefined): SeverityLevel {
```
jag-j, line 93: This has some convoluted matching logic here. Would really help to have some comments as to why we are doing this.

packages/code-analyzer-eslint-engine/src/strategy.ts
```diff
@@ -0,0 +1,131 @@
...
+        for (const ruleName of await this.getAllBaseRuleNames()) {
+            if (!this.ruleStatuses.has(ruleName)) {
+                this.ruleStatuses.set(ruleName, ESLintRuleStatus.OFF);
+            }
+        }
+
+        return this.ruleStatuses;
+    }
+
+    private async calculateRuleStatusesFor(files: string[], legacyESLint: ESLint): Promise<Map<string, ESLintRuleStatus>> {
```
jag-j, line 66: Will this be the only method that changes with eslint 9? I'm trying to understand how much of this would change with config changes in eslint 9.

packages/code-analyzer-core/src/code-analyzer.ts
```diff
@@ -68,12 +68,13 @@ export class CodeAnalyzer {
...
     public async dynamicallyAddEnginePlugin(enginePluginModulePath: string): Promise<void> {
```
jfeingold35: Nitpick: Could we add "import" to the method name, to clarify that part of the "dynamism" is that you're importing and adding an arbitrary plugin?

packages/code-analyzer-eslint-engine/src/messages.ts
```diff
@@ -0,0 +1,30 @@
+const messageCatalog : { [key: string]: string } = {
+    CantCreateEngineWithUnknownEngineName:
+        `The ESLintEnginePlugin does not support creating an engine with name '%s'.`,
```
jfeingold35, line 3: These messages should be indicated as drafts, right?

packages/code-analyzer-eslint-engine/src/enums.ts
```diff
@@ -0,0 +1,11 @@
+export enum ESLintRuleStatus {
```
jfeingold35, line 7: Could you add some comments here explaining the conceptual difference between a Status and a Type? That's something that's _really_ easy to accidentally treat as interchangeable.

packages/code-analyzer-eslint-engine/src/base-config.ts
```diff
@@ -0,0 +1,43 @@
+import {Linter} from "eslint";
+
+export function createJavascriptPlusLwcBaseConfig(allOrRecommended: string): Linter.ConfigOverride {
```
jfeingold35: Does this mean that there's no config that's just base ESLint without LWC?

packages/code-analyzer-eslint-engine/src/strategy.ts
```diff
@@ -0,0 +1,131 @@
...
+                const existingStatus: ESLintRuleStatus | undefined = ruleStatuses.get(ruleName);
+                if (!existingStatus || existingStatus < newStatus) {
+                    ruleStatuses.set(ruleName, newStatus);
+                }
+            }
+        }
+        return ruleStatuses;
+    }
+
+    private createLegacyESLint(allOrRecommended: string, applyUserConfig: boolean, overrideConfig?: Linter.Config): ESLint {
```
jfeingold35: If `all` and `recommended` are the only acceptable values here, then to prevent developer errors it should probably be either a string-enum or have an assertion to enforce that.

packages/code-analyzer-eslint-engine/src/engine.ts
```diff
@@ -0,0 +1,113 @@
...
+        return SeverityLevel.High;
+    } else if (metadata.type === ESLintRuleType.LAYOUT) {
+        return SeverityLevel.Low;
+    }
+    return SeverityLevel.Moderate;
+}
+
+function toTags(metadata: Rule.RuleMetaData, status: ESLintRuleStatus | undefined): string[] {
+    const tags: string[] = [];
+    if (status === ESLintRuleStatus.ERROR || status === ESLintRuleStatus.WARN) {
```
jfeingold35, line 111: 1. Could you add a comment explaining why these two statuses correspond to "recommended"?
2. Also, I for some reason assumed that the "recommended" tag would key off of membership in the corresponding plugin's "recommended" config. Was that something you considered and decided against?

packages/code-analyzer-eslint-engine/src/workspace.ts
```diff
@@ -0,0 +1,183 @@
+import {Workspace} from "@salesforce/code-analyzer-engine-api";
+import fs from "node:fs";
+import path from "node:path";
+
+export interface ESLintWorkspace extends Workspace {
```
jfeingold35: Question: Does this interface actually need to extend `Workspace`? Looking at `MissingESLintWorkspace`, it seems like that might be causing more problems than it solves. Could changing this into an abstract class that encloses a `Workspace` simplify things?

packages/code-analyzer-eslint-engine/src/workspace.ts
```diff
@@ -0,0 +1,183 @@
...
+        throw new Error("This method should never be called");
+    }
+
+    /* istanbul ignore next */
+    async getExpandedFiles(): Promise<string[]> {
+        throw new Error("This method should never be called");
+    }
+
+    async getJavascriptFiles(): Promise<string[]> {
+        return this.javascriptExtensions.map(ext => `${this.workspaceRoot}${path.sep}someJavascriptFile${ext}`);
```
jfeingold35: Nitpick: I'm of the opinion that 
```
path.join(this.workspaceRoot, `someJavascriptFile${ext}`)`
```
is cleaner than using `path.sep` directly.

jfeingold35: Follow-up: Could you add a comment indicating that we're using "someJavascriptFile" as a placeholder since the workspace doesn't have the information we need?

packages/code-analyzer-eslint-engine/src/strategy.ts
stephen-carter-at-sf: The new flat config files (also available in eslint 8) will indeed require us to use the loadESLint utility to get a new ESLint constructor that is for flat configuration files. Therefore we will need to:
* Construct our base configs off of the new flat config structure
* Construct the ESLint instance from loadESLint instead of manually constructing the legacy one
* And the biggest one will be... in order to get metadata for the rules, we'll need to import and process all of the users configuration files and the import and process all of the plugin modules that we find along the way to get access to the rule metadata.

packages/code-analyzer-eslint-engine/src/engine.ts
stephen-carter-at-sf: Sure... will do. Note that our eslint engine design doc contains more info: https://docs.google.com/document/d/13M9nhS4tkhendq-9mRelnqP_rGU322mP7X0Ncxu-7Fs/edit#heading=h.uvab9trr4ep7

packages/code-analyzer-eslint-engine/src/workspace.ts
```diff
@@ -0,0 +1,183 @@
...
+        return candidateFiles;
+    }
+
+    getLegacyConfigFile(): string | undefined {
+        // TODO: Once we pass in user provided config file, then if it exists return that if it is a legacy one
+        return findLegacyConfigFile(makeUnique([this.getWorkspaceRoot(), process.cwd()]));
+    }
+}
+
+function getLongestCommonPrefix(strs: string[]): string {
```
jfeingold35, line 126: Observation: if you call this method with `src/some-file-1` and `src/some-file-2`, it returns `src/some-file-` instead of `src`. Is there something about the input to this method that allows us to confidently assert that the result from this method will always be a proper directory and never a partial file name? If so, could that be documented?

packages/code-analyzer-eslint-engine/src/engine.ts
stephen-carter-at-sf: OK, added in comments.

packages/code-analyzer-eslint-engine/src/engine.ts
```diff
@@ -0,0 +1,118 @@
...
+        description: metadata.docs?.description || '',
+        resourceUrls: metadata.docs?.url ? [metadata.docs.url] : []
+    }
+}
+
+function toSeverityLevel(metadata: Rule.RuleMetaData, status: ESLintRuleStatus | undefined): SeverityLevel {
+    if (status === ESLintRuleStatus.WARN) {
+        // An ESLint "warn" status is what users typically use to inform them of things without labeling it as a
+        // violation to stop their build over. Our Info level severity is the closest to this.
+        return SeverityLevel.Info;
```
jfeingold35, line 97: My understanding is that ESLint considers "warn" to be _more_ severe than "layout". By associating `warn` with `info` and `layout` with `low`, we're actually doing the opposite. Is that intentional?

packages/code-analyzer-eslint-engine/src/strategy.ts
```diff
@@ -0,0 +1,131 @@
...
+    const legacyESLintModule: string = path.resolve(path.dirname(require.resolve('eslint')),'eslint','eslint.js');
+    const {getESLintPrivateMembers} = await import(legacyESLintModule);
+    return getESLintPrivateMembers(legacyESLint).cliEngine.getRules();
+}
+
+function getRuleStatusFromRuleEntry(ruleEntry: Linter.RuleEntry): ESLintRuleStatus {
+    if (typeof ruleEntry === "number") {
+        return ruleEntry === 2 ? ESLintRuleStatus.ERROR : ruleEntry === 1 ? ESLintRuleStatus.WARN : ESLintRuleStatus.OFF;
+    } else if (typeof ruleEntry === "string") {
+        return ruleEntry === "error" ? ESLintRuleStatus.ERROR : ruleEntry === "warn" ? ESLintRuleStatus.WARN : ESLintRuleStatus.OFF;
```
jfeingold35: Do we need to worry about casing?

packages/code-analyzer-eslint-engine/src/strategy.ts
```diff
@@ -0,0 +1,131 @@
...
+    }
+    return getRuleStatusFromRuleEntry(ruleEntry[0]);
```
jfeingold35, line 136: What is `ruleEntry[0]` going to correspond to?

packages/code-analyzer-eslint-engine/test/engine.test.ts
```diff
@@ -0,0 +1,131 @@
...
+        await testFromCwdWithoutWorkspaceDescribeRules('legacyConfigCases', '1_NoCustomization');
+    });
+
+    it('When describing rules while cwd is folder with config that modifies existing rules, then return expected', async () => {
+        await testFromCwdWithoutWorkspaceDescribeRules('legacyConfigCases', '2_CustomizationOfExistingRules');
+    });
+
+    it('When describing rules while cwd is folder with config that adds a new plugin and rules, then return expected', async () => {
+        await testFromCwdWithoutWorkspaceDescribeRules('legacyConfigCases', '3_CustomizationWithNewRules');
+    });
```
jfeingold35: I think you could consolidate these tests by using `it.each()`. E.g.,
```
it.each([
{case: 'blah', caseParentFolder: 'blah', caseName: 'blah'}
...
...
])('When describing rules while cwd is $case, then return expected', ({caseParentFolder, caseName}) => {
// Contents of `testFromCwdWithoutWorkspaceDescribeRules()`
});
```

packages/code-analyzer-eslint-engine/test/engine.test.ts
```diff
@@ -0,0 +1,131 @@
...
+
+    async function testNotFromCwdWithWorkspaceDescribeRules(caseParentFolder: string, caseName: string): Promise<void> {
```
jfeingold35: The tests using this method could also probably be consolidated with `it.each()`.

packages/code-analyzer-core/src/code-analyzer.ts
stephen-carter-at-sf: I moved this over to: https://github.com/forcedotcom/code-analyzer-core/pull/38 which solves this issue.

packages/code-analyzer-eslint-engine/src/messages.ts
stephen-carter-at-sf: All messages are draft still - so I haven't marked them as draft (which is useful only if we have some approved and some not).

packages/code-analyzer-eslint-engine/src/enums.ts
stephen-carter-at-sf: Sure. Done.

packages/code-analyzer-eslint-engine/src/base-config.ts
stephen-carter-at-sf: I haven't add in configuration settings yet so you are just seeing the defaults. Once I add in configuration, you'll then see a createJavascriptBaseConfig method as well and each of these will be built up depending on what the user has enabled/disabled in their config.

packages/code-analyzer-eslint-engine/src/engine.ts
stephen-carter-at-sf: Where did you get your understanding? I'm happy to switch the severity definition if this is indeed the case.

packages/code-analyzer-eslint-engine/src/strategy.ts
stephen-carter-at-sf: Good catch. Fixed.

packages/code-analyzer-eslint-engine/src/strategy.ts
stephen-carter-at-sf: Added comment explaining.

packages/code-analyzer-eslint-engine/test/engine.test.ts
stephen-carter-at-sf: Good idea.

packages/code-analyzer-eslint-engine/test/engine.test.ts
stephen-carter-at-sf: Updated tests per suggestion.

packages/code-analyzer-eslint-engine/test/engine.test.ts
stephen-carter-at-sf: Updated tests per suggestion.

packages/code-analyzer-eslint-engine/src/engine.ts
jfeingold35: It seems I'm mistaken. I was conflating the status ("error", "warn", "off") with type ("problem", "suggestion", "layout"). Proposal: Perhaps `ESLintRuleStatus` should be renamed to `ESLintRuleBaseSeverity`, to avoid confusion with "type"?

packages/code-analyzer-eslint-engine/src/strategy.ts
```diff
@@ -123,9 +123,14 @@ async function getAllRuleModules(legacyESLint: ESLint): Promise<Map<string, Rule
 
 function getRuleStatusFromRuleEntry(ruleEntry: Linter.RuleEntry): ESLintRuleStatus {
     if (typeof ruleEntry === "number") {
-        return ruleEntry === 2 ? ESLintRuleStatus.ERROR : ruleEntry === 1 ? ESLintRuleStatus.WARN : ESLintRuleStatus.OFF;
+        return ruleEntry === 2 ? ESLintRuleStatus.ERROR :
```
jfeingold35, line 127: Question: Since `ruleEntry` is a number here, and the status is a numerical Enum, doesn't that mean you could just do `return ruleEntry as ESLintRuleStatus`?

stephen-carter-at-sf: If the user puts in a dumb value like 3... I'd like to just turn it off. This is why i'm not just casting.

packages/code-analyzer-eslint-engine/src/engine.ts
stephen-carter-at-sf: I'm trying to avoid using the word Severity again because throughout the code we have our own Severity. Would calling it ESLintRuleState help instead?

Notice my current comment in the code:

```
/**
 * Enum that maps to the ESLint Rule Severity levels.
 * See https://eslint.org/docs/v8.x/use/configure/rules#rule-severities
 * Using the term "Status" here to differentiate from our own "Severity" term.
 * Also using numbers so that we can more easily compare them with > sign.
 */
export enum ESLintRuleStatus {
    ERROR = 2,
    WARN = 1,
    OFF = 0
}
```

jfeingold35: Fair enough. We can leave it at `status`.

packages/code-analyzer-eslint-engine/src/strategy.ts
jfeingold35: Fair.

packages/code-analyzer-eslint-engine/src/strategy.ts
stephen-carter-at-sf: Good idea. Fixed.

packages/code-analyzer-eslint-engine/src/engine.ts
stephen-carter-at-sf: Regarding #1 - yes... will add comment: 
```
        // Any rule that base config or the user's config has turned on, should be marked as 'Recommended'
        // so that code analyzer will run these rules by default just as eslint runs these rules.
```

Regarding #2 - see:
https://docs.google.com/document/d/13M9nhS4tkhendq-9mRelnqP_rGU322mP7X0Ncxu-7Fs/edit#heading=h.uvab9trr4ep7

The recommended sets that we are using do in fact use the "recommended" metadata when it comes down to the base set. But note that most plugin authors don't use the "recommended" metadata at all... but instead define different configurations that users can use (where one of their configs is usually called "recommended").

packages/code-analyzer-eslint-engine/src/workspace.ts
stephen-carter-at-sf: You are correct. I fixed this.

packages/code-analyzer-eslint-engine/src/workspace.ts
stephen-carter-at-sf: path.join does a lot more work than we need since it attempts to resolve and do other things we don't need. Since I already need to concatenate in the extension, this actually is cleaner.

And regarding someJavascriptFile... i'll change it to placeholderJavascriptFile

packages/code-analyzer-eslint-engine/src/workspace.ts
stephen-carter-at-sf: This getLongestCommonPrefix is only half the algorithm. Note that I took this exact algorithm from our old scanner code where I discover where the projectDir is.

The rest of the code is:
```
            const longestCommonStr: string = getLongestCommonPrefix(this.delegateWorkspace.getFilesAndFolders());
            this.workspaceRoot = fs.existsSync(longestCommonStr) && fs.statSync(longestCommonStr).isDirectory() ?
                longestCommonStr : path.dirname(longestCommonStr);
```

Clearly `src/some-file-` is not a folder and therefore we trim off the `-some-file-` with the path.dirname(longestCommonStr) to give the `src` folder.

packages/code-analyzer-eslint-engine/src/engine.ts
jfeingold35: I think we're talking past each other. I understand that there's usually a config called "recommended". My question was had you considered using that config to determine which rules should get the "recommended" tag, but upon deeper inspection I now see that you're already doing that, because the status itself comes from the config.
So, my bad, I think this looks fine.

jfeingold35 (approved):

PR #40, opened 2024-07-05
NEW (eslint): @W-16113413@ Add configuration abilities to eslint engine

stephen-carter-at-sf:
This PR contains:
* Moved configuration validation utilities into the engine-api to be shared by the engines and core
* Moved message catalog utility into engine-api to be shared by the engines and core
* Added in eslint configuration definitions
* Applied configuration to eslint engine and added in tests


packages/code-analyzer-eslint-engine/src/utils.ts
```diff
@@ -0,0 +1,37 @@
+import fs from "node:fs";
+import path from "node:path";
+
+export function calculateLongestCommonParentFolderOf(paths: string[]): string {
```
jag-j: I don't see tests for these. Am I missing something?

stephen-carter-at-sf: 1) This utility was actually a part of the setWorkspaceRoot method before and was moved here and renamed a utility as implementation detail. This refactor was just to help with readability mainly.
2) I like to write tests at the public boundary (in this case the engine api) and so this is implicitly tested. If every single function had its own set of test functions... then when we refactor then we'd have to rewrite a ton of tests.
3) This utility is not being used by anything other than our main functionality. So why test it? There is no outside contract on it. If all of our tests for the main engine pass (which is the main contract)... then what is the purpose of testing this implementation detail?
4) This utility code was actually written and used over in the old scanner - so it has stood the test of time.

But note that if we were to unit test this directly, then the question is... do we add in more code for cases that this utility doesn't support. For example, it doesn't support an empty array. This is because the only place we use this code already handles this case:
            this.workspaceRoot = filesAndFolders.length == 0 ? this.config.config_root :
                calculateLongestCommonParentFolderOf(filesAndFolders);
                
So if we were to test this function for this case, then we'd be adding code for things that our main functionality doesn't even care about. This is fine if we plan on reusing this function... but then we'd need to decide what the output is supposed to be in this empty array case. When doing test driven development, i write tests for the contract and then I write the source code to make those tests pass. Going the other way can make you write tests and code that you don't even need.

But, if it makes you feel more comfortable to test little utilities like this, I can do so directly... I just don't see the value in it. Let me know.

stephen-carter-at-sf: Since it only took a few minutes (and to avoid the back and forth on this), I went ahead and updated these utility functions to cover edge cases (which aren't used by my code in particular - but is available if we want to share this utility) and added in unit tests per your suggestion.

jag-j (approved):

packages/code-analyzer-eslint-engine/test/utils.test.ts
```diff
@@ -0,0 +1,68 @@
...
+    it('When a various files and folders are provided, then the longest common parent folder is returned', () => {
+        expect(calculateLongestCommonParentFolderOf([
+            path.join(__dirname, 'test-data', 'legacyConfigCases', 'expectedRules_DefaultConfig.json'),
+            path.join(__dirname, 'test-data','legacyConfigCases', 'workspace_NoCustomConfig'),
+            path.join(__dirname, 'test-data', 'legacyConfigCases', 'workspaceHasCustomConfigWithNewRules', 'dummy1.js'),
+        ])).toEqual(path.join(__dirname, 'test-data', 'legacyConfigCases'));
+    });
+
+    it('When strings are provided which are not absolute paths that share a common root folder, then return null', () => {
+        expect(calculateLongestCommonParentFolderOf(['oops', 'nope'])).toStrictEqual(null);
```
jfeingold35, line 40: Could you add a comment explaining why we chose to return null, instead of an empty string or throwing an error?

packages/code-analyzer-core/test/run.test.ts
```diff
@@ -409,33 +409,33 @@ describe("Tests for the run method of CodeAnalyzer", () => {
...
                 'stubEngine1', 'stub1RuleA', expectedBadField, expectedBadValue));
     }
-    it("When an engine returns a code location that has an invalid line or column, then an error is thrown", () => {
-        testInvalidLineOrColumnScenario(-1, 2, 3, 4, 'startLine', -1);
-        testInvalidLineOrColumnScenario(1.5, 2, 3, 4, 'startLine', 1.5);
-        testInvalidLineOrColumnScenario(1, 0, 3, 4, 'startColumn', 0);
-        testInvalidLineOrColumnScenario(1, 3.124, 3, 4, 'startColumn', 3.124);
-        testInvalidLineOrColumnScenario(1, 2, -1.2, 4, 'endLine', -1.2);
-        testInvalidLineOrColumnScenario(1, 2, 3, 0, 'endColumn', 0);
+    it("When an engine returns a code location that has an invalid line or column, then an error is thrown", async () => {
```
jfeingold35, line 416: Could you rewrite this test to use `it.each()`?

packages/code-analyzer-engine-api/test/config.test.ts
```diff
@@ -0,0 +1,364 @@
...
+    it("When calling extractFolder on a field that is a path that does not exist, then error", () => {
+        const extractor1: ConfigValueExtractor = new ConfigValueExtractor({config_root: __dirname, some_field1: 'doesNotExist'});
+        expect(() => extractor1.extractFolder('some_field1')).toThrow(
+            getMessage('ConfigPathValueDoesNotExist','some_field1', path.resolve(__dirname, 'doesNotExist')));
+        const extractor2: ConfigValueExtractor = new ConfigValueExtractor({some_field2: 'doesNotExist'}, 'engines.dummy');
+        expect(() => extractor2.extractFolder('some_field2')).toThrow(
+            getMessage('ConfigPathValueDoesNotExist','engines.dummy.some_field2', path.resolve(process.cwd(), 'doesNotExist')));
+    });
+
+    it("When calling extractFolder on a field that is a folder, then error", () => {
```
jfeingold35, line 355: "field that is a file"?

packages/code-analyzer-eslint-engine/test/utils.test.ts
stephen-carter-at-sf: See the comment on the calcualteLongestCommonParentFolderOf ... but i'll update it even more in the next PR with:
```
/**
 * Returns the longest common parent folder of the provided paths.
 *
 * @param paths It is assumed that paths is a non-empty array of absolute value paths.
 * If empty or if no common parent folder exists (like in the case on Windows machines of using two different drives
 * C: and D:) then null is returned (to allow the client of this function to handle this case without needing try/catch)
 */
 ```

packages/code-analyzer-core/test/run.test.ts
stephen-carter-at-sf: Sure.

packages/code-analyzer-engine-api/test/config.test.ts
stephen-carter-at-sf: Good catch. Will fix in next PR.

PR #41, opened 2024-07-09
NEW (eslint): @W-16113415@: Implement the runRules method of eslint engine

stephen-carter-at-sf:


packages/code-analyzer-eslint-engine/test/end-to-end.test.ts
```diff
@@ -0,0 +1,48 @@
...
+import path from "node:path";
+import {changeWorkingDirectoryToPackageRoot} from "./test-helpers";
+
+changeWorkingDirectoryToPackageRoot();
+
+/**
+ * NOTE THAT WE WANT TO KEEP THE AMOUNT OF TESTS HERE TO A MINIMUM!
+ * All functionality should be tested at the unit level. This file ideally should only contain 1 (maybe 2) tests
+ * at most to simply confirm that things can wire up correctly without failure.
+ */
```
stephen-carter-at-sf, line 20: Currently these tests don't provide much value over the actual engine tests... so if you are wondering why I have them, it is part of a bigger picture idea:

I'd like to create a pseudo end to end test for each engine... so that eventually we can see the common overlapping pattern between them all which will eventually help me create a generalized end-to-end test utility that can live in the engine api (which will eventually help engine api authors to testing more easily). After one or two more engines, i think we'll have enough information to then create the generalized utility.

jfeingold35 (approved):

PR #42, opened 2024-07-10
CHANGE (regex): @W-16150675@ Updates to regex engine

ravipanguluri:
In this PR:

  * updates to regex engine


packages/code-analyzer-regex-engine/src/messages.ts
```diff
@@ -0,0 +1,21 @@
...
+    CantCreateEngineWithUnknownEngineName:
+        `The RegexEnginePlugin does not support creating an engine with name '%s'.`,
+
+    TrailingWhitespaceRuleDescription:
+        "Detects trailing whitespace (tabs or spaces) at the end of lines of code and lines that are only whitespace.",
+
+    PrivateMethodApiVersionRuleDescription:
+        "For API versions 60.0 and below, declaring a private method in an abstract or virtual class will override a method with the same signature in a child class. Please update your API version to resolve the issue."
+}
+/*Could make this convenience function globally available, but for now just copy-pasting from eslint engine */
```
stephen-carter-at-sf: Please remove this comment. 

The getMessageFromCatalog is globally available. But your convenience getMessage function uses your MESSAGE_CATALOG so you don't need to export it and use it everywhere... so we wouldn't make that globally available.

packages/code-analyzer-regex-engine/src/RegexEnginePlugin.ts
```diff
@@ -44,13 +53,21 @@ export class RegexEngine extends Engine {
...
         ];
     }
 
     async runRules(_ruleNames: string[], runOptions: RunOptions): Promise<EngineRunResults> {
-        const executor = new RegexExecutor()
-        const fullFileList: string[] = await runOptions.workspace.getExpandedFiles()
-        const violations: Violation[] = await executor.execute(fullFileList)
+        const fullFileList: string[] =  await runOptions.workspace.getExpandedFiles()
+        const filesToUse: string[] = (this.regexExecutor instanceof SimpleRegexExecutor) ? fullFileList : fullFileList.filter((file) => (path.extname(file) === ".cls") && (fullFileList.includes((file + "-meta.xml"))))
+        const violations: Violation[] = await this.regexExecutor.execute(filesToUse)
```
stephen-carter-at-sf: Your constructor and this method seems wrong. Since your TrailingWhiteSpaceRule is associated with your SimpleRegeExecutor and your PrivateMethodApiVersionRule is assocaited with your PrivateMethodApiVersionExecutor then shouldn't you be constructing one or more executors here based on what rules are selected and then running each of them, resolving their violations into a single output?

Again this is the down side of having these very different rules in the same engine.

stephen-carter-at-sf (changes requested):

packages/code-analyzer-regex-engine/test/engine.test.ts
```diff
@@ -37,42 +44,92 @@ describe('Regex Engine Tests', () => {
...
+});
 
-    describe('runRules() tests', () => {
+    describe('ruleRules() PrivateMethodAPIVersionRule tests',  () => {
         let ruleNames: string[];
+        let engine: RegexEngine;
         beforeAll(() => {
-            ruleNames = ['TrailingWhitespaceRule']
+            ruleNames = ["PrivateMethodAPIVersionRule"]
+            engine = new RegexEngine(new PrivateMethodApiVersionExecutor())
```
stephen-carter-at-sf: This is where I believe you went wrong. If you liked this constructor approach, then you would need to pass in all of your executors, and each executor would then need to tell your engine what rules it supports.

But before going too complicated, just start with a test that allows you to run all available rules at the same time.

stephen-carter-at-sf (changes requested):

packages/code-analyzer-regex-engine/src/RegexEnginePlugin.ts
```diff
@@ -44,13 +54,24 @@ export class RegexEngine extends Engine {
...
     async runRules(_ruleNames: string[], runOptions: RunOptions): Promise<EngineRunResults> {
-        const executor = new RegexExecutor()
-        const fullFileList: string[] = await runOptions.workspace.getExpandedFiles()
-        const violations: Violation[] = await executor.execute(fullFileList)
+        const fullFileList: string[] =  await runOptions.workspace.getExpandedFiles()
+        const violations: Violation[] = [];
+        for (const executor of this.regexExecutors) {
+            const filesToUse: string[] = (executor instanceof SimpleRegexExecutor) ? fullFileList : fullFileList.filter((file) => (path.extname(file) === ".cls") && (fullFileList.includes((file + "-meta.xml"))))
+            violations.push(...await executor.execute(filesToUse))
+        }
```
stephen-carter-at-sf: A few things...
1) Shouldn't the executor determine which files to target? Any use of instanceof other than validating unknown types is usually a code smell.
2) You still are not using the ruleNames. So if the user only selected one of the rules, this would still run both of them.

packages/code-analyzer-regex-engine/src/RegexEnginePlugin.ts
```diff
@@ -22,13 +26,19 @@ export class RegexEnginePlugin extends EnginePluginV1 {
...
     }
 }
 
 export class RegexEngine extends Engine {
     static readonly NAME = "regex"
+    private readonly regexExecutors: RegexExecutor[];
+
+    constructor(executors?: RegexExecutor[]) {
+        super();
+        this.regexExecutors = executors ? executors : [new SimpleRegexExecutor(), new PrivateMethodApiVersionExecutor()];
```
stephen-carter-at-sf: What is the purposes of passing in executors to the constructor?

And if not all executors are provided, then shouldn't the list of rules be dependendent on which executors are available? 

The only way to make this design work is to essentially hand off all responsibilities to each of the executors... then the RegexEngine would essentially just be delegating all of its work to the executors.

But then what's the point of the regex engine at that point... so I think this is evidence enough that the 2nd rule doesn't belong in this regex engine.

packages/code-analyzer-regex-engine/src/RegexEnginePlugin.ts
```diff
@@ -43,14 +44,14 @@ export class RegexEngine extends Engine {
...
 
     async runRules(_ruleNames: string[], runOptions: RunOptions): Promise<EngineRunResults> {
+        const fullFileList: string[] =  await runOptions.workspace.getExpandedFiles()
         const executor = new RegexExecutor()
-        const fullFileList: string[] = await runOptions.workspace.getExpandedFiles()
-        const violations: Violation[] = await executor.execute(fullFileList)
+        const violations: Violation[] = await executor.execute(fullFileList);
         return {
             violations: violations
         };
```
stephen-carter-at-sf: You aren't allowing rule selection to work. I didn't get a chance to review this code before... but the same pattern is happening with your metadata engine. The point of us passing in the rule names into the run method is because that is the list of rules that the user has selected.

Fortunately, you only have 1 rule right now on each of your engines, so this runRules method will currently only be called by the core module with the 1 rule (and core would never call with an empty ruleNames array). But ideally you would want the code setup today to prepare for multiple rules.

packages/code-analyzer-metadata-engine/src/executor.ts
```diff
@@ -0,0 +1,137 @@
+import os from "node:os";
+import {Violation} from "@salesforce/code-analyzer-engine-api";
+import fs from "node:fs";
+import {getMessage} from "./messages";
+import {getColumnNumber, getLineNumber, getNewlineIndices} from "./utils";
+import path from "node:path";
+
+const APEX_CLASS_FILE_EXT = ".cls"
+
+export class MetadataExecutor {
```
stephen-carter-at-sf: I'm still struggling to understand what the point of the executor is (especially in the current Regex Engine). What I don't like about is that it is separating the rule description from the logic so that they aren't tied in any way together. In the future as we add in more rules... how would rule selection work?

If you were to have something like an Executor... then I would say what is its single responsibility? Currently it looks like... it executes a single rule and returns violations for that rule... but how does it know what to execute... and how is this different than the responsibility of the run method of the engine?

If instead, you wanted to tie this directly to the specific private method api version rule... then I would think that this would be named as such and would also contain the description information of the rule as well. Then the engine could essentially hold these rules and delegating to them to get their descriptions and to execute when needed.

packages/code-analyzer-metadata-engine/src/executor.ts
```diff
@@ -0,0 +1,137 @@
...
+    lineSep: string
+
+    constructor() {
+        this.lineSep = os.EOL
+    }
```
stephen-carter-at-sf: Is there a reas you are storing the lineSep as state on the executor classes? os.EOL is simple enough to just pass in where we need it. Why store it?

packages/code-analyzer-metadata-engine/src/executor.ts
```diff
@@ -0,0 +1,137 @@
...
+
+    async execute(allFiles: string[]): Promise<Violation[]> {
+        const violations: Violation[] = [];
+
+        await Promise.all(
+            allFiles.map(async (file) => {
+                const fileStats = await fs.promises.stat(file)
+                if (fileStats.isFile()) {
```
stephen-carter-at-sf: What is this isFile check for? It looks like you are already (inside of your engine) getting these allFiles from the getExpandedFiles which already returns nothing but files. So there are no folders in the mix anymore since you chose to work directly on files. 

So your entire execute method might end up simply being a single line:
```
    return Promise.all(allFiles.map(f => this.scanFile(f))).flat();
```

packages/code-analyzer-metadata-engine/src/executor.ts
```diff
@@ -0,0 +1,137 @@
...
+                    violations.push(...await this.scanFile(file));
+                }
+            })
+        );
+
+        return violations;
+    }
+
+    private getAPIVersion(fileContents: string): number {
+        const regex = /(?<=ApexClass.+>[\s\S]+<apiVersion>)[0-9\\.]+(?=<\/apiVersion>)(?=[\s\S]+<\/ApexClass>)/g
```
stephen-carter-at-sf: Parsing xml files with regex isn't ideal actually... especially if people add in comments into their xml file that you aren't considering.

So instead, see if you can find a way of using XPath with node.js. XPath is very commonly used in frontend javascript to navigate the html dom... but I believe node most likely offers the ability to start with an xml document and run an XPath query on it. Look into it.

packages/code-analyzer-metadata-engine/src/executor.ts
```diff
@@ -0,0 +1,137 @@
...
+            apiVersion = parseFloat(regexMatch[0]);
+        }
+
+        return apiVersion;
+    }
+
+    private extractVirtualAbstractClass(fileContents: string): IterableIterator<RegExpExecArray> {
+        const accessModifier = '(public|private|global)'
+        const classType = '(virtual|abstract)'
+        const classDeclaration = `${accessModifier}\\s+${classType}\\s+class[^{]+(?={)`
```
stephen-carter-at-sf: With apex, is it possible to have public and virual and class on separate lines? Might want to experiment with apex to find out. If not, then we should reduce the expression to only be spaces and possible tab characters (if they are allowed).

Basically if a user had a comment where they ended with "public abstract" and then the next line was just a regular class... then we wouldn't want to pick these up.

Lastly, is it possible to flip the order of public virtual vs virtual public? Something to look into.

packages/code-analyzer-metadata-engine/src/executor.ts
```diff
@@ -0,0 +1,137 @@
...
+        const classContentRegex = `([\\s\\S+]+)}`
```
stephen-carter-at-sf: Help me understand this.
Are you trying to find the entire class? If so, wouldn't this stop after the first if statement (or whatever else has a bracket). I don't know if we can with a regular expression find the end of the class definition. We might need to just stop at the signature without a formal AST.

packages/code-analyzer-metadata-engine/src/executor.ts
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    private async metadataExists(fileName: string): Promise<boolean> {
+        const metadataFileName = `${fileName}-meta.xml`
+        try {
+            await fs.promises.access(metadataFileName)
+            return true;
+        } catch (err) {
+            return false;
+        }
```
stephen-carter-at-sf: It's best to put this in a helper method that's called fileExists or something. This is a common pattern in node to try to access it as a means of detecting if it exists since its the fastest way to know without doing a lookup.

packages/code-analyzer-metadata-engine/src/executor.ts
```diff
@@ -0,0 +1,137 @@
...
+                    const newlineIndexes = getNewlineIndices(fileContents, this.lineSep);
+                    for (const classBlock of virtualAbstractClassCode) {
+                        const privateMethodMatches = this.extractPrivateMethods(classBlock[0]);
+                        violations.push(...this.getPrivateMethodViolations(privateMethodMatches, fileContents, fileName, newlineIndexes));
+                    }
+                }
+            }
+        }
+        return violations;
+    }
```
stephen-carter-at-sf: The McCabe Cyclomatic Complexity of this file is probably very large due to all the nested conditionals.

I can show you some refactoring tricks on how to reduce this complexity which helps with readability, and helps avoid bugs.

One small trick that you should be doing is simply doing quick returns or filter operations.

I would have earlier one just wrote a boolean method that I could then use on my initial set of files to filter it down. Then only loop over the remaining files. Alternatively, just shift the code left by negating your conditional and do a quick return.

packages/code-analyzer-metadata-engine/src/executor.ts
```diff
@@ -0,0 +1,137 @@
...
+
+    private getPrivateMethodViolations(
+        privateMethodMatches: IterableIterator<RegExpExecArray>,
+        fileContents: string,
+        fileName: string,
+        newlineIndexes: number[]
+    ): Violation[] {
+        const violations: Violation[] = [];
+        for (const privateMethod of privateMethodMatches) {
+            if (!this.isContentInComment(fileContents.slice(privateMethod.index, fileContents.length))) {
```
stephen-carter-at-sf: I'm a little confused by this. In order to tell if the privateMethod is inside of a comment... shouldn't you be checking the start of the file up to the index where you found the virtual private method in order to see if we are in a comment?
Instead it looks like you are get the oposite.

packages/code-analyzer-metadata-engine/src/executor.ts
```diff
@@ -0,0 +1,137 @@
...
+                } else if (chars === '*/') {
+                    if (stack.length === 0) {
+                        return true;
+                    }
+                    stack.pop();
+                }
+            }
+        }
+        return !(stack.length === 0)
+    }
```
stephen-carter-at-sf: This handles the /* ... */ case it seems but what about the case where a user has

```
// private virtual class myClass...
```
as an inline comment?

packages/code-analyzer-metadata-engine/src/messages.ts
```diff
@@ -5,7 +5,10 @@ const MESSAGE_CATALOG : { [key: string]: string } = {
         `The MetadataEnginePlugin does not support creating an engine with name '%s'.`,
 
     PrivateMethodApiVersionRuleDescription:
-        "Enforces a minimum API version for declaring private methods in abstract/virtual apex classes."
+        "Enforces a minimum API version for declaring private methods in abstract/virtual apex classes.",
```
stephen-carter-at-sf: I know this will get reviewed at some point... but to save the doc writer from doing more work, the term "Apex" will need a capital letter A believe. You might want to look through our help doc to see if Abstract and or Virtual also need to be capital letters

packages/code-analyzer-metadata-engine/src/plugin.ts
```diff
@@ -7,9 +7,10 @@ import {
     RuleDescription,
     RuleType,
     RunOptions,
-    SeverityLevel,
+    SeverityLevel, Violation,
```
stephen-carter-at-sf: nitpick:
If listing imports on multiple lines, then keep them all on their own line. Also, avoid trailing comma at the end of the list.

packages/code-analyzer-metadata-engine/test/engine.test.ts
```diff
@@ -2,12 +2,13 @@ import {changeWorkingDirectoryToPackageRoot} from "./test-helpers";
 import {MetadataEngine, MetadataEnginePlugin} from "../src/plugin";
 import * as testTools from "@salesforce/code-analyzer-engine-api/testtools"
 import {
-    EngineRunResults,
+    CodeLocation,
     RuleDescription,
-    RuleType,
-    SeverityLevel
+    RuleType, RunOptions,
+    SeverityLevel, Violation
```
stephen-carter-at-sf: Ditto

stephen-carter-at-sf (changes requested):
Just reviewed your source code so far. I'll review your test code soon.

packages/code-analyzer-metadata-engine/src/executor.ts
ravipanguluri: Assuming that I am not given malformed apex when I am doing the search I am checking for an extra close without a matching open (if the method content is in a comment) because it indicates the open comment came before the method declaration . I could just have easily done it the way you're suggesting, I would just be looking for an extra open comment (without a required close before the method declaration) instead of an extra close comment. I think it works the same either way you do it.

packages/code-analyzer-metadata-engine/src/executor.ts
ravipanguluri: Quickly checked both, and I think the answer to each is no. In which case, you're right in that I should just change the \s+ to [ \t]+.

packages/code-analyzer-metadata-engine/src/executor.ts
ravipanguluri: From some research, it appears that I can do it in node.js. But it will unfortunately require extra libraries to do it.

Also I highlighted the advantages of using XPath queries and parsing the dom, but I was met with resistance because others thought it was too close to AST parsing. I should have probably advocated for this more strongly since with more complicated rules, I believe it will be necessary.

packages/code-analyzer-metadata-engine/src/executor.ts
ravipanguluri: I think at this point your are correct that the regex engine should be moved over to a single class. Then for testing purposes I could just make the regex expression a class constant so then once configuration is factored it it's pretty easy to make the transfer.

For the regex to catch the private method declaration, I am worried about being able to effectively shove it all into a single regex expression (which would probably be needed to make it customizable) because of the multiline comment case. This rule feels like somewhat of a special case even in the metadata world, so I am not entirely sure how to effectively incorporate it.

packages/code-analyzer-metadata-engine/src/executor.ts
ravipanguluri: The regex only ends the match after the last close bracket because the [\s\S]+ acts as a catch-all to make it match only the last }. Anyways there are tests where there are multiple private methods and the position of the private method is changed, so I think this expression works.

packages/code-analyzer-metadata-engine/src/executor.ts
stephen-carter-at-sf: Yes if we have an xpath utility... it can be reused by us to more easily be able to create xpath based rules (and migrate the appexchange rules over here eventually). We can discuss in tomorrow's standup.

stephen-carter-at-sf:
Closing to start a new one.

PR #45, opened 2024-07-11
CHANGE: @W-16113417@ Refactor progress events

stephen-carter-at-sf:


packages/code-analyzer-core/src/code-analyzer.ts
```diff
@@ -130,20 +127,39 @@ export class CodeAnalyzer {
     }
 
     private async getAllRules(workspace?: Workspace): Promise<RuleImpl[]> {
-        const rulePromises: Promise<RuleImpl[]>[] = this.getEngineNames().map(
-            engineName => this.getAllRulesFor(engineName, {workspace: workspace}));
-        return (await Promise.all(rulePromises)).flat();
+        const cacheKey: string = workspace ? workspace.getWorkspaceId() : process.cwd();
+        if (!this.rulesCache.has(cacheKey)) {
+            this.engineRuleDiscoveryProgressAggregator.reset(this.getEngineNames());
```
jfeingold35, line 142: Do we ever intend to support multiple workspaces at once? If so, then this would need to change, otherwise each workspace would reset the progress bar from the last one, which sounds undesirable to me.

packages/code-analyzer-engine-api/src/engines.ts
```diff
@@ -41,4 +41,26 @@ export abstract class Engine {
     protected emitEvent<T extends Event>(event: T): void {
```
jfeingold35, line 41: Should this still be `protected`, or should it be downgraded to `private`?

jfeingold35 (approved):

packages/code-analyzer-core/src/code-analyzer.ts
stephen-carter-at-sf: Hmm... I see your point. If someone created a bunch of promises for the selectRules method with different promises, then indeed this would be a bit troublesome. I'll put in a comment for now so that we are aware of this edge case... so when we do open up to external clients... that we handle it then.

packages/code-analyzer-engine-api/src/engines.ts
stephen-carter-at-sf: I would like to keep this as the formal API and the other 3 I added as the convenience APIs for now. I was very reluctant in adding in the convenience methods initially because if we add a lot more events, we'd need a lot more convenience methods... but we'll see how things evolve. The formal API has the advantage of being able to build up events and hold on to them before then emitting them. Maybe we eventually just stick with these three. For now, no change needed.

packages/code-analyzer-core/src/code-analyzer.ts
stephen-carter-at-sf: Also, I think we have the same problem if someone ran the runRules multiple times as well via promise.all. Again... i'll just add in a comment.

PR #46, opened 2024-07-11
NEW (MetadataEngine) @W-16217126@ Scaffolding for metadata engine

ravipanguluri:
In this PR:
  
  * New plugin for meta-data based rules
  * Will create a new package called code-analyzer-metadata-engine
  * describeRule() will return a single rule: "PrivateMethodApiVersionRule"
  * runRules() will be a no-op for now

### DIFF
--- a/packages/code-analyzer-metadata-engine/src/plugin.ts
+++ b/packages/code-analyzer-metadata-engine/src/plugin.ts
@@ -1,0 +1,54 @@
+import {
+    ConfigObject,
+    DescribeOptions,
+    Engine,
+    EnginePluginV1,
+    EngineRunResults,
+    RuleDescription,
+    RuleType,
+    RunOptions,
+    SeverityLevel,
+} from "@salesforce/code-analyzer-engine-api";
+
+export class MetadataEnginePlugin extends EnginePluginV1 {
+
+    getAvailableEngineNames(): string[] {
+        return [MetadataEngine.NAME];
+    }
+
+    async createEngine(engineName: string, _config: ConfigObject): Promise<Engine> {
+        if (engineName === MetadataEngine.NAME) {
+            return new MetadataEngine()
+        }  else {
+            throw new Error(`Unsupported engine name: ${engineName}`);
+        }
+    }
+}
+
+export class MetadataEngine extends Engine {
+    static readonly NAME = "metadata"
+
+    getName(): string {
+        return MetadataEngine.NAME;
+    }
+
+    async describeRules(_describeOptions: DescribeOptions): Promise<RuleDescription[]> {
+        return [
+            {
+                name: "PrivateMethodAPIVersionRule",
+                severityLevel: SeverityLevel.High,
+                resourceUrls: [],
+                tags: ["Recommended", "Security"],
+                type: RuleType.Standard,
+                description: "Enforces a minimum API version for declaring private methods in abstract/virtual apex classes."
+            }
+        ];
+    }
+
+    async runRules(_ruleNames: string[], _runOptions: RunOptions): Promise<EngineRunResults> {
+        /*TODO: Implement logic for running rules */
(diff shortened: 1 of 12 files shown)


packages/code-analyzer-metadata-engine/src/plugin.ts
```diff
@@ -0,0 +1,55 @@
...
+    getAvailableEngineNames(): string[] {
+        return [MetadataEngine.NAME];
+    }
+
+    async createEngine(engineName: string, _config: ConfigObject): Promise<Engine> {
+        if (engineName === MetadataEngine.NAME) {
+            return new MetadataEngine()
+        }  else {
+            throw new Error(getMessage('CantCreateEngineWithUnknownEngineName', engineName));
+        }
```
stephen-carter-at-sf: Nitpick... no need for else statement because you are doing a return. So you can either do
```
if (engineName !== MetadtaEngine.NAME) {
    throw ...
}
return ...
```
or 
```
if (engineName === MetadtaEngine.NAME) {
    return ...
}
throw ...
```

packages/code-analyzer-metadata-engine/test/engine.test.ts
```diff
@@ -0,0 +1,90 @@
...
+import {
+    EngineRunResults,
+    RuleDescription,
+    RuleType,
+    SeverityLevel
+} from "@salesforce/code-analyzer-engine-api";
+import {
+    VIRTUAL_ABSTRACT_CLASS_API_RULE_DESCRIPTION,
+    VIRTUAL_ABSTRACT_CLASS_API_RULE_RESOURCE_URLS
+} from "./test-config";
```
stephen-carter-at-sf: To avoid all this test config stuff... please just put the constants here in this file. Only if they were to be used in multiple files would we want to do additional importing.

packages/code-analyzer-metadata-engine/test/engine.test.ts
```diff
@@ -0,0 +1,90 @@
...
+        ];
+        expect(rules_desc).toEqual(engineRules)
+    });
+
+    it("Check that running runRules() is a no-op that just returns an empty list of violations", async () => {
+        const runOptions = {workspace: testTools.createWorkspace([])}
+        const expResults: EngineRunResults = {violations: []}
+        const results = await engine.runRules(["PrivateMethodAPIVersionRule"], runOptions);
+        expect(results).toEqual(expResults)
+    })
```
stephen-carter-at-sf: Best not to even have this test since it isn't a part of the long term contract that we want. If something isn't implemented fully - I recommend just leaving it untouched which will then show up in the code coverage as something we still have to hit.

packages/code-analyzer-metadata-engine/test/engine.test.ts
```diff
@@ -0,0 +1,90 @@
...
+                description: VIRTUAL_ABSTRACT_CLASS_API_RULE_DESCRIPTION,
+                resourceUrls: VIRTUAL_ABSTRACT_CLASS_API_RULE_RESOURCE_URLS
+            }
+        ];
+        const engineRules: RuleDescription[] = await pluginEngine.describeRules({workspace: testTools.createWorkspace([])})
+        expect(engineRules).toStrictEqual(expEngineRules)
+    });
+
+    it('If I make an engine with an invalid name, it should throw an error with the proper error message', async () => {
+        await expect(enginePlugin.createEngine('OtherEngine', {})).rejects.toThrow("The MetadataEnginePlugin does not support creating an engine with name 'OtherEngine'.");
```
stephen-carter-at-sf: Consider using the getMessage(...) again here instead of locking the text into the test. This way if we update the message catalog - we won't need to update the tests.

stephen-carter-at-sf (approved):

PR #48, opened 2024-07-15
FIX(esint): @W-16113417@ Ignore files that eslint ignores and fix base config rule list

stephen-carter-at-sf:


packages/code-analyzer-eslint-engine/src/strategy.ts
```diff
@@ -164,16 +168,22 @@ export class LegacyESLintStrategy implements ESLintStrategy {
...
-        );
+        const allBaseRuleStatuses: Map<string, ESLintRuleStatus> =
+            await this.calculateRuleStatusesFor(candidateFiles, eslintForBaseRuleNameDiscovery);
+        const baseRulesThatAreOn: string[] = [];
+        for (const [ruleName, status] of allBaseRuleStatuses) {
+            if (status !== ESLintRuleStatus.OFF) {
+                baseRulesThatAreOn.push(ruleName);
+            }
+        }
+        return baseRulesThatAreOn;
```
stephen-carter-at-sf, line 186: I above used to have

```
            if (!this.ruleStatuses.has(ruleName) && !explicitlyTurnedOffRules.has(ruleName)) {
                this.ruleStatuses.set(ruleName, ESLintRuleStatus.OFF);
            }
```
but this was based off of the "recommended" list and therefore I was accidentally removing rules that could be on from the "all" base rule list.

So to fix this issue, all I had to do is switch the all base rule list to those rules from "all" which actually are still on. This works because typescript's all list actually removes certain rules from the eslint base config. So if javascript and typescript are both in the mix, they stay present... but if only typescript is in the mix, they correctly get omitted.

--
I caught this issue because some rules were running still even though they weren't in the rule list. Now they are explicitly turned off as they should be so that they don't run.

packages/code-analyzer-eslint-engine/test/test-data/legacyConfigCases/rules_DefaultConfig.goldfile.json
```diff
@@ -1999,6 +1999,18 @@
       "https://eslint.org/docs/latest/rules/no-alert"
     ]
   },
+  {
+    "name": "no-array-constructor",
```
stephen-carter-at-sf, line 2003: This was one of the rules that was accidentally omitted from the default case. It should only be omitted in the case when no javascript files are available and only typescript config is applied.

packages/code-analyzer-eslint-engine/test/test-data/legacyConfigCases/rules_DefaultConfig_NoJavascriptFilesInWorkspace.goldfile.json
```diff
@@ -1504,18 +1504,6 @@
       "https://eslint.org/docs/latest/rules/capitalized-comments"
     ]
   },
-  {
```
stephen-carter-at-sf, line 1507: We now correctly remove the javascript only rules now. I thought I had this correct before... but doing a more thorough check helped me find this. So now we correctly remove this as shown by the off value at https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/src/configs/all.ts

packages/code-analyzer-eslint-engine/src/strategy.ts
```diff
@@ -61,23 +61,24 @@ export class LegacyESLintStrategy implements ESLintStrategy {
             return this.ruleStatuses;
         }
 
+        const eslintOptions: ESLint.Options = this.createESLintOptions(BaseRuleset.RECOMMENDED);
+        const eslint: ESLint = new LegacyESLint(eslintOptions);
+        const filterFcn: AsyncFilterFnc<string> = createFilterFcn(eslint);
+
         const candidateFiles: string[] = this.userConfigEnabled() ?
```
jfeingold35, line 68: Absolute nitpick, but my preference for multi-line ternaries is :
```
x
? y
: z;
```

jfeingold35 (approved):

PR #49, opened 2024-07-16
NEW(eslint): @W-16253572@: Make eslint work in the background

stephen-carter-at-sf:
…so it doesn't block main thread

### DIFF
--- a/packages/code-analyzer-eslint-engine/src/strategy.ts
+++ b/packages/code-analyzer-eslint-engine/src/strategy.ts
@@ -6,6 +6,7 @@ import {BaseRuleset, LegacyBaseConfigFactory} from "./base-config";
 import {ESLintEngineConfig} from "./config";
 import {LogLevel} from "@salesforce/code-analyzer-engine-api";
 import {getMessage} from "./messages";
+import {Worker} from "node:worker_threads";
 
 export interface ESLintStrategy {
     calculateRuleStatuses(): Promise<Map<string, ESLintRuleStatus>>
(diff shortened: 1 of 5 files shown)


packages/code-analyzer-eslint-engine/worker-scripts/run-eslint.mjs
```diff
@@ -0,0 +1,12 @@
+// Note that workerData is the data that the parent thread passes to this worker
+import {parentPort, workerData} from "node:worker_threads";
+import {ESLint} from "eslint";
+
+const filesToScan = workerData.filesToScan;
+const eslintOptions = workerData.eslintOptions;
+
+const eslint = new ESLint(eslintOptions);
```
jfeingold35, line 8: Question: Why can this one be an unwrapped `ESLint` where the old one had to be our wrapper?

packages/code-analyzer-eslint-engine/src/strategy.ts
jfeingold35, line 1: Are the methods on `LegacyESLint` used anywhere now?

packages/code-analyzer-eslint-engine/worker-scripts/run-eslint.mjs
stephen-carter-at-sf: All the wrapping does is add a try/catch around the corresponding method being used. Here, the errors are returned by the worker. You'll notice the parent thread takes the error message that it receives and wraps it with:
```
worker.on('error', (err: Error) => reject(wrapESLintError(err, 'ESLint.lintFiles', eslintOptions)));
```

packages/code-analyzer-eslint-engine/src/strategy.ts
stephen-carter-at-sf: Yes. The other methods I still do in the main thread. I only moved the lintFiles to the background thread because it is the only heavy handed one that seems to have a syncronous loop.

stephen-carter-at-sf: And we don't want to create a background thread if not needed... especially for the rules command (which we want to keep as fast as possible).

jfeingold35 (approved):

PR #50, opened 2024-07-16
NEW(pmd): @W-15884535@: Set up scalfolding for pmd and cpd engines

stephen-carter-at-sf:


packages/code-analyzer-pmd-engine/src/messages.ts
```diff
@@ -0,0 +1,15 @@
+import {getMessageFromCatalog} from "@salesforce/code-analyzer-engine-api";
+
+const MESSAGE_CATALOG : { [key: string]: string } = {
+    CantCreateEngineWithUnknownEngineName:
+        `The PmdCpdEnginesPlugin does not support creating an engine with name '%s'.`
```
jag-j, line 5: Do you use the '***' markers in core as well or is it just at the CLI level?

packages/code-analyzer-pmd-engine/src/engines.ts
```diff
@@ -0,0 +1,46 @@
+import {
+    DescribeOptions,
+    Engine,
+    EngineRunResults,
+    LogLevel,
+    RuleDescription,
+    RunOptions
+} from "@salesforce/code-analyzer-engine-api";
+
+export class CpdEngine extends Engine {
```
jag-j, line 10: @stephen-carter-at-sf @ravipanguluri - This is different from how Ravi has implemented the two engines right? He created two directories at the packages level for the engines (regex and metadata). Why is this different?

packages/code-analyzer-pmd-engine/src/messages.ts
stephen-carter-at-sf: Right now we are just doing that at the CLI level. We had a meeting with Juliet yesterday and she will simply add in comments for each message she has approved in the this repo. So none of these have been reviewed by her yet - but they are all on her todo list. Soon you'll see things like:

```
    CantCreateEngineWithUnknownEngineName: // Approved on xx-xx-2024
        `The PmdCpdEnginesPlugin does not support creating an engine with name '%s'.`
```

packages/code-analyzer-pmd-engine/src/engines.ts
stephen-carter-at-sf: The regex and metadata plugins are going to be very different in the end. Therefore it doesn't make sense for them to share the same plugin in my opinion. But the CPD and PMD tools share nearly all of the same guts and dependencies behind the scenes (both depend on pmd jars)... thus this single plugin will supply both.

jfeingold35 (approved):

PR #51, opened 2024-07-17
CHANGE (regex): @W-16150675@ Updates to regex engine

ravipanguluri:
In this PR:
*Updates to regex engine


packages/code-analyzer-regex-engine/src/rule.ts
```diff
@@ -0,0 +1,8 @@
+interface Rule {
+    regex: RegExp;
+    fileExtensions: string[];
+    message: string;
+    description: string;
+}
+
+export type RuleMap = { [ruleName: string]: Rule };
```
stephen-carter-at-sf: Move this to engine.ts file since it's only used there.

Also... Record<string, Rule> might be easier than { [ruleName: string]: Rule}

packages/code-analyzer-regex-engine/src/engine.ts
```diff
@@ -0,0 +1,99 @@
...
+import fs from "node:fs";
+import {getColumnNumber, getLineNumber, getNewlineIndices} from "./utils";
+import {getMessage} from "./messages";
+import {RuleMap} from "./rule";
+
+const APEX_CLASS_FILE_EXT: string = ".cls"
+
+export class RegexEngine extends Engine {
+    static readonly NAME = "regex"
+    private regexRules: RuleMap = {};
```
stephen-carter-at-sf: readonly

packages/code-analyzer-regex-engine/src/engine.ts
```diff
@@ -0,0 +1,99 @@
...
+    constructor() {
+        super();
+        this.regexRules = {
+            "TrailingWhitespaceRule": {
+                regex: /[ \t]+((?=\r?\n)|(?=$))/g,
+                fileExtensions: [APEX_CLASS_FILE_EXT],
+                description: getMessage('TrailingWhitespaceRuleDescription'),
+                message: getMessage('TrailingWhitespaceRuleDescription')
+            }
+        }
```
stephen-carter-at-sf: Just move this to line 21. No need for constructor then.

packages/code-analyzer-regex-engine/src/messages.ts
```diff
@@ -0,0 +1,21 @@
...
+
+const MESSAGE_CATALOG : { [key: string]: string } = {
+    CantCreateEngineWithUnknownEngineName:
+        `The RegexEnginePlugin does not support creating an engine with name '%s'.`,
+
+    TrailingWhitespaceRuleDescription:
+        "Detects trailing whitespace (tabs or spaces) at the end of lines of code and lines that are only whitespace.",
+
+    PrivateMethodApiVersionRuleDescription:
+        "For API versions 60.0 and below, declaring a private method in an abstract or virtual class will override a method with the same signature in a child class. Please update your API version to resolve the issue."
```
stephen-carter-at-sf: Remove this.

packages/code-analyzer-regex-engine/src/engine.ts
```diff
@@ -0,0 +1,99 @@
...
+
+    /*TODO: swap hardcoding of rule for addRules() method that processes a config file*/
+    constructor() {
+        super();
+        this.regexRules = {
+            "TrailingWhitespaceRule": {
+                regex: /[ \t]+((?=\r?\n)|(?=$))/g,
+                fileExtensions: [APEX_CLASS_FILE_EXT],
+                description: getMessage('TrailingWhitespaceRuleDescription'),
+                message: getMessage('TrailingWhitespaceRuleDescription')
```
stephen-carter-at-sf: Consider making a generic message for any type of regex violation found... instead of having a violation message be attached to the this particular rule.

packages/code-analyzer-regex-engine/src/engine.ts
```diff
@@ -0,0 +1,99 @@
...
+            }
+        }
+    }
+
+    getName(): string {
+        return RegexEngine.NAME;
+    }
+
+    async describeRules(_describeOptions: DescribeOptions): Promise<RuleDescription[]> {
```
stephen-carter-at-sf, line 40: Add in a TODO so that we don't return rules that are not applicable to the files in the workspace if provided.

packages/code-analyzer-regex-engine/src/engine.ts
```diff
@@ -0,0 +1,99 @@
...
+        return [
```
stephen-carter-at-sf, line 41: Eventually, we don't want to hard code this... but extract it from the rulesMap or whatever above. So for now, just put in a todo.

packages/code-analyzer-regex-engine/src/engine.ts
```diff
@@ -0,0 +1,99 @@
...
+            {
+                name: "TrailingWhitespaceRule",
+                severityLevel: SeverityLevel.Low,
+                type: RuleType.Standard,
+                tags: ["Recommended", "CodeStyle"],
+                description: "Detects trailing whitespace (tabs or spaces) at the end of lines of code and lines that are only whitespace.",
```
stephen-carter-at-sf: Reuse message here

packages/code-analyzer-regex-engine/src/engine.ts
```diff
@@ -0,0 +1,99 @@
...
+            const rulesToRun = ruleNames.filter(rule => this.shouldScanFile(file, rule));
+            const violationPromises = rulesToRun.map(async (rule) => await this.scanFile(file, rule));
+            violations.push(...(await Promise.all(violationPromises)).flat());
+        });
+
+        await Promise.all(filePromises);
+
+        return {
+            violations: violations
+        };
```
stephen-carter-at-sf: Consider having a runRulesForFile(file: string, ruleNames:string[]): Promise<Violation[]>

then this method will simplify to:

```
const fullFileList: string[] = await runOptions.workspace.getExpandedFiles();
return {
    violations: (await Promise.all(fullFileList.map(file => runRulesForFile(file, ruleNames)))).flat()
};

something to this effect.

```

packages/code-analyzer-regex-engine/src/engine.ts
```diff
@@ -0,0 +1,99 @@
...
+                startLine: getLineNumber(match.index, newlineIndexes),
+                startColumn: getColumnNumber(match.index, newlineIndexes),
+                endLine: getLineNumber(match.index + match[0].length, newlineIndexes),
+                endColumn: getColumnNumber(match.index + match[0].length, newlineIndexes)
+            }
+            violations.push({
+                ruleName: ruleName,
+                codeLocations: [codeLocation],
+                primaryLocationIndex: 0,
+                message: this.regexRules[ruleName].message
```
stephen-carter-at-sf: Here is where we might have a more generic message in the future.

packages/code-analyzer-regex-engine/src/utils.ts
```diff
@@ -0,0 +1,29 @@
+import os from "node:os";
+
+export function getColumnNumber(charIndex: number, newlineIndexes: number[]): number {
```
stephen-carter-at-sf: I suggest moving these to keep them as implementation detail for the one file that uses it.

packages/code-analyzer-regex-engine/test/engine.test.ts
```diff
@@ -40,42 +46,73 @@ describe('Regex Engine Tests', () => {
...
+    it("If trailing whitespace rule is run on an Apex class without trailing whitespace ensure there are no erroneous violations", async () => {
+        const filePath = path.resolve("test", "test-data", "apexClassWhitespace", "3_apexClassWithoutWhitespace", "myOuterClass.cls")
+        const runOptions: RunOptions = {workspace: testTools.createWorkspace([filePath])}
+        const runResults: EngineRunResults = await engine.runRules(ruleNames, runOptions);
+        const expViolations: Violation[] = []
+        expect(runResults.violations).toStrictEqual(expViolations);
+
+    });
+
+    it("Ensure execute() can be called on a list Apex classes and properly emits errors", async () => {
```
stephen-carter-at-sf: No execute left

packages/code-analyzer-regex-engine/test/engine.test.ts
```diff
@@ -40,42 +46,73 @@ describe('Regex Engine Tests', () => {
...
+        expect(runResults.violations).toHaveLength(EXPECTED_VIOLATION_3.length)
+        expect(runResults.violations).toContainEqual(EXPECTED_VIOLATION_3[0])
+        expect(runResults.violations).toContainEqual(EXPECTED_VIOLATION_3[1])
+        expect(runResults.violations).toContainEqual(EXPECTED_VIOLATION_3[2])
+        expect(runResults.violations).toContainEqual(EXPECTED_VIOLATION_3[3])
+    });
 });
 
-describe('RegexEnginePlugin Tests' , () => {
+describe('Plugin Tests' , () => {
```
stephen-carter-at-sf: Move these to a plugin.test.ts file instead... because eventually we will be adding in configuration and validation that takes place at this level.

packages/code-analyzer-regex-engine/test/test-config.ts
```diff
@@ -4,12 +4,19 @@ import { CodeLocation, Violation} from "@salesforce/code-analyzer-engine-api";
...
+export const FILE_LOCATION_2 = path.resolve(__dirname,  "test-data", "apexClassWhitespace", "2_apexClasses", "myClass.cls")
+
+
 export const TRAILING_WHITESPACE_RULE_NAME = "TrailingWhitespaceRule"
 export const TRAILING_WHITESPACE_RULE_MESSAGE = "Detects trailing whitespace (tabs or spaces) at the end of lines of code and lines that are only whitespace.";
 export const TRAILING_WHITESPACE_RESOURCE_URLS = []
 
+export const VIRTUAL_ABSTRACT_CLASS_API_RULE_NAME = "PrivateMethodAPIVersionRule"
+export const VIRTUAL_ABSTRACT_CLASS_API_RULE_MESSAGE = "For API versions 60.0 and below, declaring a private method in an abstract or virtual class will override a method with the same signature in a child class. Please update your API version to resolve the issue."
+export const VIRTUAL_ABSTRACT_CLASS_API_RULE_DESCRIPTION = "Enforces a minimum API version for declaring private methods in abstract/virtual apex classes."
```
stephen-carter-at-sf: Irrelevant to Regex

stephen-carter-at-sf (changes requested):

packages/code-analyzer-regex-engine/src/engine.ts
```diff
@@ -0,0 +1,128 @@
...
+
+const APEX_CLASS_FILE_EXT: string = ".cls"
+
+interface Rule {
+    regex: RegExp;
+    fileExtensions: string[];
+    description: string;
+}
+
+export type RuleMap = Record<string, Rule>;
```
stephen-carter-at-sf: No need to export this is there?

Not in this review... but soon... we might want to make this a Map<string,RegexRule> and then prepopulate it with the hard coded rules... and then we'll be adding to this list when processing config.

packages/code-analyzer-regex-engine/src/engine.ts
```diff
@@ -0,0 +1,128 @@
...
+import {getMessage} from "./messages";
+import os from "node:os";
+
+const APEX_CLASS_FILE_EXT: string = ".cls"
+
+interface Rule {
+    regex: RegExp;
+    fileExtensions: string[];
+    description: string;
+}
```
stephen-carter-at-sf, line 22: Not in this review... but soon... we'll want to add "name" here, and severity, and tags. And I would recommend we call it RegexRule.

packages/code-analyzer-regex-engine/src/engine.ts
```diff
@@ -0,0 +1,128 @@
...
+                tags: ["Recommended", "CodeStyle"],
+                description: getMessage("TrailingWhitespaceRuleDescription"),
+                resourceUrls: []
+            }
+        ];
+    }
+
+    async runRules(ruleNames: string[], runOptions: RunOptions): Promise<EngineRunResults> {
+        const fullFileList: string[] = await runOptions.workspace.getExpandedFiles();
+        const ruleRunPromises = fullFileList.map(file => this.runRulesForFile(file, ruleNames))
```
stephen-carter-at-sf: nitpick... try to consistently use `;` or not. Also... try to always provide type if you can. It helps when others read the code to immediately know that 
const ruleRunPromises: Promise<Violation[]>[]
is what we are looking at.

packages/code-analyzer-regex-engine/src/messages.ts
```diff
@@ -0,0 +1,21 @@
...
+
+const MESSAGE_CATALOG : { [key: string]: string } = {
+    CantCreateEngineWithUnknownEngineName:
+        `The RegexEnginePlugin does not support creating an engine with name '%s'.`,
+
+    TrailingWhitespaceRuleDescription:
+        "Detects trailing whitespace (tabs or spaces) at the end of lines of code and lines that are only whitespace.",
+
+    RuleViolationMessage:
+        "A violation for %s was found in %s on line %d starting at column %d and ending on line %d at column %d"
```
stephen-carter-at-sf: All the location information will already be available in the violation ... so I recommend instead adding in the regular expression, the rule name, and the rule description to help make up the violation message. So maybe we do something like:

```
    RuleViolationMessage:
        `A match of the regular expression %s was found for rule '%s': %s`
```

where we fill in the holes with the '/' + regex.source +'/' + (whatever else is needed to reform it), the rule name, and the rule description. No need to add in code location information because that'll show up with the violation (whereas the regex and description wont).

So for our hard coded rule it will end up looking like:
```
A match of the regular expression /[ \t]+((?=\r?\n)|(?=$))/g was for rule 'NoTrailingWhitespace': Detects trailing whitespace (tabs or spaces) at the end of lines of code and lines that are only whitespace.
```

packages/code-analyzer-regex-engine/src/engine.ts
```diff
@@ -0,0 +1,128 @@
...
+    };
+
+    getName(): string {
+        return RegexEngine.NAME;
+    }
+    /*TODO: Use describeOptions so that rules not applicable to the workspace aren't returned and rules extracted from a data structure instead of being hardcoded */
+    async describeRules(_describeOptions: DescribeOptions): Promise<RuleDescription[]> {
+        return [
+            {
+                name: "TrailingWhitespaceRule",
```
stephen-carter-at-sf: I think the rule name should be either NoTrailingWhitespace (which is my vote) or just TrailingWhitespace. I don't think we should have the word "Rule" in our rule names.

stephen-carter-at-sf (changes requested):

packages/code-analyzer-regex-engine/test/engine.test.ts
stephen-carter-at-sf: Still have this to fix.

stephen-carter-at-sf (approved):

PR #52, opened 2024-07-17
FIX (RetireJS) @W-16253670@ Reasons for ZIP failure are properly contextualized.

jfeingold35:


packages/code-analyzer-retirejs-engine/src/zip-decorator.ts
```diff
@@ -0,0 +1,44 @@
+import StreamZip from 'node-stream-zip';
+import {getMessage} from './messages';
+
+export class DecoratedStreamZip {
```
jfeingold35, line 4: I tried initially for this to extend `StreamZip.StreamZipAsync` instead of enclosing it, but I kept getting a weird TypeError. This was the best workaround I could find.

packages/code-analyzer-retirejs-engine/src/zip-decorator.ts
```diff
@@ -0,0 +1,44 @@
...
+    private readonly streamZip: StreamZip.StreamZipAsync;
+    private readonly file: string;
+
+    public constructor(config: StreamZip.StreamZipOptions) {
+        this.streamZip = new StreamZip.async(config);
+        this.file = config.file || /* istanbul ignore next - Not worth testing */ '';
```
stephen-carter-at-sf: Why not instead just do 
```
this.file = config.file!;
```
We wouldn't pass in undefined would we? If not, then the `!` just asserts that it is not undefined or null... making compiler happy.

packages/code-analyzer-retirejs-engine/src/zip-decorator.ts
stephen-carter-at-sf: Yeah... formally the Decorator pattern actually takes in an object into its constructor, implements the same interface as the object it takes in... and then delegates all its responsibilities to the composed delegate. But this is practically the same thing without actually implementing the interface formally. There a many flavors of this..... So I like this.

stephen-carter-at-sf (approved):

PR #54, opened 2024-07-29
NEW (retire-js): @W-16263157@: Enhance retire-js engine to not return rules if no target files are in the provided workspace

stephen-carter-at-sf:


packages/code-analyzer-retirejs-engine/src/engine.ts
```diff
@@ -11,11 +11,24 @@ import {
...
+    ZIPPED_FILE_MARKER
+} from "./executor";
 import {Finding, Vulnerability} from "retire/lib/types";
 import {getMessage} from "./messages";
+import path from "node:path";
+
+// To speed up execution, we will only target files with extension among EXTENSIONS_TO_TARGET
+// that don't live under one of the FOLDERS_TO_SKIP
+const EXTENSIONS_TO_TARGET = [...JS_EXTENSIONS, '.resource', '.zip'];
+const FOLDERS_TO_SKIP = ['node_modules', 'bower_components'];
```
stephen-carter-at-sf, line 31: The code to reduce the files down to the target file list was moved from the executor to the engine. Now the executor does not deal with workspaces at all... but instead just deals with the targeted file list. This way we can use this information to help the engine determine whether or not to return the rules.

packages/code-analyzer-retirejs-engine/src/engine.ts
```diff
@@ -59,16 +73,30 @@ export class RetireJsEngine extends Engine {
         return RetireJsEngine.NAME;
     }
 
-    async describeRules(_describeOptions: DescribeOptions): Promise<RuleDescription[]> {
+    async describeRules(describeOptions: DescribeOptions): Promise<RuleDescription[]> {
+        if (describeOptions.workspace && (await this.getTargetFiles(describeOptions.workspace)).length === 0) {
+            return [];
+        }
```
stephen-carter-at-sf, line 79: This is basically the enhancement. If no files are targeted, then we return zero rules. For efficiency, the calculated targetedFiles list is cached based on the workspace id so that runRules doesn't need to recalculate anything if the workspace is the same.

jfeingold35 (approved):

PR #55, opened 2024-07-30
NEW (regex) @W-16267963@: Add customization to regex engine

ravipanguluri:
In this PR:

A/C
* Users will be able to add custom rules to the engine via a custom configuration format (via code-analyzer-config.yml file)
* describeRules() will ingest rules and return only rules relevant to the user's workspace
* runRules() will run engine's default rule(s) and custom rules returning all violations in a given workspace/directory


packages/code-analyzer-regex-engine/src/config.ts
```diff
@@ -0,0 +1,209 @@
...
+import {getMessage} from "./messages";
+
+const APEX_CLASS_FILE_EXT: string = ".cls"
+
+export interface Rule {
+    regex: RegExp;
+    file_extensions: string[];
+    description: string;
+    violation_message: string;
+}
```
stephen-carter-at-sf: Did you mean to make this an interface instead of a type?

I recommend this to be a type and I would also recommend calling this RegexRule since it is exported (because the name Rule is kind of generic and overloaded).

packages/code-analyzer-regex-engine/src/config.ts
```diff
@@ -0,0 +1,209 @@
...
+export type RuleMap = Record<string, Rule>;
+
+export type RegexEngineConfig = {
+    disable_engine: boolean
```
stephen-carter-at-sf: The disabling of an engine is done at the core level so you don't need to ever worry about the disable_engine flag or even know it exists at the engine level. See https://github.com/forcedotcom/code-analyzer-core/pull/56

So please remove this.

packages/code-analyzer-regex-engine/src/config.ts
```diff
@@ -0,0 +1,209 @@
...
+    config_root: string
+    rules: RuleMap
+}
+
+export const DEFAULT_CONFIG: RegexEngineConfig = {
+    disable_engine: false,
+    config_root: process.cwd(),
```
stephen-carter-at-sf: You aren't using config_root anywhere with this engine... so please remove it.  If we need it in the future, then we'll make sure to add it in at that time.

packages/code-analyzer-regex-engine/test/plugin.test.ts
```diff
@@ -33,11 +99,418 @@ describe('RegexEnginePlugin Tests' , () => {
...
+
+    it('When non-default config_root is provided to createEngine, then it is set on the config', async () => {
+        const engine: RegexEngine = (await enginePlugin.createEngine('regex', {config_root: __dirname})) as RegexEngine;
+        expect(engine.getConfig().config_root).toEqual(__dirname);
+    });
+
+    it('When config_root is invalid, then createEngine errors', async () => {
+        await expect(enginePlugin.createEngine('regex', {config_root: 3})).rejects.toThrow(
+            getMessageFromCatalog(SHARED_MESSAGE_CATALOG, 'ConfigValueMustBeOfType', 'config_root', 'string', 'number'));
+    });
```
stephen-carter-at-sf: This is all unnecessary since this engine doesn't need the config_root that is provided from core. If we need it in the future, then we'll add it at that time.

packages/code-analyzer-regex-engine/src/config.ts
```diff
@@ -0,0 +1,209 @@
...
+        this.regexConfigObj = rawConfig
+    }
+    /*TODO: move/create a generic version of this function to extract arrays at API level (allows me to get rid of this.regexConfigObj) */
+    extractObjectArray(fieldName: string, defaultValue?: object[]): object[] | undefined {
+        return isFieldUndefined(fieldName, this.regexConfigObj) ? defaultValue :
+            RegexValueValidator.validateObjectArray(this.regexConfigObj[fieldName], this.getFullFieldName(fieldName));
+    }
+
+    extractRules(): RuleMap {
+        const rulesFieldName = 'rules';
```
stephen-carter-at-sf: I thought we were going to call this custom_rules?

packages/code-analyzer-regex-engine/src/config.ts
```diff
@@ -0,0 +1,209 @@
...
+
+export const DEFAULT_CONFIG: RegexEngineConfig = {
+    disable_engine: false,
+    config_root: process.cwd(),
+    rules: {
+        "NoTrailingWhitespace": {
+            regex: /[ \t]+((?=\r?\n)|(?=$))/g,
+            file_extensions: [APEX_CLASS_FILE_EXT],
+            description: getMessage('TrailingWhitespaceRuleDescription'),
+            violation_message: getMessage('RuleViolationMessage', /[ \t]+((?=\r?\n)|(?=$))/g.toString(), "NoTrailingWhitespace", getMessage('TrailingWhitespaceRuleDescription'))
```
stephen-carter-at-sf: I thought we decided for our built in rules, that we were going to have better violation messages and that the generic violation message would only be for custom rules that didn't define a violation_message?

Please update this to something that says something like "Trailing whitespace found" or whatever.

packages/code-analyzer-regex-engine/src/config.ts
```diff
@@ -0,0 +1,209 @@
...
+    /* TODO: Extract to engine API level so there's no code repetition */
+    extractExtensionsValue(fieldName: string, defaultValue: string[]): string[] {
+        const fileExts: string[] = this.extractStringArray(fieldName, defaultValue)!;
+        return fileExts.map((fileExt, i) => validateStringMatches(
+            RuleConfigValueExtractor.FILE_EXT_PATTERN, fileExt, `${this.getFullFieldName(fieldName)}[${i}]`));
+    }
+
+}
+
+class RegexValueValidator extends ValueValidator {
```
stephen-carter-at-sf: No need to extend ValueValidator in this case because ValueValidator only contains static methods. In fact, since these validation methods won't be reused... you really don't even need a RegexValueValidator class... you could just make these validate methods functions.

packages/code-analyzer-regex-engine/src/config.ts
```diff
@@ -0,0 +1,209 @@
...
+}
+
+
+class RuleConfigValueExtractor extends ConfigValueExtractor {
+    private static readonly FILE_EXT_PATTERN: RegExp = /^[.][a-zA-Z0-9]+$/;
+    /*If methods can't/shouldn't be moved then maybe the configObj for ConfigValueExtractor could be changed to protected so this instance variable isn't needed */
+    private readonly regexConfigObj: ConfigObject;
+
+    constructor(rawConfig: ConfigObject, index: number) {
+        super(rawConfig, `engines.regex.rules[${index}]`);
```
stephen-carter-at-sf: You should probably keep these field names as constants somewhere and use them... because I'd like for the "rules" to be renamed to "custom_rules" and when you rename, you'll need to be super careful not to miss one of these spots.

packages/code-analyzer-regex-engine/src/config.ts
```diff
@@ -0,0 +1,209 @@
...
+    }
+}
+
+
+class RuleConfigValueExtractor extends ConfigValueExtractor {
+    private static readonly FILE_EXT_PATTERN: RegExp = /^[.][a-zA-Z0-9]+$/;
+    /*If methods can't/shouldn't be moved then maybe the configObj for ConfigValueExtractor could be changed to protected so this instance variable isn't needed */
+    private readonly regexConfigObj: ConfigObject;
+
+    constructor(rawConfig: ConfigObject, index: number) {
```
stephen-carter-at-sf: Why not pass in the entire field string here instead of just the index. The parent RegexEngineConfigValueExtractor already had the engines.regex piece. So I would have anticipated this to be called as:
```
new RuleConfigValueExtractor(ruleObj, `${this.getFullFieldName('custom_rules')}[${i}]`)
```
over on line 79

packages/code-analyzer-regex-engine/src/config.ts
```diff
@@ -0,0 +1,209 @@
...
+
+        if (!rulesArray){
+            return {}
+        }
+
+        const customRules: RuleMap = {}
+
+        rulesArray.forEach((ruleObj, i) => {
+            const ruleConfigValueExtractor: RuleConfigValueExtractor = new RuleConfigValueExtractor(ruleObj, i)
+            const [ruleName, rule] = ruleConfigValueExtractor.extractRule()
```
stephen-carter-at-sf: This seems a bit away from the current extractor pattern. I would have anticipated that the RuleConfigValueExtractor would have extract methods for each of the fields that were on the custom_rule object. So I would have expected something like:
```
const ruleName = ruleConfigValueExtractor.extractName();
customRules[ruleName] = {
   regex: ruleConfigValueExtractor.extractRegex('regex'),
   description: ValueValidator.validateString(this.configObj[fieldName], this.getFullFieldName(fieldName)), // TODO:  Add in extractRequiredString to engine-api to make this easier
   ... // etc
}

Comment (ravipanguluri) [reply packages/code-analyzer-regex-engine/src/config.ts]: Ok I see. The way that I interpreted this was that extractRules() method at that level has the responsibility of extracting the rule array, then a different method has the responsibility of extracting each rule. This is similar to the pattern in the ESLint engine where you extract the file extensions. Just want to make sure I fully understand the pattern you were going for.

Comment (stephen-carter-at-sf) [reply packages/code-analyzer-regex-engine/src/config.ts]: We'll discuss this more during our meeting later.

File: packages/code-analyzer-regex-engine/src/config.ts [hunk 11]
```diff
@@ -0,0 +1,115 @@
...
+        }
+        seenFlags.add(flag);
+    }
+
+    if (seenFlags.has('u') && seenFlags.has('v')){
+        return false;
+    }
+
+    return true;
+}
```
stephen-carter-at-sf: Instead of all this separate validation on the modifiers... couldn't you just validate the regular expression in one go with something like:

```
const match = rawRegexString.match(/^\/(.*?)\/([gimsuvy]*)$/);
    if (!match) {
        throw new Error(`${rawRegexString} is not a valid regex format`); // dummy placeholder message
    }
    const pattern = match[1];
    const flags = match[2];
    return new RegExp(pattern, flags);
```

ravipanguluri: One thing:

*   This would not catch the case of a user having repeated modifiers, and would instead dispatch that error to the Regexp constructor. Which is likely fine, but does change the intended behavior.

For instance, with this type of a regex, if I pass in a rule with the regex /TODO:/lpr then I get the 'InvalidRegexString' message. Whereas, if I hit a repeated modifiers case with something like /TODO:/gig, I get the RegExp constructor's invalid flag message instead. So in both cases we have a modifier issue, but the messages are different. I thought of doing it this way initially, but my worry is that this isn't the most intuitive way for users to interact with the engine even though it makes the code shorter.

packages/code-analyzer-regex-engine/src/config.ts
```diff
@@ -0,0 +1,96 @@
...
+export const DEFAULT_CONFIG: RegexEngineConfig = {
+    rules: {
+        "NoTrailingWhitespace": {
+            regex: /[ \t]+((?=\r?\n)|(?=$))/g,
+            file_extensions: [APEX_CLASS_FILE_EXT],
+            description: getMessage('TrailingWhitespaceRuleDescription'),
+            violation_message: getMessage('TrailingWhitespaceRuleMessage')
+        }
+    }
+};
```
stephen-carter-at-sf: This is very close to what you have the users provide. I think we should just make this the structure that users provide. We should propose this to the team (but rename rules to custom_rules). 

So instead of
```
engines:
  regex:
    custom_rules:
      - name: "RuleName1"
        regex: "/somePattern1/g"
        description: "some rule description 1"
        violation_message:  "some violation message 1"   # optional
      - name: "RuleName2"
        regex: "/somePattern2/g"
        description: "some rule description 2"
        violation_message:  "some violation message 2"   # optional
```

I propose we have

```
engines:
  regex:
    custom_rules:
      RuleName1:
        regex: "/somePattern1/g"
        description: "some rule description 1"
        violation_message:  "some violation message 1"   # optional
      RuleName2:
        regex: "/somePattern2/g"
        description: "some rule description 2"
        violation_message:  "some violation message 2"   # optional
```

This is because it is the same format that they would supply rule overrides already. So putting it all together would look like:

```
engines:
  regex:
    custom_rules:
      RuleName1:
        regex: "/somePattern1/g"
        description: "some rule description 1"
        violation_message:  "some violation message 1"   # optional
      RuleName2:
        regex: "/somePattern2/g"
        description: "some rule description 2"
        violation_message:  "some violation message 2"   # optional

rules:
  regex:
    RuleName1:
      severity: 1
      tags: ['Recommended', 'Custom']
    RuleName2:
      severity: 4
      tags: ['Custom']
```

packages/code-analyzer-regex-engine/test/plugin.test.ts
```diff
@@ -33,11 +101,329 @@ describe('RegexEnginePlugin Tests' , () => {
...
+                "NoTodos": {
+                    description: BASELINE_NO_TODOS_RULE.description,
+                    file_extensions: BASELINE_NO_TODOS_RULE.file_extensions
+                }
+            }
+        }
+        await expect(enginePlugin.createEngine("regex", rawConfig)).rejects.toThrow(getMessageFromCatalog(SHARED_MESSAGE_CATALOG, 'ConfigValueMustBeOfType', 'engines.regex.custom_rules.NoTodos.regex', 'string', 'undefined'));
+    });
+
+    it("If rule name is empty, ensure proper error is emitted", async () => {
```
ravipanguluri, line 191: I tested reading in a YAML file with the empty string as a key value into JS and it worked. So I added this test to defend against that case.

packages/code-analyzer-regex-engine/src/config.ts
```diff
@@ -0,0 +1,100 @@
...
+const FILE_EXT_PATTERN: RegExp = /^[.][a-zA-Z0-9]+$/;
+
+export function validateAndNormalizeConfig(rawConfig: ConfigObject): RegexEngineConfig {
+    const valueExtractor: ConfigValueExtractor = new ConfigValueExtractor(rawConfig, REGEX_ENGINE_ROOT);
+    const customRulesExtractor: ConfigValueExtractor = valueExtractor.extractObjectAsExtractor(RULES_FIELD_NAME);
+    const customRules: RegexRuleMap = {};
+    const keys: string[] = Object.keys(customRulesExtractor.getObject())
+
+    for (const ruleName of keys) {
+        if (!ruleName){
```
stephen-carter-at-sf: Instead of just looking for empty... should we enforce things like... you can't have spaces in your rule names, etc. 

We might want to instead do a regex match on what the rule name format should be. Something like /^[A-Za-z@][A-Za-z_0-9@\-\/]*/

ravipanguluri: Hmm that is a good point that I hadn't thought of. I can make that update.

packages/code-analyzer-regex-engine/src/config.ts
stephen-carter-at-sf: I don't think its worth having a ton more code to just improve the error message in this case. Most of the time users will write a valid regular expression. Only when they do not would we provide an error message. If the error message is generic and says something as simply as - invalid regular expression pattern... that is enough for them to get started. Even better if we say invalid because it doesn't match this pattern... then they can figure things out. To your point of having repeated modifiers resulting in an odd error message from RegExp constructor... as long as an error is thrown - I think this is fine.

But i'll let you decide at this point.

ravipanguluri: I ended up liking having the regex constructor handle more of the edge cases if you look at the latest version because I realized that most of the messages are pretty good.

packages/code-analyzer-regex-engine/src/config.ts
```diff
@@ -0,0 +1,100 @@
...
+export type RegexRule = {
+    regex: RegExp;
+    file_extensions: string[];
+    description: string;
+    violation_message: string;
+}
+export type RegexRuleMap = Record<string, RegexRule>;
+
+export type RegexEngineConfig = {
+    rules: RegexRuleMap
```
stephen-carter-at-sf: I would call this custom_rules again now that we have a match between the raw config and this normalized config. This way we can document this whole section of the code as the config documentation.

stephen-carter-at-sf (approved):

packages/code-analyzer-regex-engine/src/config.ts
ravipanguluri: I can change it for now, but I'm a little hesitant about locking this in because with my current design this naming change implies that default rules are part of the custom rules.

PR #56, opened 2024-07-30
NEW (core): @W-16360293@: Add in ability to disable engines

stephen-carter-at-sf:


packages/code-analyzer-core/src/config.ts
```diff
@@ -193,7 +194,11 @@ function extractEnginesValue(configValueExtractor: engApi.ConfigValueExtractor):
         return DEFAULT_CONFIG.engines;
     }
     for (const [engineName, settingsForEngine] of Object.entries(engineSettingsObj)) {
-        engApi.ValueValidator.validateObject(settingsForEngine, `${FIELDS.ENGINES}.${engineName}`);
+        const settingsForEngineObj: object = engApi.ValueValidator.validateObject(settingsForEngine, `${FIELDS.ENGINES}.${engineName}`);
+        if (FIELDS.DISABLE_ENGINE in settingsForEngineObj) {
```
jfeingold35, line 198: Why do we validate `disableEngine` separately?

stephen-carter-at-sf: I'm not sure I understand the question. What do you mean? After validating that "engines.<engine_name>" is an object. We then want to validate that if "engines.<engine_name>.disable_engine" is a boolean.

But I am currently working with Ravi on ways we can clean up his config validation, and so I am currently working on some improvements to our validation logic - that will hopefully clean this up rather nicely as well. So either we can submit this and then i'll clean this up. Or just wait for one more PR to come out in a little while and we'll get those changes in and then I'll update this PR to make this a little cleaner.

Basically the idea will be to have the parent ConfigValueExtractor have something like a extractConfigObject(fieldName) which returns yet another ConfigValueExtractor so that we can continue validating into the next level down. Stay tuned.

jfeingold35 (approved):

PR #57, opened 2024-07-30
CHANGE (core): @W-16267963@: Improve config value extractors and validators

stephen-carter-at-sf:


packages/code-analyzer-core/src/config.ts
```diff
@@ -166,8 +168,9 @@ function extractRuleOverrideFor(ruleOverridesObj: object, engineName: string, ru
             `${FIELDS.RULES}.${engineName}.${ruleName}.${FIELDS.SEVERITY}`);
     }
     if (FIELDS.TAGS in ruleOverrideObj ) {
-        extractedValue.tags = engApi.ValueValidator.validateStringArray(ruleOverrideObj[FIELDS.TAGS],
-            `${FIELDS.RULES}.${engineName}.${ruleName}.${FIELDS.TAGS}`);
+        extractedValue.tags = engApi.ValueValidator.validateArray(ruleOverrideObj[FIELDS.TAGS],
+            `${FIELDS.RULES}.${engineName}.${ruleName}.${FIELDS.TAGS}`,
+            ValueValidator.validateString);
     }
```
stephen-carter-at-sf, line 175: In an attempt to keep this PR short... I'll use the new methods to get a sub extractor and then simplify the logic in here with a separate PR. But things will simplify nicely soon.

packages/code-analyzer-engine-api/src/config.ts
```diff
@@ -23,118 +24,169 @@ export class ConfigValueExtractor {
...
+        return ValueValidator.validateBoolean(this.configObj[fieldName], this.getFieldPath(fieldName));
+    }
+
     extractBoolean(fieldName: string, defaultValue?: boolean): boolean | undefined {
-        return this.isUndefined(fieldName) ? defaultValue :
-            ValueValidator.validateBoolean(this.configObj[fieldName], this.getFullFieldName(fieldName));
+        return this.isUndefined(fieldName) ? defaultValue : this.extractRequiredBoolean(fieldName);
+    }
+
+    extractRequiredNumber(fieldName: string): number {
```
jag-j, line 48: @stephen-carter-at-sf: I know you are just expanding on the existing code. But is it possible to simplify these with extract<T>() method that takes a type parameter?

stephen-carter-at-sf: The validator does this... but I'd prefer to have the extractors be explicit. Specifically the ConfigObject is already limited in the types that it can take... so this actually is easy to have an exhaustive list and it prevents users from providing a type that isn't supported by ConfigObject.

ravipanguluri (approved):

PR #58, opened 2024-07-31
CHANGE (core): @W-16376452@: Simplify config validation

stephen-carter-at-sf:


packages/code-analyzer-core/test/config.test.ts
```diff
@@ -233,8 +230,8 @@ describe("Tests for creating and accessing configuration values", () => {
     });
 
     it("When engines.stubEngine1.disable_engine is not a boolean, then we error", () => {
-        expect(() => CodeAnalyzerConfig.fromObject({engines: {stubEngine1: {disable_engine: null}}})).toThrow(
+        expect(() => CodeAnalyzerConfig.fromObject({engines: {stubEngine1: {disable_engine: 5}}})).toThrow(
```
stephen-carter-at-sf, line 233: We now treat null values the same as undefined everywhere instead of just in some places. So users can always use null in their config as a placeholder that basically says - use the default values. Thus I updated the tests to reflect this.

packages/code-analyzer-engine-api/test/config.test.ts
```diff
@@ -675,4 +685,14 @@ describe("Tests for ConfigValueExtractor", () => {
         expect(subValueExtractors[2].getFieldPath()).toEqual('engines.dummy1.dummy2[5].some_field[2]');
         expect(subValueExtractors[2].getObject()).toEqual({c: 3});
     });
+
```
jfeingold35, line 719: Nitpick: Could you use `it.each()` to separate these into discrete tests instead of jamming them all into one?

packages/code-analyzer-engine-api/src/config.ts
```diff
@@ -120,10 +124,8 @@ export class ConfigValueExtractor {
         return this.fieldRoot && this.fieldRoot.length > 0 ? `${this.fieldRoot}.${fieldName}` : fieldName;
     }
 
-    protected isUndefined(fieldName: string): boolean {
-        // Note a user could provide a boolean false value, which is why we don't just do an if statement on the value
-        // but instead check to see if it is undefined or null.
-        return this.configObj[fieldName] === undefined || this.configObj[fieldName] === null;
+    hasValueDefinedFor(fieldName: string): boolean {
+        return this.configObj[fieldName] !== undefined && this.configObj[fieldName] !== null;
```
jfeingold35, line 128: In theory, you should be able to do `!= null` to catch both `null` and `undefined`, since `(undefined == null) === true`.

packages/code-analyzer-engine-api/test/config.test.ts
stephen-carter-at-sf: I think it.each in this case would be a bit of an overkill.

packages/code-analyzer-engine-api/src/config.ts
stephen-carter-at-sf: I like the readability of what I have here so we aren't left to guess what javascript does here.

ravipanguluri: Also just to clarify for myself: is there a particular reason why you did the negation from isUndefined() to hasValueDefinedFor()

ravipanguluri (approved):

packages/code-analyzer-engine-api/src/config.ts
stephen-carter-at-sf: It was more for readability on my part and so that I could provide hasValueDefinedFor as a first class citizen. null or undefined is sufficient to make hasValueDefinedFor equal to false.

PR #59, opened 2024-08-01
CHANGE(eslint): @W-16358224@: Swap out disable_config_lookup to instead be auto_discover_eslint_config and emit info event

stephen-carter-at-sf:


packages/code-analyzer-eslint-engine/src/config.ts
```diff
@@ -40,7 +40,7 @@ export type ESLintEngineConfig = {
 export const DEFAULT_CONFIG: ESLintEngineConfig = {
     config_root: process.cwd(),
     eslint_config_file: undefined,
-    disable_config_lookup: false,
+    auto_discover_eslint_config: false,
```
jfeingold35, line 43: Observation: This inverts the default behavior, since it's essentially going from `dontDoX = false` to `doX = true`. Was that intentional?

packages/code-analyzer-eslint-engine/src/workspace.ts
```diff
@@ -10,13 +10,34 @@ export interface ESLintWorkspace {
...
+        this.engineConfig = config;
+        this.autoDiscoveredFile = autoDiscoveredFile;
+    }
+
+    getUserConfigFile(): string | undefined {
+        return this.engineConfig.eslint_config_file ||
+            (this.engineConfig.auto_discover_eslint_config ? this.autoDiscoveredFile : undefined);
+    }
+
+    userConfigIsRelevant(): boolean {
```
jfeingold35: I don't love the use of `relevant` in the method name. Is there an alternative word we could use that's more descriptive?

packages/code-analyzer-eslint-engine/test/engine.test.ts
```diff
@@ -408,9 +437,28 @@ describe('Typical tests for the runRules method of ESLintEngine', () => {
...
+
+        const relPathFromCwd: string = path.join(workspaceThatHasCustomConfigWithNewRules.slice((process.cwd() + path.sep).length), '.eslintrc.yml');
+        expect(logEvents).toContainEqual({
+            type: EventType.LogEvent,
+            logLevel: LogLevel.Info,
+            message: getMessage('UnusedEslintConfigFile', relPathFromCwd, relPathFromCwd)
+        });
+    });
+
+    it('When runRules is called on workspace that ignores files by config, then those files are ignored', async () => {
```
jfeingold35: "Ignores files by config" is a confusing phrase. is there something else we could say instead? E.g., "workspace with a config that ignores files"?

packages/code-analyzer-eslint-engine/src/config.ts
stephen-carter-at-sf: Yes... that's the whole point. We discussed this the other day. Maybe you were sick on that day. I pinged you on slack to show you the discussion.

packages/code-analyzer-eslint-engine/src/workspace.ts
stephen-carter-at-sf: I could use some help with suggestions... basically it means that the user has either supplied a config file directly or turned on auto config lookup. Ideas?

packages/code-analyzer-eslint-engine/test/engine.test.ts
stephen-carter-at-sf: Sure. Fixed.

packages/code-analyzer-eslint-engine/src/workspace.ts
stephen-carter-at-sf: Maybe just enabled?

stephen-carter-at-sf: Changed this to userConfigIsEnabled.

jfeingold35 (approved):

PR #61, opened 2024-08-05
CHANGE: @W-16263070@: Move Workspace implementation to directly be in the engine api

stephen-carter-at-sf:
This work removes the unneeded testtools.createWorkspace stuff I put in temporarily. Having Workspace be a first class citizen on the engineApi means we no longer need redundant code and engines will test with the real thing.

This prepares for adding in more functionality to the workspace soon.


packages/code-analyzer-core/src/code-analyzer.ts
```diff
@@ -302,24 +339,27 @@ function validateRuleDescriptions(ruleDescriptions: engApi.RuleDescription[], en
...
+        const pathStartPointPromises: Promise<engApi.PathPoint[]>[] = runOptions.pathStartPoints.map(extractEnginePathStartPoints);
+        engineRunOptions.pathStartPoints = (await Promise.all(pathStartPointPromises)).flat();
     }
     validatePathStartPointsAreInsideWorkspace(engineRunOptions);
     return engineRunOptions;
 }
 
-// TODO: Eventually we should make all these validations async
-function validateFileOrFolder(fileOrFolder: string): string {
+async function validateFileOrFolder(fileOrFolder: string): Promise<string> {
```
stephen-carter-at-sf, line 357: I went ahead and handled one of the TODOs in this file - to make file validation work asyncronously.

packages/code-analyzer-core/test/workspace.test.ts
```diff
@@ -42,15 +47,4 @@ describe("Tests for the createWorkspace method of CodeAnalyzer", () => {
             'test/test-data/sampleWorkspace/.gitignore',]);
         expect(workspace.getFilesAndFolders()).toEqual([path.resolve('test', 'test-data', 'sampleWorkspace', 'someFile.txt')]);
     });
-
-    it("When calling getExpandedFiles, then all files underneath all subfolders are found", async () => {
```
stephen-carter-at-sf, line 46: This test just got moved to the engine api's workspace tests.

packages/code-analyzer-engine-api/src/workspace.ts
```diff
@@ -1,21 +1,24 @@
...
-export interface Workspace {
-    getWorkspaceId(): string
-    getFilesAndFolders(): string[]
-    getExpandedFiles(): Promise<string[]>
-}
+const UNWANTED_FOLDERS: string[] = ['node_modules', '.git', '.github'];
+const UNWANTED_FILES: string[] = ['code_analyzer_config.yml', 'code_analyzer_config.yaml', '.gitignore'];
 
-export class WorkspaceImpl implements Workspace, engApi.Workspace {
+export class Workspace {
```
stephen-carter-at-sf, line 7: This file was basically just moved from the core module to here and some minor touchups were done.

packages/code-analyzer-eslint-engine/test/engine.test.ts
```diff
@@ -81,7 +81,7 @@ describe('Tests for the describeRules method of ESLintEngine', () => {
             auto_discover_eslint_config: true
         });
         const ruleDescriptions: RuleDescription[] = await engine.describeRules({
-            workspace: testTools.createWorkspace([caseObj.folder])});
+            workspace: new Workspace([caseObj.folder])});
```
stephen-carter-at-sf, line 84: Lastly, a ton of tests were updated to have this cleaner workspace construction.

packages/code-analyzer-engine-api/src/workspace.ts
```diff
@@ -1,21 +1,24 @@
...
-        this.workspaceId = workspaceId;
-        this.filesAndFolders = removeRedundantPaths(absFilesAndFolders).filter(isWantedPath);
+    /**
+     * Constructs a workspace with a list of absolute file and folder paths
+     * * This constructor assumes that the list of files and folder paths exists and that they are absolute paths
+     * @param absoluteFileAndFolderPaths Absolute file and folder paths
+     * @param workspaceId Optional workspace identifier
+     */
+    constructor(absoluteFileAndFolderPaths: string[], workspaceId?: string) {
+        this.workspaceId = workspaceId || `workspace${++Workspace.nextId}`;
```
jfeingold35, line 20: Question: Should manually specifying a workspace ID trigger an attempt to update `nextId`, or have some other way of preventing duplicate IDs? Definitely not a blocker, but just something I noticed.

jfeingold35 (approved):

jag-j (approved):

packages/code-analyzer-engine-api/src/workspace.ts
stephen-carter-at-sf: The true workspace Ids actually come from the UniqueIdGenerator from the CodeAnalyzer's createWorkspace class. So the real IDs are just passed in when clients use CodeAnalyzer the way they should. But to help simulate uniqueness in tests when creating tests for the engines, this incrementing value was added. It has no impact on production. But to answer your question, even if we incremented every time, then that nextId is always new and so at worse - we would just be skipping numbers which isn't a big deal since the numbers are still unique.

PR #63, opened 2024-08-06
NEW: @W-16263070@: Exclude from the workspace any dot files/folders unless explicitly added

stephen-carter-at-sf:
…and other excludes that are underneath the workspace root

### DIFF
--- a/packages/code-analyzer-engine-api/src/workspace.ts
+++ b/packages/code-analyzer-engine-api/src/workspace.ts
@@ -108,6 +108,7 @@ export class Workspace {
         if(this.getWorkspaceRoot()) {
             return fileOrFolder.slice(this.getWorkspaceRoot()!.length);
         }
+        /* istanbul ignore next */
         return fileOrFolder;
     }
 }
--- a/packages/code-analyzer-engine-api/src/workspace.ts
+++ b/packages/code-analyzer-engine-api/src/workspace.ts
@@ -1,13 +1,15 @@
 import fs from "node:fs";
 import path from "node:path";
 
-const UNWANTED_FOLDERS: string[] = ['node_modules', '.git', '.github'];
-const UNWANTED_FILES: string[] = ['code_analyzer_config.yml', 'code_analyzer_config.yaml', '.gitignore'];
+const NON_DOT_FOLDERS_TO_EXCLUDE: string[] = ['node_modules'];
+const NON_DOT_FILES_TO_EXCLUDE: string[] = ['code_analyzer_config.yml', 'code_analyzer_config.yaml'];
 
 export class Workspace {
     private static nextId: number = 0;
     private readonly workspaceId: string;
-    private readonly filesAndFolders: string[];
+    private readonly rawFilesAndFolders: string[];
+
+    private normalizedFilesAndFolders?: string[];
     private expandedFiles?: string[];
     private workspaceRoot?: string | null;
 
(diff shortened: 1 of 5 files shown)


packages/code-analyzer-engine-api/src/workspace.ts
```diff
@@ -59,15 +62,54 @@ export class Workspace {
...
+     *   then we can add an enhancement to receive the workspace root explicitly from the user.
+     */
+    private shouldExclude(fileOrFolder: string): boolean {
+        const relativeFileOrFolder: string = this.makeRelativeToWorkspaceRoot(fileOrFolder);
+        if (relativeFileOrFolder.length === 0) { // folder is equal to the workspace root
+            return false;
+        }
+        return NON_DOT_FOLDERS_TO_EXCLUDE.some(f => relativeFileOrFolder.includes(`${path.sep}${f}${path.sep}`)) ||
+            NON_DOT_FILES_TO_EXCLUDE.includes(path.basename(relativeFileOrFolder)) ||
+            relativeFileOrFolder.includes(`${path.sep}.`);
```
jfeingold35, line 97: Is it still possible to deliberately target a dot file if you want to?

stephen-carter-at-sf: Yeah this is where things get funny. The problem is that I don't actually get the input as the user provided it to the client. I only get the input as it was provided by the client. So consider the glob case. Imagine I'm in a folder '/abc/def' and I have the glob '**/*.ts' which finds:
* Scenario 1: It finds '/abc/def/good/file.ts'  - then this is what I receive and all is well
* Scenario 2: It finds '/abc/def/.bad/file.ts' - then this is all I receive.. and so the workspace root is assumed to be '/abc/def/.bad/' which makes 'file.ts' valid and so yes this would come through.
* Scenario 3: It finds '/abc/def/anotherFile.ts' and '/abc/def/.bad/file.ts'... then the workspace root is assumed to be '/abc/def' and thus the '.bad/file.ts' would be dropped.

Most likely we want to have these dot folders drop. But as a core framework, I don't know if the glob produced the file or if the user directly input it. So my solution basically works in 99.9% of cases... where we exclude it unless it is like the only thing explicitly given... while still allowing dotFolders to be above the workspace root.

As I stated in my comment. If this becomes an issue (which I doubt it will) - especially since the majority of users will be simply supply a directory as their workspace... then we can add in a workspaceRoot flag or something to the cli.

jfeingold35: Okay. If the user can't possibly target `.` files, then that does mean some of my tests are excessive, but I'll probably leave them in anyway on general principle.
Question: Do we imagine that future rules (perhaps rules for the regex engine?) might want to target `.` files? If so, then we may need to revisit this in the future.

stephen-carter-at-sf: Let's have a quick chat about this in person... there are a number of areas we could go with this - based on the communication we currently have between the client and core side of things.

stephen-carter-at-sf: Reworked algorithm to adhere to new behavior:

Code analyzer will attempt to exclude hidden (.dot) files/folders, node_modules, and code-analyzer.yml/code-analyzer.yaml files that live within the folders of our workspace, unless they explicitly included in your workspace. So if you target the current directory . and it contains one of these underneath it, then they will be excluded. But if you add to --workspace  a glob that explicitly includes or a path that is one of these normally excluded files/folders then they will be included in the anaylsis.

jfeingold35 (approved):

PR #64, opened 2024-08-07
NEW (FlowTest) @W-16360668@ Added scaffolding for FlowTest engine.

jfeingold35:


stephen-carter-at-sf (approved):

jag-j (approved):

stephen-carter-at-sf:
The only thing I think we need to be cautious about here is… the name flowtest. I don’t know if that is the name we want to use ultimately.
So things like the package.json file - if we submit this now… we’ll need to remember to go back and update the name later.

PR #65, opened 2024-08-07
CHANGE @W-16438030@ Make file extensions an optional field for regex engine rules

ravipanguluri:
In this PR:

A/C

 * Users can create custom regex rules without file extensions given
 * Any regex rule with no extensions given will run on all files in a workspace if they have text.


packages/code-analyzer-regex-engine/src/engine.ts
```diff
@@ -61,6 +62,9 @@ export class RegexEngine extends Engine {
     }
 
     private shouldScanFile(fileName: string, ruleName: string): boolean {
+        if (this.regexRules[ruleName].file_extensions.length === 0){
+            return isTextFile(fileName);
+        }
         const ext: string = path.extname(fileName).toLowerCase();
```
jag-j, line 72: @ravipanguluri - So, we don't check for binary files if the extension is specified? .exe or .bat files for example?

ravipanguluri: This follows from how Stephen checks it in the retire-js engine. Based on [documentation ](https://www.npmjs.com/package/isbinaryfile) for the underlying package, they don't seem to check by file extension, but by the contents of the file itself.  Also shouldScanFile() gets the entire file path, not just the name. So i'm assuming in most .exe or .bat files, there's enough non-ASCII characters to raise alarm bells.

packages/code-analyzer-regex-engine/src/config.ts
```diff
@@ -55,8 +55,8 @@ export function validateAndNormalizeConfig(rawConfig: ConfigObject): RegexEngine
         const description: string = ruleExtractor.extractRequiredString('description')
         const rawRegexString: string = ruleExtractor.extractRequiredString('regex')
         const regex: RegExp = validateRegex(rawRegexString, ruleExtractor.getFieldPath('regex'))
-        const rawFileExtensions: string[] = ruleExtractor.extractRequiredArray('file_extensions',
-            (element, fieldPath) => ValueValidator.validateString(element, fieldPath, FILE_EXT_PATTERN));
+        const rawFileExtensions: string[] = ruleExtractor.extractArray('file_extensions',
+            (element, fieldPath) => ValueValidator.validateString(element, fieldPath, FILE_EXT_PATTERN), [])!;
```
stephen-carter-at-sf: OK we need to be careful here. If the user has [] in their config file... do we treat this as none or all? 
I would instead make rawFileExtensions be of type `string[] | undefined` here and do not supply a default value. This way, we know whether or not they supplied the value or not by checking if rawFileExtensions is undefined. If not then we can treat it as all. Otherwise, I believe we should accept `[]` as none from the user.

stephen-carter-at-sf: So please update line 30 and 31 so that
```
        /// The extensions of the files that you would like to test the regular expression against.
        file_extensions: string[];
```

it is instead something like:
```
        // The extensions of the files that you would like to test the regular expression against.
        // If not defined, or equal to null, the all text-based files of any file extension will be tested.
        file_extensions?: string[];
```

packages/code-analyzer-regex-engine/src/engine.ts
```diff
@@ -61,6 +62,9 @@ export class RegexEngine extends Engine {
     }
 
     private shouldScanFile(fileName: string, ruleName: string): boolean {
+        if (this.regexRules[ruleName].file_extensions.length === 0){
```
stephen-carter-at-sf: Along with my other comments, this should instead be:
```
if (this.regexRules[ruleName].file_extensions.length){
```

packages/code-analyzer-regex-engine/src/engine.ts
```diff
@@ -121,10 +117,14 @@ function getLineNumber(charIndex: number, newlineIndexes: number[]): number{
 function getNewlineIndices(fileContents: string): number[] {
     const newlineRegex: RegExp = new RegExp(os.EOL, "g");
     const matches = fileContents.matchAll(newlineRegex);
-    const newlineIndexes = [0];
+    const newlineIndexes = [-1];
```
stephen-carter-at-sf, line 126: Why did this change? Did you find a bug or something? Help me understand this a bit.

packages/code-analyzer-regex-engine/src/engine.ts
```diff
@@ -11,6 +11,7 @@ import path from "node:path";
 import fs from "node:fs";
 import os from "node:os";
 import {RegexRules} from "./config";
+import {isBinaryFileSync} from "isbinaryfile";
```
stephen-carter-at-sf, line 14: You are using isbinaryfile, but you didn't update package.json to add this dependency for this engine.
If the retire-js engine isn't installed... then this dependency would be missing.

Specifically us the same version as the retire-js engine as well:
    "isbinaryfile": "^5.0.2",

packages/code-analyzer-regex-engine/src/engine.ts
```diff
@@ -61,6 +62,9 @@ export class RegexEngine extends Engine {
     }
 
     private shouldScanFile(fileName: string, ruleName: string): boolean {
+        if (this.regexRules[ruleName].file_extensions.length === 0){
+            return isTextFile(fileName);
```
stephen-carter-at-sf: Please add in a //TODO here so that we eventually speed this up. If every single rule that doesn't have a file extension checks if the file is a text file (which is a slow operation) then we should be caching this information against the file. Let's fix the bug first, then refactor afterward.

packages/code-analyzer-regex-engine/src/engine.ts
```diff
@@ -121,10 +117,14 @@ function getLineNumber(charIndex: number, newlineIndexes: number[]): number{
...
+    const newlineIndexes = [-1];
 
     for (const match of matches) {
         newlineIndexes.push(match.index)
     }
     return newlineIndexes;
+}
+
+function isTextFile(fileName: string | Buffer): boolean {
+    return !isBinaryFileSync(fileName);
```
stephen-carter-at-sf: Add in a //TODO so that we don't do this synchronously. We should optimize the performance here later on so that we aren't doing one file at a time synchronously. And do me a favor and add in //TODO inside of code-analyzer-retire-js-engine/src/utils.ts on line 63 with the same todo.

stephen-carter-at-sf (changes requested):

packages/code-analyzer-regex-engine/src/engine.ts
ravipanguluri: Yes, I found a bug. Basically if a violation occurs at the very beginning of file without any newlines it reported the beginning position as line: 0, column: 0. It came up in one of the tests that I made. Should I leave it as-is for here and make the change in it's own PR.

stephen-carter-at-sf: I'm all for fix as you go. Just was asking. Thanks.

packages/code-analyzer-regex-engine/src/engine.ts
stephen-carter-at-sf: In fact... while the TODO remains... let's try to short circuit this answer as much as possible. For now, create a constant in the file as a Set<string> that contains a set of all the most common text based file extensions. Then we can first just check the file extension - and if not... then we back up to use the isBinaryFileSync function for now.

packages/code-analyzer-regex-engine/src/config.ts
ravipanguluri: I see your point. I guess it just seemed odd to me that a user would want a rule that is run on no files. But thinking more, I can see some use cases.

packages/code-analyzer-regex-engine/src/engine.ts
ravipanguluri: How exhaustive should this set be? Should it just be the most common code files that we expect code analyzer to be run against?

stephen-carter-at-sf: Well the target audience is salesforce developers who will be dealing with salesforce files in a salesforce project. I'll send you a list of sample project repos in slack... just walk through some of those repos to get a good idea of what extensions might live in them.

packages/code-analyzer-regex-engine/src/config.ts
```diff
@@ -27,8 +27,9 @@ export type RegexRules = {
         // The regular expression that triggers a violation when matched against the contents of a file.
         regex: RegExp;
 
-        /// The extensions of the files that you would like to test the regular expression against.
-        file_extensions: string[];
+        // The extensions of the files that you would like to test the regular expression against.
+        // If not defined, or equal to null, the all text-based files of any file extension will be tested.
```
stephen-carter-at-sf: "the all" should be "then all"

packages/code-analyzer-regex-engine/src/engine.ts
```diff
@@ -11,6 +11,25 @@ import path from "node:path";
 import fs from "node:fs";
 import os from "node:os";
 import {RegexRules} from "./config";
+import {isBinaryFileSync} from "isbinaryfile";
+
+const TEXT_BASED_FILE_EXTS = new Set<string>(
```
stephen-carter-at-sf, line 16: nitpick... no need to have each extension on its own line. It can probably fit on one or two lines.

Also, please alphabetize the list and add in a few more that I just thought of:
'.json', '.md', '.htm', '.page', '.csv', '.yml', '.yaml', '.mdt', '.xsl', '.xslt', '.rtf'

packages/code-analyzer-regex-engine/src/engine.ts
```diff
@@ -60,9 +79,14 @@ export class RegexEngine extends Engine {
...
 
+    /*TODO: cache whether or not a file is text-based so calls to isTextFile() are minimized */
     private shouldScanFile(fileName: string, ruleName: string): boolean {
-        const ext: string = path.extname(fileName).toLowerCase();
-        return this.regexRules[ruleName].file_extensions.includes(ext);
+        if (this.regexRules[ruleName].file_extensions){
+            const ext: string = path.extname(fileName).toLowerCase();
+            return this.regexRules[ruleName].file_extensions.includes(ext);
+        }
+        return isTextFile(fileName);
```
stephen-carter-at-sf: Since the ext is needed for the quick return, let's keep the isTextFile as it was before. and instead put the quick return logic here (so that later we only need to cache the things that we can't do quickly - keeping the cache small).

So basically I propose doing something like:
```
    const ext: string = path.extname(fileName).toLowerCase();
    return this.regexRules[ruleName].file_extensions ? 
        this.regexRules[ruleName].file_extensions.includes(ext) :
        TEXT_BASED_FILE_EXTS.includes(ext) || isTextFile(fileName);
```

packages/code-analyzer-regex-engine/src/engine.ts
```diff
@@ -60,9 +79,14 @@ export class RegexEngine extends Engine {
         return (await Promise.all(violationPromises)).flat();
     }
 
+    /*TODO: cache whether or not a file is text-based so calls to isTextFile() are minimized */
```
stephen-carter-at-sf: Now that I think about it... just move this comment down to isTextFile function

```
// TODO: Make this async and cache results to improve performance
function isTextFile(fileName: string): boolean {
    return !isBinaryFileSync(fileName);
}
```

stephen-carter-at-sf (changes requested):

stephen-carter-at-sf (approved):

PR #66, opened 2024-08-07
FIX(RegexEngine) @W-16441880@ Custom rules without 'g' modifier are not properly handled by engine

ravipanguluri:
In this PR:

*If user does not enable the global modifier flag on their custom regular expression rule, they are dispatched to a custom error message which tells them that the regex engine does not currently support regular expressions w/o the modifier enabled.

NOTE: The reason I chose to not allow users to enable to global modifier for now is that when .match is called with the 'g' flag, the regex matches do not come with important information like the match.index that I need to report violations. So therefore, when 'g' is enabled, I have no choice but to use matchAll(). In theory, I could dispatch between using match/matchAll depending on the 'g' flag but, it just makes the engine more clunky and gives us more code to maintain for a use case that I just can't imagine anyone wanting. I think it's just better to see if users come to us with somewhere it would actually be needed.

### DIFF
--- a/packages/code-analyzer-regex-engine/src/messages.ts
+++ b/packages/code-analyzer-regex-engine/src/messages.ts
@@ -17,7 +17,10 @@ const MESSAGE_CATALOG : { [key: string]: string } = {
         `The '%s' configuration value is invalid. The value could not be converted into a regular expression: %s`,
 
     InvalidRuleName:
-        `The rule name '%s' defined within the '%s' configuration value is invalid. The rule name must match the regular expression: '%s'`
+        `The rule name '%s' defined within the '%s' configuration value is invalid. The rule name must match the regular expression: '%s'`,
+
+    GlobalModifierNotProvided:
+        `The '%s' configuration value is invalid. The regex engine does not currently support rules without the 'g' modifier. Please use '%s' instead of '%s'.`
 }
 
 /**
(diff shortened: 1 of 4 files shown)


packages/code-analyzer-regex-engine/src/messages.ts
```diff
@@ -17,7 +17,10 @@ const MESSAGE_CATALOG : { [key: string]: string } = {
         `The '%s' configuration value is invalid. The value could not be converted into a regular expression: %s`,
 
     InvalidRuleName:
-        `The rule name '%s' defined within the '%s' configuration value is invalid. The rule name must match the regular expression: '%s'`
+        `The rule name '%s' defined within the '%s' configuration value is invalid. The rule name must match the regular expression: '%s'`,
+
+    GlobalModifierNotProvided:
+        `The '%s' configuration value is invalid. The regex engine does not currently support rules without the 'g' modifier enabled. `
```
stephen-carter-at-sf: Let's hold their hand a bit here. Let's make this message be:

```
The '%s' configuration value is invalid. The regex engine does not currently support rules without the 'g' modifier. Please use '%s' instead of '%s'.
```
and then have line 96 of config.ts be:
```
throw new Error(getMessage('GlobalModifierNotProvided', fieldName, `/${pattern}/g${modifiers}`, value));
```

stephen-carter-at-sf (changes requested):

stephen-carter-at-sf (approved):

PR #69, opened 2024-08-09
NEW (RegexEngine) @W-16384199@ Create new default rules for regex engine

ravipanguluri:
In this PR:

* 4 new rules are introduced: UseInclusiveHierarchicalTerms, UseInclusiveSecurityTerms, UpdateOldApexApiVersion, UseInclusiveOutageTerms
* Rule attributes are exactly as defined in regex engine configuration document
* Rules are run and described with runRules()/describeRules() in any directory containing text-based files (since file extensions are not provided for these rules).


packages/code-analyzer-regex-engine/src/plugin.ts
```diff
@@ -14,6 +14,47 @@ export const BASE_REGEX_RULES: RegexRules = {
         severityLevel: SeverityLevel.Info,
         tags: ['Recommended', 'CodeStyle'],
         resourceUrls: []
+    },
+    UseInclusiveSecurityTerms: {
```
stephen-carter-at-sf: John should be the one to decide whether these are rules we want to add, and what tags / severity levels they should be. You should set up a meeting with him.

ravipanguluri: I have already set one up for tomorrow.

ravipanguluri: As an update to this, I received approval on the rules from John with some minor updates which are reflected in the most recent changes.

packages/code-analyzer-regex-engine/src/messages.ts
```diff
@@ -10,6 +10,24 @@ const MESSAGE_CATALOG : { [key: string]: string } = {
     TrailingWhitespaceRuleMessage:
         `Found trailing whitespace at the end of a line of code.`,
 
+    AvoidColorizedTermsRuleDescription:
+        `"Detects usage of '%s’ or ‘%s'. This phrasing uses color as a metaphor for good and bad which can reinforce implicit bias.""`,
+
+    AvoidColorizedTermsRuleMessage:
+        `One of the terms '%s or '%s' was found. Replace it with a more inclusive term like '%s' or '%s'"`,
```
stephen-carter-at-sf: 1) It looks like your single quotes are a bit off in here. 
`'%s’` should be `'%s'`
`‘%s'` should be `'%s'`
`'%s` should be `'%s'`

2) What is with the two double quotes at the end of line 14? `""` And do we even want a quote at the beginning of line 14?
3) What is with the double quote at the end of line 17? `"`... should that instead be `.`?
4) What do you mean "for good and bad" ... Can we reword this or is this something you pulled from a doc page?
5) For AvoidColorizedTermsRuleDescription, instead of passing in exactly what we are looking for, can we make this more generic so that we can expand the list in the future if we wanted to. Something like a rule called AvoidTermsWithImplicitBias that has a simple description of:
`Detects usage of terms that reinforce implicit bias.`

6) The AvoidColorzedTermsRuleMessage... this looks a lot like the UseInclusiveHierarchicalTermsRuleMessage... can we instead reuse this message and not have it be `'%s' or '%s'` but just have what was found and what it should be replaced with? If we have multiple options, then maybe have the message be generic as something like:
`The term '%s' was found. Replace it with a more inclusive term, like one of the following: %s `
where %s here would be the output of JSON.stringify(stringArrayOfAcceptedTerms)

packages/code-analyzer-regex-engine/src/messages.ts
```diff
@@ -10,6 +10,24 @@ const MESSAGE_CATALOG : { [key: string]: string } = {
...
+
+    UseInclusiveHierarchicalTermsRuleDescription:
+        `Detects usage of the word 'slave' which perpetuates systemic biases.'`,
+
+    UseInclusiveHierarchicalTermsRuleMessage:
+        `The term 'slave' was found. Replace it with a more inclusive term like 'secondary' or 'follower'`,
+
+    UpdateOldApexApiVersionRuleDescription:
+        `Detects Apex classes relying on API versions that are below 60.0`,
```
stephen-carter-at-sf: Can you look through all the the rule descriptions and determine whether or not the descriptions in general have a period `.` at the end or not and try to be consistent with all of our descriptions?

packages/code-analyzer-regex-engine/src/messages.ts
```diff
@@ -10,6 +10,24 @@ const MESSAGE_CATALOG : { [key: string]: string } = {
...
+
+    UpdateOldApexApiVersionRuleMessage:
+        `Found an Apex class relying on an API version 3 years or older. Update your API version.`,
```
stephen-carter-at-sf: This message seems different than the description. Also consider what happens 2 years from now... will this still say 3 years or older?

In fact, there is an open question on... should this rule keep getting updated dynamically instead of being fixed? If so, then shouldn't we calculate the rule version based on the current date or something? Or do we want the rule to be fixed in which case this message should just say "version 60.0".

What is the rule requirement here? If it indeed is to detect versions based on time and how old they are... then let's write a utility function that calculate based on dateTime what version was associated with that date. Then we can use that utility function to automatically update the rule everytime the version is expected to change. For example, I believe we have 3 releases a year (i.e. every 4 months). So if we know the estimated dates of when this will be... then we should be able to make this dynamic. I'd like to avoid having to keep updating this rule every 4 months manually.

stephen-carter-at-sf (changes requested):

packages/code-analyzer-regex-engine/src/messages.ts
ravipanguluri: From the spike document, the intent of the rule is to detect when an API version is 3 years or older just to make sure that users are not relying on an API version that is very outdated. In that case, I can write a utility to do that.

packages/code-analyzer-regex-engine/src/messages.ts
ravipanguluri: To clarify, for 5 the specific language is taken from the resource url for the [rule](https://www.salesforce.com/news/stories/salesforce-updates-technical-language-in-ongoing-effort-to-address-implicit-bias/). I agree the phrasing is a bit clunky though, so I'll make a second pass while keeping the intent the same.

packages/code-analyzer-regex-engine/test/utils.test.ts
```diff
@@ -0,0 +1,29 @@
...
+    it('Throw an error, if date is outside valid range to generate a valid API version ', () => {
+        const earlyDate = new Date('2006-01-01');
+        const lateDate = new Date('2050-01-01');
+        expect(() => getDeprecatedApiVersionRegex(earlyDate)).toThrowError(
+            "maxNumber must be between 0 and 99"
+        );
+        expect(() => getDeprecatedApiVersionRegex(lateDate)).toThrowError(
+            "maxNumber must be between 0 and 99"
+        );
+    });
```
stephen-carter-at-sf: Can you make a test that will fail in the future but currently passes right now? That is use new Date() as input for example. It shouldn't throw  today but put a comment in that says if it throws the maxNumber must be between 0 and 99 in the future then the rule will need to be rewritten. Basically, we want our build to fail in the future with this error before it reaches customers.

packages/code-analyzer-regex-engine/src/utils.ts
```diff
@@ -0,0 +1,39 @@
...
+    const year = date.getFullYear();
+    const month = date.getMonth();
+
+    if (month >= 0 && month < 4) {
+        return (year - 2004) * 3;
+    } else if (month >= 4 && month < 8) {
+        return (year - 2004) * 3 + 1;
+    } else {
+        return (year - 2004) * 3 + 2;
+    }
```
stephen-carter-at-sf: * I see Winter '25 [v62.0] finishes going out in October (10/12/2024 specifically).
* I see the spring release typically goes out in February (like feb 12th I believe is when Spring 24 went GA on the R2b instances last year) - feb 15th 2024 will be Spring '25 [v63.0]. 
* I see  Summer '25 [v64.0] finishes going out in June (6/14/2025 specifically).

So even though we'll probably not get the dates perfect, shouldn't this approximation instead be:
```
    if (month >= 1 && month < 5) { // Feb through May
        return (year - 2004) * 3;
    } else if (month >= 5 && month < 9) { // Jun through Sep
        return (year - 2004) * 3 + 1;
    } else { // Oct through Jan
        return (year - 2004) * 3 + 2;
    }
```

stephen-carter-at-sf (changes requested):

packages/code-analyzer-regex-engine/src/utils.ts
ravipanguluri: I had it as the first time that customers get the release since Salesforce tends to stagger their releases across multiple dates. But you're right that it makes much more sense to do it for the final release date.

packages/code-analyzer-regex-engine/src/plugin.ts
```diff
@@ -11,9 +12,45 @@ export const BASE_REGEX_RULES: RegexRules = {
         violation_message: getMessage('TrailingWhitespaceRuleMessage'),
         severity: SeverityLevel.Info,
         tags: ['Recommended', 'CodeStyle']
+    },
+    UseInclusiveSecurityTerms: {
```
stephen-carter-at-sf: Please combine the 3 rules into one. I think your message catalog is revealing the true rule name: AvoidTermsWithImplicitBias.

packages/code-analyzer-regex-engine/src/plugin.ts
```diff
@@ -11,9 +12,45 @@ export const BASE_REGEX_RULES: RegexRules = {
...
+        severity: SeverityLevel.Info
+    },
+    UseInclusiveHierarchicalTerms: {
+        regex: /\bslaves?\b/gi,
+        description: getMessage('AvoidTermsWithImplicitBiasRuleDescription', JSON.stringify(['slave'])),
+        violation_message: getMessage('AvoidTermsWithImplicitBiasRuleMessage', JSON.stringify(['slave']), JSON.stringify(['secondary', 'follower'])),
+        tags: ["Recommended"],
+        severity: SeverityLevel.Info
+    },
+    UpdateOldApexApiVersion: {
```
stephen-carter-at-sf: Since all of our rules use negative wording... like "No*" and "Avoid*" maybe this should be something like "AvoidOldApiVersions"

packages/code-analyzer-regex-engine/src/messages.ts
```diff
@@ -10,6 +10,18 @@ const MESSAGE_CATALOG : { [key: string]: string } = {
...
         `Found trailing whitespace at the end of a line of code.`,
 
+    AvoidTermsWithImplicitBiasRuleDescription:
+        `"Detects usage of terms: %s which reinforce implicit bias.`,
+
+    AvoidTermsWithImplicitBiasRuleMessage:
+        `A term from the following: '%s' was found. Replace it with a more inclusive term like one of the following: %s.`,
+
+    UpdateOldApexApiVersionRuleDescription:
+        `Detects Apex classes relying on API versions that are below 52.0.`,
```
stephen-carter-at-sf: We either need to pass in the version here or or make it generic.

packages/code-analyzer-regex-engine/src/messages.ts
```diff
@@ -10,6 +10,18 @@ const MESSAGE_CATALOG : { [key: string]: string } = {
...
+
+    UpdateOldApexApiVersionRuleMessage:
+        `Found an Apex class relying on an API version 3 years or older. Update your API version.`,
```
stephen-carter-at-sf: Do we want the violation message to say what the minimum api version is and pass that in?

packages/code-analyzer-regex-engine/src/messages.ts
```diff
@@ -10,6 +10,18 @@ const MESSAGE_CATALOG : { [key: string]: string } = {
     TrailingWhitespaceRuleMessage:
         `Found trailing whitespace at the end of a line of code.`,
 
+    AvoidTermsWithImplicitBiasRuleDescription:
+        `"Detects usage of terms: %s which reinforce implicit bias.`,
```
stephen-carter-at-sf: I'd remove the semi colon here.
Also, I wonder if we need to actually display all the different terms we are looking for. If this grows to be really large, then do we want the rule description saying all of them.

I think having the list in the violation makes sense... but i'm on the fence about it being in the description. You can ask others what they think about this. Maybe it doesn't matter right now as long as the list is under 10 terms.

stephen-carter-at-sf (approved):

PR #70, opened 2024-08-09
CHANGE @W-16461375@ Make default severity/tags override-able in rules declaration for regex engine

ravipanguluri:
In this PR:

A/C:

* Users can specify tags/severity levels for their custom rules at engines.regex.custom_rules configuration level.
* If a user doesn't give values for these fields, the same defaults as previously defined in configuration documentation will be maintained.

### DIFF
--- a/packages/code-analyzer-regex-engine/src/config.ts
+++ b/packages/code-analyzer-regex-engine/src/config.ts
@@ -3,8 +3,6 @@ import {
     ConfigValue,
     ConfigValueExtractor,
     getMessageFromCatalog,
-    RuleDescription,
-    RuleType,
     SeverityLevel,
     SHARED_MESSAGE_CATALOG,
     ValueValidator
@@ -27,30 +25,32 @@ export type RegexRules = {
     //       violation: "A comment with a TODO statement was found. Please remove TODO statements from your Apex code."
     //       severity: "Info"
     //       tags: ['TechDebt']
-    [ruleName: string] : RuleDescription & {
-        // The regular expression that triggers a violation when matched against the contents of a file.
-        regex: RegExp;
+    [ruleName: string] : RegexRule
+}
 
-        // The extensions of the files that you would like to test the regular expression against.
-        // If not defined, or equal to null, then all text-based files of any file extension will be tested.
-        file_extensions?: string[];
+export type RegexRule = {
+    // The regular expression that triggers a violation when matched against the contents of a file.
+    regex: RegExp;
 
-        // A description of the rule's purpose.
-        description: string;
+    // The extensions of the files that you would like to test the regular expression against.
+    // If not defined, or equal to null, then all text-based files of any file extension will be tested.
+    file_extensions?: string[];
 
-        // [Optional] The message emitted when a rule violation occurs. This message is intended to help the user understand the violation.
-        // Default: `A match of the regular expression <regex> was found for rule '<ruleName>': <description>`
-        violation_message: string;
+    // A description of the rule's purpose.
+    description: string;
 
-        // [Optional] The string array of tag values to apply to this rule by default.
-        // Default: ['Recommended']
-        tags: string[];
+    // [Optional] The message emitted when a rule violation occurs. This message is intended to help the user understand the violation.
+    // Default: `A match of the regular expression <regex> was found for rule '<ruleName>': <description>`
+    violation_message: string;
 
-        // [Optional] The severity level to apply to this rule by default.
-        // Possible values: 1 or 'Critical', 2 or 'High', 3 or 'Moderate', 4 or 'Low', 5 or 'Info'
-        // Default: 'Moderate'
-        severityLevel: SeverityLevel;
-    }
+    // [Optional] The string array of tag values to apply to this rule by default.
+    // Default: ['Recommended']
+    tags: string[];
+
+    // [Optional] The severity level to apply to this rule by default.
+    // Possible values: 1 or 'Critical', 2 or 'High', 3 or 'Moderate', 4 or 'Low', 5 or 'Info'
+    // Default: 'Moderate'
+    severity: SeverityLevel;
 }
 
 export const FILE_EXT_PATTERN: RegExp = /^[.][a-zA-Z0-9]+$/;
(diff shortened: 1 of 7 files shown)


packages/code-analyzer-regex-engine/src/config.ts
```diff
@@ -58,20 +70,22 @@ export function validateAndNormalizeConfig(rawConfig: ConfigObject): RegexEngine
         const regex: RegExp = validateRegex(rawRegexString, ruleExtractor.getFieldPath('regex'))
         const rawFileExtensions: string[] | undefined = ruleExtractor.extractArray('file_extensions',
             (element, fieldPath) => ValueValidator.validateString(element, fieldPath, FILE_EXT_PATTERN));
+        const rawSeverityLevelValue: number | undefined = ruleExtractor.extractNumber('severity_level')
```
stephen-carter-at-sf: You are being too restrictive making this a number.
See 
https://github.com/forcedotcom/code-analyzer-core/blob/dev/packages/code-analyzer-core/src/config.ts#L163
and 
https://github.com/forcedotcom/code-analyzer-core/blob/dev/packages/code-analyzer-core/src/config.ts#L169

stephen-carter-at-sf (changes requested):

packages/code-analyzer-regex-engine/src/config.ts
ravipanguluri: Ah I see. How should I integrate this extractor in for now then? Should I just copy-paste it for now with a TODO? Or would you want to move this to somewhere it can be used in core and the regex engine?

ravipanguluri: Disregard the previous comment. It needs to be modified slightly for the engine anyways, so I have my answer.

ravipanguluri: Slacked you something along these lines: but it's actually somewhat tricky to write a validator with the current engine API as it exists. The cleanest way I can think of changing validateType<T> to something like the following:

```
function validateType<T>(expectedTypes: string[], value: unknown, fieldPath: string): T {
    const dataType = getDataType(value);

    if (!expectedTypes.includes(dataType)) {
        throw new Error(getMessage('ConfigValueMustBeOfType', fieldPath, expectedTypes.join(' or '), dataType))
    }
    
    return value as T;
}
```

Then change the calls to validateType (in validateObject, validateBoolean, etc.) so the expected types are all wrapped in an array. In this case, I could make a simple validator like so: 

```
static validateNumberOrString(value: unknown, fieldPath: string): number | string {
        return validateType<number | string>(["number", "string"], value, fieldPath);
    }
```

If this is good, I can go ahead with it. Otherwise, if you have a different idea just let me know.

stephen-carter-at-sf: what's wrong with the code in the links I sent. Can we not just copy/paste that validation code? (and in the future we can decide if we wanted to make it available via the engine-api or not)

```
function validateSeverityValue(value: unknown, valueKey: string): SeverityLevel {
    // Note that Object.values(SeverityLevel) returns [1,2,3,4,5,"Critical","High","Moderate","Low","Info"]
    if ((typeof value !== 'string' && typeof value !== 'number')
        || !Object.values(SeverityLevel).includes(value as string | number)) {
        throw new Error(getMessage('ConfigValueNotAValidSeverityLevel', valueKey,
            JSON.stringify(Object.values(SeverityLevel)), JSON.stringify(value)));
    }
    if (typeof value === 'string') {
        // We can't type cast to enum from a string, so instead we choose the enum based on the string as a key.
        value = SeverityLevel[value as keyof typeof SeverityLevel];
    }
    // We can type cast to enum safely from a number
    return value as SeverityLevel;
}
```

ravipanguluri: In this proposed solution I am using the validateSeverityValue function as-is. But i'm creating an extra layer to extract the number/string value in the config field before validating that it's a valid severity value. I can just use extractString()/extractNumber() to do this, but i'd have to verify the type of the value at that configuration field before calling the functions because they throw an error when you give them the wrong type.

packages/code-analyzer-regex-engine/src/config.ts
```diff
@@ -1,5 +1,5 @@
 import {
-    ConfigObject,
+    ConfigObject, ConfigValue,
```
stephen-carter-at-sf: nitpick, put each entry on its own line

packages/code-analyzer-regex-engine/src/config.ts
```diff
@@ -37,13 +37,25 @@ export type RegexRules = {
         // The message emitted when a rule violation occurs. This message is intended to help the user understand the violation.
         // Default: `A match of the regular expression {regex} was found for rule '{ruleName}': {description}`
         violation_message: string;
+
+        // A list of labels to help categorize the rule. Rules can also be filtered by these tags.
+        // Default: ['Recommended']
+        tags: string[];
+
+        // Determines the severity level of a rule. You can assign a severity level to a rule to indicate how important the rule is.
+        // The severity levels correspond to numbers as follows: Critical = 1, High = 2, Moderate = 3, Low = 4, Info = 5
```
stephen-carter-at-sf: Instead of defining what tags are or what a severity level is... let's assume that they know. I slack messaged you with a link to a document that contains my updates to these comments.

stephen-carter-at-sf (changes requested):

packages/code-analyzer-regex-engine/src/config.ts
```diff
@@ -37,13 +37,25 @@ export type RegexRules = {
...
+        // Default: Moderate
+        severityLevel: SeverityLevel;
     }
 }
 
 export const FILE_EXT_PATTERN: RegExp = /^[.][a-zA-Z0-9]+$/;
 export const RULE_NAME_PATTERN: RegExp = /^[A-Za-z@][A-Za-z_0-9@\-/]*$/;
 export const REGEX_STRING_PATTERN: RegExp = /^\/(.*?)\/(.*)$/;
 
+export const DEFAULT_TAGS: string[] = ['Recommended']
```
stephen-carter-at-sf: nitpick; end with semi-colon please. Lines 56, 57, 73 and any others that are needed.

packages/code-analyzer-regex-engine/src/config.ts
ravipanguluri: I think auto-importing messes it up every time. I'll have to remind myself to check that before every PR.

packages/code-analyzer-regex-engine/src/config.ts
ravipanguluri: I'm not seeing the slack message. Did you send it previously?

packages/code-analyzer-regex-engine/src/config.ts
```diff
@@ -34,44 +38,58 @@ export type RegexRules = {
...
         violation_message: string;
+
+        // [Optional] The string array of tag values to apply to this rule by default.
+        // Default: ['Recommended']
+        tags: string[];
+
+        // [Optional] The severity level to apply to this rule by default.
+        // Possible values: 1 or 'Critical', 2 or 'High', 3 or 'Moderate', 4 or 'Low', 5 or 'Info'
+        // Default: 'Moderate'
+        severityLevel: SeverityLevel;
```
stephen-carter-at-sf: Oh wait... this should be severity instead of severityLevel so that we can be consistent with rule override section.

stephen-carter-at-sf (changes requested):

packages/code-analyzer-regex-engine/src/config.ts
stephen-carter-at-sf: I see that severityLevel is what the engineApi has on RuleDescription... and so it complicates things a bit. To resolve this for now, you'll need to set the value on both "severity" and "severityLevel" so as long as we have each of the RegexRules entries being an extended RuleDescription object for convenience. I have some ideas to clean this up afterward.

packages/code-analyzer-regex-engine/src/config.ts
```diff
@@ -34,44 +38,58 @@ export type RegexRules = {
...
-        const description: string = ruleExtractor.extractRequiredString('description')
-        const rawRegexString: string = ruleExtractor.extractRequiredString('regex')
-        const regex: RegExp = validateRegex(rawRegexString, ruleExtractor.getFieldPath('regex'))
+        const ruleExtractor: ConfigValueExtractor = customRulesExtractor.extractRequiredObjectAsExtractor(ruleName);
+        const description: string = ruleExtractor.extractRequiredString('description');
+        const rawRegexString: string = ruleExtractor.extractRequiredString('regex');
+        const regex: RegExp = validateRegex(rawRegexString, ruleExtractor.getFieldPath('regex'));
         const rawFileExtensions: string[] | undefined = ruleExtractor.extractArray('file_extensions',
             (element, fieldPath) => ValueValidator.validateString(element, fieldPath, FILE_EXT_PATTERN));
+        const rawSeverityLevelValue: ConfigValue = ruleExtractor.getObject()['severity_level'];
```
stephen-carter-at-sf: This should be 'severity' instead of 'severity_level' so that it matches the rule overrides section.

stephen-carter-at-sf (changes requested):

stephen-carter-at-sf (approved):

PR #71, opened 2024-08-10
CHANGE (RegexEngine) @W-16462482@ Make isTextFile utility async and cache results of isBinaryFile() calls

ravipanguluri:
In this PR:

* isTextFile() will run asynchronously instead of synchronously
* When a file is checked to see if it is a text file (and will be scanned by rules that iterate through all text files) results are cached so that they can be retrieved so we can reduce calls to costly isBinaryFile() function

### DIFF
--- a/packages/code-analyzer-regex-engine/src/engine.ts
+++ b/packages/code-analyzer-regex-engine/src/engine.ts
@@ -5,13 +5,14 @@ import {
     EngineRunResults,
     RuleDescription,
     RunOptions,
-    Violation
+    Violation,
+    Workspace
 } from "@salesforce/code-analyzer-engine-api";
 import path from "node:path";
 import fs from "node:fs";
 import os from "node:os";
 import {RegexRules} from "./config";
-import {isBinaryFileSync} from "isbinaryfile";
+import {isBinaryFile} from "isbinaryfile";
 
 const TEXT_BASED_FILE_EXTS = new Set<string>(
     [
@@ -23,6 +24,7 @@ const TEXT_BASED_FILE_EXTS = new Set<string>(
 export class RegexEngine extends Engine {
     static readonly NAME = "regex";
     readonly regexRules: RegexRules;
+    private readonly textFilesCache: Map<string, string[]> = new Map();
 
     constructor(regexRules: RegexRules) {
         super();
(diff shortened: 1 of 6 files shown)


packages/code-analyzer-regex-engine/test/end-to-end.test.ts
```diff
@@ -22,20 +22,19 @@ describe('End to end test', () => {
             custom_rules: {
                 "NoTodos": {
                     description: "Detects TODO comments in code",
-                    file_extensions: ['.ts', '.cls'],
```
ravipanguluri, line 25: I updated this test to verify that my cache works. I created a rule with no file extensions and pointed it to a workspace containing an ambiguous file (dummy1.py). When describeRules() was called, it should have cached that the file is a text file. Later, when runRules() checked if it should scan the file, it retrieved the cached result. I’m confident this worked because my code coverage is 100%, and no other test would trigger this retrieval.

packages/code-analyzer-regex-engine/src/engine.ts
```diff
@@ -47,7 +58,7 @@ export class RegexEngine extends Engine {
         const ruleDescriptions: RuleDescription[] = [];
 
         for (const regexRule of Object.values(this.regexRules)) {
-            if (fullFileList.some(fileName => this.shouldScanFile(fileName, regexRule.name))){
+            if (await this.shouldDescribeRule(fullFileList, regexRule.name)){
                 ruleDescriptions.push(regexRule);
             }
         }
```
stephen-carter-at-sf: Having an await in a loop doesn't help things all that much. Instead, we should be getting a promise array out and doing a Promise.all on it to then get a boolean array. This can then be used to quickly determine which ones to push.

stephen-carter-at-sf (changes requested):

packages/code-analyzer-regex-engine/src/engine.ts
```diff
@@ -38,20 +40,24 @@ export class RegexEngine extends Engine {
...
-        for (const regexRule of Object.values(this.regexRules)) {
-            if (fullFileList.some(fileName => this.shouldScanFile(fileName, regexRule.name))){
-                ruleDescriptions.push(regexRule);
-            }
-        }
-        return ruleDescriptions;
+        const ruleNames: string[] = Object.keys(this.regexRules);
+        const shouldDescribeResults: boolean[] = await Promise.all(ruleNames.map(ruleName => this.shouldDescribeRule(fullFileList, ruleName)));
+        const rules = Object.values(this.regexRules)
+        return rules.filter((_rule: RuleDescription, index: number) => shouldDescribeResults[index]);
```
stephen-carter-at-sf: I see we are losing our quick returns (like the if(...some...)). Same issue we would face if we converted lines 44 through 46 to use Promise.all.

I have some ideas, can you schedule a meeting with me to discuss how we can boost performance here without losing the quick returns?

packages/code-analyzer-regex-engine/src/engine.ts
```diff
@@ -5,13 +5,13 @@ import {
     EngineRunResults,
     RuleDescription,
     RunOptions,
-    Violation
+    Violation, Workspace
```
stephen-carter-at-sf: nitpick, new line between entries

stephen-carter-at-sf (approved):
Just one small nitpick... but otherwise approved. Good work.

PR #72, opened 2024-08-12
NEW: @W-15806485@: Add in HTML Format

stephen-carter-at-sf:


packages/code-analyzer-core/src/output-format.ts
```diff
@@ -144,6 +150,33 @@ class XmlOutputFormatter implements OutputFormatter {
...
 
+const HTML_TEMPLATE_VERSION: string = '0.0.1';
+const HTML_TEMPLATE_FILE: string = path.resolve(__dirname, '..', 'output-templates', `html-template-${HTML_TEMPLATE_VERSION}.txt`);
+class HtmlOutputFormatter implements OutputFormatter {
+    private static readonly TIMESTAMP_HOLE: string = '{{###TIMESTAMP###}}';
+    private static readonly RUNDIR_HOLE: string = '{{###RUNDIR###}}';
+    private static readonly VIOLATIONS_HOLE: string = '{{###VIOLATIONS###}}';
+    private readonly clock: Clock;
+
+    constructor(/* istanbul ignore next */ clock: Clock = new RealClock()) {
```
jag-j: Why are you creating a new instance of clock here?

packages/code-analyzer-core/src/output-format.ts
```diff
@@ -144,6 +150,33 @@ class XmlOutputFormatter implements OutputFormatter {
     }
 }
 
+const HTML_TEMPLATE_VERSION: string = '0.0.1';
```
jag-j, line 153: Nit: Will be good at some point to move these to constant.ts

packages/code-analyzer-core/src/output-format.ts
stephen-carter-at-sf: Instead of hard coding (new Date())... we pass in a clock. This way in my tests I can fix the clock to a specific date so that the tests are deterministic. For production, we always use a RealClock... but for tests we use a FixedClock.

packages/code-analyzer-core/src/code-analyzer.ts
```diff
@@ -132,6 +132,7 @@ export class CodeAnalyzer {
         const engineRunResultsList: EngineRunResults[] = await Promise.all(runPromises);
 
         const runResults: RunResultsImpl = new RunResultsImpl();
+        runResults._setClock(this.clock);
```
jfeingold35: `_setClock` is commented as being for tests only, and has the `_` prefix. Should we be invoking it in this production code?

stephen-carter-at-sf: In order to support passing down the clock that was given to the _setClock method of CodeAnalyzer (in tests)... I need to forward the clock to the runResults as well (again in tests). Do you have an alternative suggestion?

stephen-carter-at-sf: The more I think about it... since RunResultsImpl is not something exposed at for the user, let me just pass in the clock to the constructor here... since yeah - this is a code smell. Hold on...

stephen-carter-at-sf: Done.

packages/code-analyzer-core/src/output-format.ts
stephen-carter-at-sf: I see your point, but right now it would be the only one in a separate constants file. I think this is fine for now.

packages/code-analyzer-core/src/output-format.ts
stephen-carter-at-sf: I updated this so that I didn't create a default instance here per your suggestion.

jag-j (approved):

PR #73, opened 2024-08-12
NEW (FlowTest) @W-16360741@ FlowTest engine validates local versions of Python.

jfeingold35:


packages/code-analyzer-flowtest-engine/src/config.ts
jfeingold35, line 1: Much of this file was derived from looking at the regex engine.

packages/code-analyzer-flowtest-engine/src/engine.ts
```diff
@@ -2,30 +2,54 @@ import {
...
 export class FlowTestEngine extends Engine {
     public static readonly NAME: string = 'flowtest';
+    private readonly config: FlowTestConfig;
+    private readonly pythonVersionManager: PythonVersionManager;
+
+    public constructor(config: FlowTestConfig, dependencies: FlowTestEngineDependencies) {
+        super();
+        this.config = config;
+        const emitLogEventFunction = (l: LogLevel, m: string) => this.emitLogEvent(l, m);
+        this.pythonVersionManager = new PythonVersionManager(dependencies.pythonVersionIdentifier, emitLogEventFunction);
```
jfeingold35: Constructing the version manager within the class itself so I can pass the `emitLogEvent` function to the constructor instead of the `getPythonCommand` method.

packages/code-analyzer-flowtest-engine/src/lib/python-versioning/PythonVersionIdentifier.ts
```diff
@@ -0,0 +1,56 @@
...
+}
+
+export class PythonVersionIdentifierImpl implements PythonVersionIdentifier{
+    /**
+     * Resolves to a {@link SemVer} representing the version of the Python instance invoked by the provided
+     * command
+     * @param pythonCommand
+     */
+    public identifyPythonVersion(pythonCommand: string): Promise<SemVer> {
+        return new Promise<SemVer>((res, rej) => {
```
jfeingold35: The internals of this Promise may get refactored into a superclass for use when actually running the Python tool, but for now it's all in here.

packages/code-analyzer-flowtest-engine/src/plugin.ts
```diff
@@ -1,18 +1,22 @@
...
-    public async createEngine(engineName: string, _engineConfig: ConfigObject): Promise<Engine> {
+    public async createEngine(engineName: string, engineConfig: ConfigObject): Promise<Engine> {
         if (engineName !== FlowTestEngine.NAME) {
             throw new Error(getMessage('CantCreateEngineWithUnknownName', engineName));
         }
-        return new FlowTestEngine();
+        return new FlowTestEngine(validateAndNormalizeConfig(engineConfig), {
+            pythonVersionIdentifier: new PythonVersionIdentifierImpl()
+        });
     }
```
stephen-carter-at-sf: We should be validating the python version during createEngine. If python doesn't exist on the users machine, then we shouldn't be adding the engine to the list when selecting rules or running etc. So all validation should be done in createEngine if possible. This way, when the engine is constructed, it receives the normalized config with all validation done - so that it doesn't need to make validation its responsibility.
This also allows us to eventually move utilities like these into the engine api officially as an extractor method. I don't know if python will ever be used again, but for example, when we get to fetching the java version. It'll be nice to share that extraction routine among the SFGE and PMD engines.

All that to say, I recommend, you, as part of taking the raw config and converting it to a normalized config... have the normalized config have the derived values from the environment whenever the raw config doesn't supply it.

stephen-carter-at-sf (changes requested):

packages/code-analyzer-flowtest-engine/src/config.ts
```diff
@@ -0,0 +1,17 @@
+import {ConfigObject, ConfigValueExtractor} from "@salesforce/code-analyzer-engine-api";
+
+
+export type FlowTestConfig = {
+    /**
+     * Optional. Indicates the specific Python command to use. If none is supplied, the engine will try to find one.
+     * E.g., {@code python}, @{code /Library/Frameworks/Python.framework/Versions/3.12/bin/python3}.
+     */
+    python_command_path?: string;
```
stephen-carter-at-sf: We should aim to have no optional values here. We should take the optional raw config and always (whenever possible) convert it to exact values. This way, we remove conditional code from the engine.

stephen-carter-at-sf (changes requested):

packages/code-analyzer-flowtest-engine/src/config.ts
```diff
@@ -0,0 +1,72 @@
...
+     * E.g., {@code python}, @{code /Library/Frameworks/Python.framework/Versions/3.12/bin/python3}.
+     */
+    [PYTHON_COMMAND_PATH]: string;
+}
+
+type FactoryDependencies = {
+    pythonVersionIdentifier: PythonVersionIdentifier;
+}
+
+export interface FlowTestConfigFactory {
```
jfeingold35: The config and its factory share the same file because it seemed excessive to have a file dedicated to one type.

packages/code-analyzer-flowtest-engine/src/lib/python/PythonVersionIdentifier.ts
```diff
@@ -0,0 +1,49 @@
+import {SemVer, coerce} from 'semver';
+import * as cp from 'node:child_process';
+
+export interface PythonVersionIdentifier {
+    identifyPythonVersion(pythonCommand: string): Promise<SemVer|null>;
```
jfeingold35: CHANGE: This method now returns null when a version can't be parsed instead of throwing an error here. It makes the method signature a little more complicated but it keeps the upstream code much cleaner (and honestly simplifies some stuff here too).

packages/code-analyzer-flowtest-engine/src/messages.ts
jfeingold35, line 1: Messages have been changed aggressively.

packages/code-analyzer-flowtest-engine/src/plugin.ts
```diff
@@ -1,18 +1,28 @@
...
         if (engineName !== FlowTestEngine.NAME) {
             throw new Error(getMessage('CantCreateEngineWithUnknownName', engineName));
         }
-        return new FlowTestEngine();
+        const configFactory: FlowTestConfigFactory = this._getConfigFactory();
+        return new FlowTestEngine(await configFactory.create(engineConfig));
+    }
+
+    /* istanbul ignore next: Difficult to test */
+    public _getConfigFactory(): FlowTestConfigFactory {
```
jfeingold35: Calling attention to this part here. I couldn't think of a better way to force the plugin to use a stubbed factory than by having this method be publicly exposed and stubbed in tests.

packages/code-analyzer-flowtest-engine/stubs/lib/python/StubPythonVersionIdentifier.ts
jfeingold35, line 1: Stubs are now in their own top-level-rooted directory whose structure mirrors `src`. I like this convention, and may implement it on scanner v5.

packages/code-analyzer-flowtest-engine/test/config.test.ts
jfeingold35, line 1: Tests here are equivalent to the ones that used to be in `PythonVersionManager.test.ts`.

packages/code-analyzer-flowtest-engine/src/plugin.ts
stephen-carter-at-sf: I recommend passing in dependencies to constructor. For example
```
    private configNormalizer: FlowTestConfigNormalizer;

    public constructor(pythonVersionIdentifier: PythonVersionIdentifier = new RuntimePythonVersionIdentifier()) {
        this.configNormalizer = new ConfigNormalizer(pythonVersionIdentifier);
    }
...

    public async createEngine(engineName: string, rawConfig: ConfigObject): Promise<Engine> {
        ...
        const normalizedConfig: FlowTestConfig = await this.configNormalizer.normalize(rawConfig);
        return new FlowTestEngine(normalizedConfig);
    }
```

Note that in my version I renamed your `PythonVersionIdentifierImpl` to be `RuntimePythonVersionIdentifier`. Also I renamed your `FlowTestConfigFactory` to be `ConfigNormalizer` (I don't think you need an interface here) since I think all engines will basically have this normalization step at some point so I like the name. Alternative names could be ConfigValidator or ConfigResolver. I'm no the fence here.

These are just suggestions, but I hope that helps.

packages/code-analyzer-flowtest-engine/stubs/stub-config.ts
```diff
@@ -0,0 +1,16 @@
+import {ConfigObject} from '@salesforce/code-analyzer-engine-api';
+import {FlowTestConfig, FlowTestConfigFactory} from '../src/config';
+
+export class StubFlowTestConfigFactory implements FlowTestConfigFactory {
```
stephen-carter-at-sf: What is this for? We can just stub out the Version Identifier can't we?

Oh wait... I see now... your PythonVersionIdentifer should probably instead be a class that gives both... the version and the path. Then you just need to stub 1 class.

Maybe call the interface PythonVersionExtractor can find the python execution path (if it hasn't been provided), can validate a python execution path (if it has been provided), and can give a python version based on a path.

packages/code-analyzer-flowtest-engine/src/config.ts
```diff
@@ -0,0 +1,68 @@
...
+    /**
+     * Indicates the specific Python command to use. Can be specified on the ConfigObject, or inferred automatically.
+     * E.g., {@code python}, @{code /Library/Frameworks/Python.framework/Versions/3.12/bin/python3}.
+     */
+    [PYTHON_COMMAND_PATH]: string;
+}
+
+type FactoryDependencies = {
+    pythonVersionIdentifier: PythonVersionIdentifier;
+}
```
stephen-carter-at-sf: Not sure why this dependencies type is needed. Why not just store the pythonVersionIdentifier on the ConfigNormalizer directly?

packages/code-analyzer-flowtest-engine/test/config.test.ts
```diff
@@ -0,0 +1,95 @@
+import {SemVer} from "semver";
+import {ConfigObject} from "@salesforce/code-analyzer-engine-api";
+import {FlowTestConfig, ConfigNormalizer, PYTHON_COMMAND_PATH} from "../src/config";
+import {StubPythonVersionIdentifier} from "../stubs/lib/python/StubPythonVersionIdentifier";
+import {getMessage} from "../src/messages";
+import {changeWorkingDirectoryToPackageRoot} from "./test-helpers";
+
+changeWorkingDirectoryToPackageRoot();
+
+describe('FlowTestConfigFactory implementations', () => {
```
stephen-carter-at-sf: Can we instead test through the front door (i.e. the plugin's createEngine method) so that we don't couple our tests to implementation details. This is the reason why the plugin receives in its constructor its dependencies.

packages/code-analyzer-flowtest-engine/test/lib/python/PythonVersionIdentifier.test.ts
```diff
@@ -0,0 +1,28 @@
+import {SemVer} from 'semver';
+import {RuntimePythonVersionIdentifier} from '../../../src/lib/python/PythonVersionIdentifier';
+
+describe('PythonVersionIdentifier implementations', () => {
+    describe('PythonVersionIdentifierImpl', () => {
+
+        it('When command outputs parseable version, resolves to that version', async () => {
+            const identifier = new RuntimePythonVersionIdentifier();
+            // NOTE: We can't guarantee that the current machine has Python on it, but we _can_ guarantee that it has Node.
+            //       So we'll tell it to provide Node's version, and then just compare that to the version of this node process.
```
stephen-carter-at-sf: Why can't we? We can add it to our test environment's github workflow.

packages/code-analyzer-flowtest-engine/test/plugin.test.ts
```diff
@@ -3,24 +3,31 @@ import {FlowTestEnginePlugin} from "../src";
...
 
     it('#createEngine() returns a FlowTestEngine when asked for one', async () => {
+        const plugin: EnginePluginV1 = new FlowTestEnginePlugin(new StubPythonVersionIdentifier(new Map([
+            // Simulate the machine having a viable version of Python on it.
+            ['python3', new SemVer('3.10.0')]
+        ])));
         expect(await plugin.createEngine('flowtest', {})).toBeInstanceOf(FlowTestEngine);
     });
 
     it('#createEngine() throws error when asked for an unsupported engine', async () => {
```
stephen-carter-at-sf: I was expecting all of the configuration testing to be done here as well instead of in the config.test.ts file since the public entry point for core (the client of this engine) is the createEngine method.

packages/code-analyzer-flowtest-engine/tsconfig.json
```diff
@@ -7,6 +7,7 @@
   },
   "include": [
     "./src",
-    "./test"
+    "./test",
+    "./stubs"
```
stephen-carter-at-sf: Why not put stubs as a subdirectory underneath the test directory? Or better yet inside of the test directory simply as a file called stubs.ts

stephen-carter-at-sf (changes requested):

packages/code-analyzer-flowtest-engine/src/config.ts
jfeingold35: It seemed like better future-proofing in case we need to start passing in more things. But yeah I could just pass the identifier in directly.

packages/code-analyzer-flowtest-engine/test/lib/python/PythonVersionIdentifier.test.ts
jfeingold35: I meant we shouldn't necessarily assume that a _dev's_ machine has Python on it...

packages/code-analyzer-flowtest-engine/test/config.test.ts
jfeingold35: I'm not sure I understand. In what way is this more coupled to the implementation than the alternative?

packages/code-analyzer-flowtest-engine/tsconfig.json
jfeingold35: Having them all in one file might work here, but it's definitely not workable for the other projects. And having `stubs` at the top level seemed cleaner to me, though I'm happy to move it back down too.

packages/code-analyzer-flowtest-engine/src/config.ts
jfeingold35: Done.

packages/code-analyzer-flowtest-engine/test/config.test.ts
stephen-carter-at-sf: If you look at eslint, regex, engines for example... you'll see that the tests don't instantiate any helper tools or classes... instead it always uses the plugin interface. This way as we keep updating the implementation and refactoring - we never need to update the tests.

Adding in a _getFlowTestConfig method on the engine is helpful here.

So each test is a plugin test. You construct the plugin for test purposes (passing in any test specific stubs or dependencies)... and then call createEngine('flowtest',rawConfig) and validate what you expect... which might be simply validating that the engine that you have received has the correct config handed to it (in which you can use test only engine._getFlowTestConfig() method).

packages/code-analyzer-flowtest-engine/test/lib/python/PythonVersionIdentifier.test.ts
stephen-carter-at-sf: I suppose... it's just a funny test in my opinion using node instead of python. I think it is reasonable to assume that developers will have python 3+ on their machine. But if you are truly worried then maybe we make this code conditional on an environment variable that we set inside of github. Just thinking out loud. I would love to hear Jag's thoughts on this one to be honest - because i'm on the fence.

packages/code-analyzer-flowtest-engine/tsconfig.json
stephen-carter-at-sf: I think we should be cautious with premature optimization in this case. I don't think the flowtest is going to have a lot of stubs. In fact, i would argue that a stub should be in the file it is used in unless it is shared across files. So I personally would just put the stub at the bottom of your plugin.test.ts file

jfeingold35: Yeah, I'm pretty sure most of the stubs are only going to be used in one test file, so I'll move the one I have into that test file for now.

packages/code-analyzer-flowtest-engine/test/lib/python/PythonVersionIdentifier.test.ts
jfeingold35: @jag-j , do you have thoughts on this?

packages/code-analyzer-flowtest-engine/tsconfig.json
jfeingold35: Actually never mind. The stubs are used in multiple files. So I'll just move them into a `test/stubs/` folder and have that folder be flat.

stephen-carter-at-sf (approved):

PR #74, opened 2024-08-15
FIX(eslint): @W-16509878@: Prevent reading babel config files from LWC base config

stephen-carter-at-sf:
... and consolidate each fine log even to 1 line each for now (until we add in the ability to control logging better).

### DIFF
--- a/packages/code-analyzer-eslint-engine/src/base-config.ts
+++ b/packages/code-analyzer-eslint-engine/src/base-config.ts
@@ -67,6 +67,8 @@ export class LegacyBaseConfigFactory {
             parserOptions: {
                 requireConfigFile: false,
                 babelOptions: {
+                    babelrc: false,
+                    configFile: false,
                     parserOpts: {
                         plugins: [
                             "classProperties",
--- a/packages/code-analyzer-eslint-engine/src/engine.ts
+++ b/packages/code-analyzer-eslint-engine/src/engine.ts
@@ -103,8 +103,8 @@ export class ESLintEngine extends Engine {
                 if (ruleNames.has(ruleName)) {
                     violations.push(toViolation(eslintResult.filePath, resultMsg));
                 } else {
-                    // This theoretically should never be possible. But for sanity sakes, we'll emit a warning if this ever happens.
-                    this.emitLogEvent(LogLevel.Warn, getMessage('ViolationFoundFromUnregisteredRule', ruleName, JSON.stringify(violation,null,2)))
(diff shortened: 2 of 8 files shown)


packages/code-analyzer-eslint-engine/src/base-config.ts
```diff
@@ -67,6 +67,8 @@ export class LegacyBaseConfigFactory {
             parserOptions: {
                 requireConfigFile: false,
                 babelOptions: {
+                    babelrc: false,
+                    configFile: false,
```
stephen-carter-at-sf, line 71: This essentially is the fix.

packages/code-analyzer-eslint-engine/src/engine.ts
```diff
@@ -103,8 +103,8 @@ export class ESLintEngine extends Engine {
                 if (ruleNames.has(ruleName)) {
                     violations.push(toViolation(eslintResult.filePath, resultMsg));
                 } else {
-                    // This theoretically should never be possible. But for sanity sakes, we'll emit a warning if this ever happens.
-                    this.emitLogEvent(LogLevel.Warn, getMessage('ViolationFoundFromUnregisteredRule', ruleName, JSON.stringify(violation,null,2)))
+                    // This may be possible if a user tries to suppress an eslint rule in their code that isn't available. We just ignore it but debug it just in case.
+                    this.emitLogEvent(LogLevel.Debug, getMessage('ViolationFoundFromUnregisteredRule', ruleName, JSON.stringify(violation,null,2)))
```
stephen-carter-at-sf, line 107: Figured out how this could happen. Basically, if a user in a file has an eslint ignore statement for a rule that doesn't exist anywhere in eslint, then we used to throw a warning. This to me can be ignored, and so I reduced this to a debug statement in the logs.

packages/code-analyzer-eslint-engine/src/strategy.ts
```diff
@@ -177,6 +175,7 @@ export class LegacyESLintStrategy implements ESLintStrategy {
         return {
             cwd: __dirname, // This is needed to make the base plugins discoverable. Don't worry, user's plugins are also still discovered.
             errorOnUnmatchedPattern: false,
+            reportUnusedDisableDirectives: 'off',
```
stephen-carter-at-sf, line 178: This didn't seem to change anything... but I figured might as well explicitly turn it make it equal to 'off' anyway since that is what we want.

jfeingold35 (approved):

PR #75, opened 2024-08-16
FIX: @W-16515074@: escape html characters for text fields in html report

stephen-carter-at-sf:


packages/code-analyzer-core/src/output-format.ts
```diff
@@ -188,33 +188,32 @@ function toResultsOutput(results: RunResults) {
...
 }
 
-function toViolationOutputs(violations: Violation[], runDir: string): ViolationOutput[] {
-    return violations.map(v => createViolationOutput(v, runDir));
+function toViolationOutputs(violations: Violation[], runDir: string, sanitizeFcn: (text: string) => string = t => t): ViolationOutput[] {
+    return violations.map(v => createViolationOutput(v, runDir, sanitizeFcn));
 }
 
-function createViolationOutput(violation: Violation, runDir: string): ViolationOutput {
+function createViolationOutput(violation: Violation, runDir: string, sanitizeFcn: (text: string) => string): ViolationOutput {
```
jag-j, line 199: Why are you passing the escaptHtml function through multiple levels and using it in this function instead of directly using it from here?

stephen-carter-at-sf: Because those levels are used by all formats including JSON and CSV for example. Only html format needs/should sanitize with escapeHtml. (edited) 
Thankfully XML gets santized by the 3rd party util we are using… so I don’t need to escape there either… but if we switch our implementation or if we introduce new formats that needs sanitization… then we can easily just pass in its own sanitization function.

jag-j (approved):

PR #80, opened 2024-08-28
NEW: @W-16605635@: Refactor engine-api and core to return resolved config and descriptions

stephen-carter-at-sf:


packages/code-analyzer-core/src/messages.ts
```diff
@@ -21,20 +59,29 @@ const MESSAGE_CATALOG : MessageCatalog = {
...
+    PluginErrorWhenCreatingEngine:
+        `Failed to create engine with name '%s' due to the following error:\n` +
         `%s\n\n` +
         `If you wish to ignore this error and disable this engine, then update your Code Analyzer configuration with:\n` +
         `engines:\n` +
         `  %s:\n` +
         `    disable_engine: true\n`,
 
+    PluginErrorWhenCreatingEnginePriorToDisableEngineCheck:
+        `Failed to create engine with name '%s' due to the following error:\n%s\n`,
```
jfeingold35: It feels a little odd for these to be separate messages. Maybe you could have "Failed to create..." be one message and "If you wish to ignore this error and disable this engine..." be a second message, and then you append the second to the first in situations where it's relevant?

packages/code-analyzer-core/src/code-analyzer.ts
```diff
@@ -220,28 +238,41 @@ export class CodeAnalyzer {
...
+        try {
+            const engineConfigDescription: engApi.ConfigDescription = enginePluginV1.describeEngineConfig(engineName);
+            addDisableEngineFieldDescription(engineConfigDescription, engineName);
+            this.engineConfigDescriptions.set(engineName, engineConfigDescription);
+        } catch (err) {
+            throw new Error(getMessage('PluginErrorWhenCreatingEnginePriorToDisableEngineCheck', engineName, (err as Error).message));
+        }
+
+        const engineOverrides: EngineOverrides = this.config.getEngineOverridesFor(engineName);
+        if (engineOverrides[FIELDS.DISABLE_ENGINE]) {
```
jfeingold35, line 250: Can you add a comment here indicating that we use the override itself in place of the config for a disabled engine?
Seems like something that will save us grief later on.

packages/code-analyzer-core/src/messages.ts
```diff
@@ -1,6 +1,44 @@
 import {getMessageFromCatalog, MessageCatalog} from "@salesforce/code-analyzer-engine-api";
 
 const MESSAGE_CATALOG : MessageCatalog = {
+    ConfigOverview:
+        `CODE ANALYZER CONFIGURATION\n` +
+        `To learn more about this configuration, visit: __LINK_COMING_SOON__`,
+
+    ConfigFieldDescription_config_root:
+        `The absolute folder path to which all other path values in this configuration may be relative to.\n` +
+        `If not specified or if equal to null, then the value is automatically chosen to be the parent folder of your Code Analyzer\n` +
```
jfeingold35: Perhaps "if not specified, or if specified as null", so it's clear that the null value comes from the user?

packages/code-analyzer-core/src/messages.ts
```diff
@@ -1,6 +1,44 @@
...
+        `configuration file if it exists, or the current working directory otherwise.`,
+
+    ConfigFieldDescription_log_folder:
+        `Folder where to store log files. May be an absolute path or a path relative to config_root.\n` +
+        `If not specified or if equal to null, then the value is automatically chosen to be your machine's default temporary directory.`,
```
jfeingold35: "If not specified, or if specified as null"?

packages/code-analyzer-core/src/config.ts
```diff
@@ -108,6 +111,10 @@ export class CodeAnalyzerConfig {
         this.config = config;
     }
 
+    public getConfigDescription(): ConfigDescription {
```
jfeingold35: I think this method makes more sense to be static.

packages/code-analyzer-engine-api/src/engine-plugins.ts
```diff
@@ -9,11 +9,45 @@ export interface EnginePlugin  {
...
     abstract getAvailableEngineNames(): string[]
 
-    abstract createEngine(engineName: string, config: ConfigObject): Promise<Engine>
+    /**
+     * Returns a configuration description object that helps describe the configuration and its fields
+     * Note to plugin author: This method should be overridden to return the configuration that you wish to expose to the client.
+     *                        Otherwise, an empty configuration description will be returned.
+     * @param engineName The name of the engine
+     */
+    describeEngineConfig(engineName: string): ConfigDescription { // eslint-disable-line @typescript-eslint/no-unused-vars
```
jfeingold35, line 32: Why is this method concrete instead of abstract? Is it just so that engines always have an implementation even if the author is too lazy to provide one? If so, then I think we should still make it abstract so that the author has to be _deliberately_ lazy. If they want an empty object, that's between them and God, but they should have to _choose_ it.

packages/code-analyzer-engine-api/test/config.test.ts
```diff
@@ -185,44 +185,15 @@ describe("Tests for ConfigValueExtractor", () => {
...
+        expect(extractor.getConfigRoot()).toEqual(process.cwd());
     });
 
-    it("When config_root is supplied with valid folder, then extractConfigRoot returns it", () => {
-        const config: ConfigObject = {config_root: __dirname};
-        const extractor: ConfigValueExtractor = new ConfigValueExtractor(config);
-        expect(extractor.extractConfigRoot()).toEqual(__dirname);
-    });
-
-    it("When config_root value is not a string, then extractConfigRoot throws an error", () => {
```
jfeingold35, line 200: Where did these tests get moved to?

packages/code-analyzer-core/src/messages.ts
stephen-carter-at-sf: Good idea. Done.

packages/code-analyzer-core/src/code-analyzer.ts
stephen-carter-at-sf: Actually you raise a good point here... This code was living under the assumption that we were using the engine config as the source of truth minus the internalUseOnlyFields. But since we switched to make the config descriptions the source of truth... now the cli will need when implementing the config command to check whether the field even exists on the engine config when the engine is disabled... and either make the values null (which means unset) or just not add them to the config. But yeah i'll add a comment here (even though I have an explicit test for this already that serves as a comment). So definitely add a test in the config command for this (disable an engine and have just a few of its config values set).

packages/code-analyzer-core/src/messages.ts
stephen-carter-at-sf: Good idea. Done.

packages/code-analyzer-core/src/messages.ts
stephen-carter-at-sf: Good idea. Done.

packages/code-analyzer-core/src/config.ts
stephen-carter-at-sf: Yeah I can't come up with a good reason why not (other than maybe in the future the field descriptions will be dependent on the values - but that isn't the case right now)... so i'll make the change.

packages/code-analyzer-engine-api/src/engine-plugins.ts
stephen-carter-at-sf: Typically when a class is abstract, you only make the methods abstract if there isn't a reasonable default behavior and if the method is something every subclass needs to implement. In our case, plugins don't need to have configurations. Only plugins that have configurations will implement createEngineConfig and only plugins that want to expose their configurations will implement describeEngineConfig. So I think keeping to having a default implementation makes sense here. For example, retire-js doesn't need these methods at all.

packages/code-analyzer-engine-api/test/config.test.ts
stephen-carter-at-sf: We basically had 2 sets of tests before - so these duplicate tests at the engine-api layer got removed when I moved the validation of config_root back into the core module - which also had config_root tests.

Since the config root validation is solely owned by the core module once again... you'll find all these tests on lines 225 to 244 of code-analyzer-core/test/config.test.ts.

packages/code-analyzer-core/src/code-analyzer.ts
stephen-carter-at-sf: Done.

packages/code-analyzer-core/src/config.ts
stephen-carter-at-sf: Done.

jfeingold35 (approved):

PR #82, opened 2024-09-02
NEW(pmd): @W-16113598@ Implement describeRules method for PMD's default standard rules

stephen-carter-at-sf:


packages/code-analyzer-pmd-engine/pmd-wrapper/src/main/java/com/salesforce/sfca/pmdwrapper/PmdWrapper.java
```diff
@@ -1,18 +1,59 @@
...
+        } else {
+            throw new RuntimeException("Bad first argument to PmdWrapper. Expected \"describe\" or \"run\". Received: \"" + args[0] + "\"");
+        }
+
+        long endTime = System.currentTimeMillis();
+        System.out.println("END OF CALL TO \"PmdWrapper\": " + (endTime - startTime) + " milliseconds");
+    }
+
+    private static void invokeDescribeCommand(String[] args) {
+        if (args.length != 3) {
```
jag-j, line 43: Nit: Can we move all the argument checks to main (lines 30 to 36)?

jag-j (approved):

packages/code-analyzer-pmd-engine/pmd-wrapper/src/test/java/com/salesforce/sfca/pmdwrapper/PmdWrapperTest.java
```diff
@@ -1,18 +1,81 @@
...
+        String stdOut = callPmdWrapper(args);
+
+        // Read the file and convert the json back into a list of PmdRuleInfo instances
+        String fileContents = Files.readString(tempFile);
+        Gson gson = new Gson();
+        Type pmdRuleInfoListType = new TypeToken<List<PmdRuleInfo>>(){}.getType();
+        List<PmdRuleInfo> pmdRuleInfoList = gson.fromJson(fileContents, pmdRuleInfoListType);
+
+        // Keeping assertions to a minimal since we will lock in most of the output details in our other tests
+        assertThat(pmdRuleInfoList, hasSize(greaterThan(60))); // Actual amount is 65 but keeping this test flexible
```
jfeingold35: In the spirit of keeping the test flexible, we should remove the hardcoded number from the comment.

packages/code-analyzer-pmd-engine/src/constants.ts
```diff
@@ -0,0 +1,40 @@
...
+export const extensionToPmdLanguage: Record<string, PmdLanguage> = {
+    // (Apex - 'apex') Language Model: https://github.com/pmd/pmd/blob/master/pmd-apex/src/main/java/net/sourceforge/pmd/lang/apex/ApexLanguageModule.java
+    '.cls': PmdLanguage.APEX,
+    '.trigger': PmdLanguage.APEX,
+
+    // (Java - 'java') Language Model: https://github.com/pmd/pmd/blob/master/pmd-java/src/main/java/net/sourceforge/pmd/lang/java/JavaLanguageModule.java
+    '.java': PmdLanguage.JAVA,
+
+    // (JavaScript - 'ecmascript') Language Model: https://github.com/pmd/pmd/blob/master/pmd-javascript/src/main/java/net/sourceforge/pmd/lang/ecmascript/EcmascriptLanguageModule.java
+    '.js': PmdLanguage.JAVASCRIPT,
```
jfeingold35, line 25: Do we need to support `.ts`, `.jsx`, or other ECMAScript extensions?

packages/code-analyzer-pmd-engine/pmd-wrapper/src/main/java/com/salesforce/sfca/pmdwrapper/PmdWrapper.java
stephen-carter-at-sf: I'm hesitant to do so because I'd like to keep the main function as a quick switchyard to the invoke functions... leaving the main function to only contain the common functionality across all commands. If I move lines 43 to 45... then id want to move lines 46 to 47 as well.

As we add in the run command and anything else in the future... I'd hate for the complexity of each method to increase. An easy measure of complexity is to see how many conditionals inside of conditionals there are. If I move the argument validation of each command inside of the top level main's conditional statements... that increases the mccabe cyclomatic complexity. Whenever possible I like to have a single level of conditions in any given method. So this pattern that I have seems most beneficial for those reasons.

Besides attempting to put all argument handing in one method, is there a reason for moving these lines up to main?

packages/code-analyzer-pmd-engine/src/utils.ts
```diff
@@ -0,0 +1,58 @@
...
+    constructor(javaCommand: string = 'java') {
+        this.javaCommand = javaCommand;
+    }
+
+    async exec(javaCmdArgs: string[], javaClassPaths: string[] = [], stdoutCallback: (stdOutMsg: string) => void = (_m: string) => {}): Promise<void> {
+        return new Promise<void>((resolve, reject) => {
+            const stderrMessages: string[] = [];
+            const allJavaArgs: string[] = javaClassPaths.length == 0 ? javaCmdArgs :
+                ['-cp', javaClassPaths.join(path.delimiter), ... javaCmdArgs];
+            const javaProcess: ChildProcessWithoutNullStreams = spawn('java', allJavaArgs);
```
jfeingold35, line 31: Observation: The Java version is no longer configurable. Do we anticipate needing to explain to anyone why we made this decision? If so, what's our rationale?

packages/code-analyzer-pmd-engine/src/pmd-engine.ts
```diff
@@ -0,0 +1,140 @@
...
+    priority: string,
+    ruleSetFile: string,
+    class: string
+}
+
+export class PmdEngine extends Engine {
+    static readonly NAME: string = "pmd";
+
+    // TODO: Eventually these will be configurable
+    private readonly availableLanguages: PmdLanguage[] = [PmdLanguage.APEX, PmdLanguage.VISUALFORCE];
```
jfeingold35, line 33: So right now the Java, JS, etc rules are present but invisible to the TS layer?

packages/code-analyzer-pmd-engine/src/constants.ts
stephen-carter-at-sf: If you look at https://github.com/pmd/pmd/blob/master/pmd-javascript/src/main/java/net/sourceforge/pmd/lang/ecmascript/EcmascriptLanguageModule.java you'll find that it only supports '.js' files. We can't control this... when we send over a workspace - it determines whether to run the rule based on the language which is based on the file extension.

If you are asking about typescript though... there is https://github.com/pmd/pmd/blob/master/pmd-javascript/src/main/java/net/sourceforge/pmd/lang/typescript/TsLanguageModule.java which is something we can discuss at some point. But first, let's get feature parity of what we offer today.

That is we can offer typescript in the future... but I didn't support it yet because it has no rules for PMD... it's only there for CPD currently.

packages/code-analyzer-pmd-engine/pmd-wrapper/src/main/java/com/salesforce/sfca/pmdwrapper/PmdWrapper.java
jag-j: No other reason. I thought it would be nice to have all arg related checks in one place. This makes sense. You can leave it as is.

packages/code-analyzer-pmd-engine/src/utils.ts
stephen-carter-at-sf: That is a future work item. No configuration abilities with this part. So yes we will get there... notice the comment on  line 108 of pmd-engine.ts

packages/code-analyzer-pmd-engine/src/pmd-engine.ts
stephen-carter-at-sf: Yes... 1 step at a time. Will add in configuration abilities in a future work item. It was easy enough to just add in that ability in the JAVA layer first to prepare.

jfeingold35 (approved):

PR #85, opened 2024-09-05
CHANGE: @W-16606152@: Normalize engine configs to better support the config command for the cli

stephen-carter-at-sf:
* Make the disable_engine field appear first in the field descriptions
* Ensure every engine has an engine overview

### DIFF
--- a/packages/code-analyzer-core/src/code-analyzer.ts
+++ b/packages/code-analyzer-core/src/code-analyzer.ts
@@ -240,7 +240,7 @@ export class CodeAnalyzer {
 
         try {
             const engineConfigDescription: engApi.ConfigDescription = enginePluginV1.describeEngineConfig(engineName);
-            addDisableEngineFieldDescription(engineConfigDescription, engineName);
+            normalizeEngineConfigDescription(engineConfigDescription, engineName);
             this.engineConfigDescriptions.set(engineName, engineConfigDescription);
         } catch (err) {
             throw new Error(getMessage('PluginErrorWhenCreatingEngine', engineName, (err as Error).message));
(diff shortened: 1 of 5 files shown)


packages/code-analyzer-core/src/code-analyzer.ts
```diff
@@ -524,10 +524,15 @@ function isIntegerBetween(value: number, leftBound: number, rightBound: number):
...
+function normalizeEngineConfigDescription(engineConfigDescription: ConfigDescription, engineName: string): void {
+    // Every engine config should have an overview, so if missing, then we add in a generic one
+    if (!engineConfigDescription.overview) {
+        engineConfigDescription.overview = getMessage('GenericEngineConfigOverview', engineName.toUpperCase());
+    }
+
+    // Every engine config should have a disable_engine field which we prefer to be first in the object for display purposes
+    engineConfigDescription.fieldDescriptions = {
+        [FIELDS.DISABLE_ENGINE]: getMessage('EngineConfigFieldDescription_disable_engine', engineName),
+        ... engineConfigDescription.fieldDescriptions
```
jfeingold35, line 536: I'm not sure how I feel about this method of putting `disable_engine` at the front.
Question: Is this something we want to enforce at the Core level or is it a CLI constraint only? If the latter, then perhaps I could just use `yaml.dump()`'s options parameter and supply `sortKeys` with a function that treats `disable_engine` as less than every other key?

stephen-carter-at-sf: This is just a different implementation of what was already here... allowing it to be at the top instead of the bottom. But there is no test enforcing this on the core side. It is just a field on the object afterall. On the CLI side, you'll need a test case to lock in order if we care about order over there... so this was just to help reduce your source code implementation if it was already done at the core level.

jfeingold35 (approved):

PR #86, opened 2024-09-06
NEW(pmd): @W-16113607@: Add in progress and additional log events to PMD

stephen-carter-at-sf:


packages/code-analyzer-pmd-engine/pmd-wrapper/src/main/java/com/salesforce/sfca/pmdwrapper/PmdRuleRunner.java
```diff
@@ -14,7 +18,35 @@ void runRules(String ruleSetInputFile, String filesToScanInputFile, String resul
...
+            }
+        };
+    }
+
+    @Override
+    public synchronized FileAnalysisListener startFileAnalysis(TextFile textFile) {
+        // Note that this method must be synchronized so that multiple threads cannot mess
+        // up the order of progress with race conditions.
+        fileCount++;
+        System.out.println("[Progress]" + fileCount + "::" + totalNumFiles);
```
jfeingold35, line 46: Maybe I'm misremembering, but I believe we had a goal of not using stdout to communicate between Java and TS. Where there any alternative options? And if so, why were they inferior to this one?

jfeingold35 (approved):

packages/code-analyzer-pmd-engine/pmd-wrapper/src/main/java/com/salesforce/sfca/pmdwrapper/PmdRuleRunner.java
stephen-carter-at-sf: As mentioned over on https://docs.google.com/document/d/1i4lPeIyeybqdaVUcWgvCpmRi0w2fTIide4BuyQmIq6A/edit 

WHERE WE LANDED: After further investigation and discussion we agreed that we indeed would not communicate final result information via StdOut…. And use files instead. And for progress and log events, we agreed that although web sockets are more efficient, they are less performant and may have unintended consequences on the users machine. So for now, we will attempt to make StdOut as reliable as possible and use it for now since these types of progress and log events don’t need to be perfectly reliable.

jfeingold35: Fair enough.

PR #92, opened 2024-09-13
NEW @W-16606100@ FlowTest engine help text callable.

jfeingold35:


packages/code-analyzer-flowtest-engine/pipx.pyz
jfeingold35, line 1: This file came from PipX's repository, and is a [file that allows us to run PipX without actually requring the user to install it](https://github.com/pypa/pipx?tab=readme-ov-file#using-pipx-without-installing-via-zipapp).

packages/code-analyzer-flowtest-engine/test/python/FlowTestCommandWrapper.test.ts
```diff
@@ -0,0 +1,20 @@
...
+import {RunTimeFlowTestCommandWrapper} from "../../src/python/FlowTestCommandWrapper";
+
+const PATH_TO_PYTHON_EXE: string = sync('python3');
+
+describe('FlowTestCommandWrapper implementations', () => {
+    describe('RunTimeFlowTestCommandWrapper', () => {
+        // TODO: This is a temporary test serving as a proof-of-concept for bundling and invoking the FlowTest engine.
+        //       Once the FlowTest engine's API has stabilized, this test should be removed and replaced with ones that
+        //       interface with that API.
+        it('Help text is properly retrieved', async () => {
```
jfeingold35, line 11: This test is acting as a proof-of-concept, verifying that we can successfully call FlowTest with the `-h` flag and get usage information back.

packages/code-analyzer-flowtest-engine/test/python/FlowTestCommandWrapper.test.ts
```diff
@@ -0,0 +1,20 @@
...
+            const wrapper: RunTimeFlowTestCommandWrapper = new RunTimeFlowTestCommandWrapper(PATH_TO_PYTHON_EXE);
+
+            const helpText: string = await wrapper.getFlowTestHelpText();
+
+            expect(helpText).toContain('usage: flowtest');
+        // For the sake of CI/CD, set the timeout to a truly absurd value.
+        }, 30000);
```
jfeingold35, line 18: On my machine, the default 5-second timeout is often not enough, which means that it's _definitely_ not enough on the Windows GHA machine. So I've upped it to 30 seconds, which should definitely be enough.

packages/code-analyzer-flowtest-engine/src/python/FlowTestCommandWrapper.ts
```diff
@@ -0,0 +1,38 @@
...
+const PATH_TO_FLOWTEST_ROOT = path.join(__dirname, '..', '..', 'FlowTest');
+
+export class RunTimeFlowTestCommandWrapper implements FlowTestCommandWrapper {
+    private readonly pythonCommandExecutor: PythonCommandExecutor;
+
+    public constructor(pythonCommand: string) {
+        this.pythonCommandExecutor = new PythonCommandExecutor(pythonCommand);
+    }
+
+    public async getFlowTestHelpText(): Promise<string> {
```
jfeingold35, line 19: Eventually, this method will be replaced with `describeRules`() and `runRules()`. But for now, we're just trying to make sure that we can actually run FlowTest at all, and the `-h` flag is a good way to test that.

packages/code-analyzer-flowtest-engine/src/python/PythonCommandExecutor.ts
```diff
@@ -0,0 +1,46 @@
...
+const NO_OP = () => {};
+
+export class PythonCommandExecutor {
+    private readonly pythonCommand: string;
+
+    public constructor(pythonCommand: string) {
+        this.pythonCommand = pythonCommand;
+    }
+
+    public async exec(pythonCmdArgs: string[], processStdout: ProcessStdOutFn = NO_OP): Promise<void> {
```
jfeingold35, line 14: Most of this method is just adapted from the corresponding PMD engine class, `JavaCommandExecutor`.

packages/code-analyzer-flowtest-engine/package.json
```diff
@@ -36,6 +36,8 @@
   "files": [
     "dist",
     "LICENSE",
+    "pipx.pyz",
+    "FlowTest",
```
jfeingold35, line 40: Adding the new files to the package output.

packages/code-analyzer-flowtest-engine/FlowTest/README.md
```diff
@@ -0,0 +1,105 @@
...
+
+The combination of all of these factors means that there is a great need to rapidly scan flows for security issues,
+as well as to provide feedback to partners so that they can be warned when an input variable can be a selector for
+something like record deletes, or when user data can modify existing records in an org, even if they don't belong
+to the user.
+
+At a minimum, this is true in the Appexchange review, but tools like this should be made available to developers and
+partners as well.
+
+## Installation
```
jag-j, line 31: Is this accurate? This seems to be for standalone flowtest, not when it is bundled as part of code analyzer.

jag-j: Never mind. I just saw your message that I could skip everything under FlowTest directory.

jfeingold35: Actually, removing the READMEs is probably a decent idea. I'll do that once you're done with the rest of the review.

jfeingold35: (Although, it might not matter in the long-term. As Robert makes more changes, we'll probably have to do this copy-paste routine a few more times.)

packages/code-analyzer-flowtest-engine/pipx.pyz
jag-j: Just for my understanding - this means we don't require internet to download pip, but we still build the python module for the underlying architecture and OS right?

jag-j (approved):

packages/code-analyzer-flowtest-engine/pipx.pyz
jfeingold35: The user still needs to be connected to the internet so that PipX can download and build the dependencies, but they don't need to manually install PipX itself because we're already including it. As long as they have Python and an Internet connection, they're good to go.

PR #93, opened 2024-09-25
NEW @W-16360876@ Implemented describeRules() functionality for FlowTest

jfeingold35:
This PR does the following:
- Deletes a bunch of files from the locally stored FlowTest folder that never should have been committed in the first place.
- Updates several files within the locally stored FlowTest folder to provide the new API for rule description.
- Utilizes that API to get descriptions for the FlowTest queries, and converts those into the CodeAnalyzer house style.


packages/code-analyzer-flowtest-engine/.gitignore
```diff
@@ -0,0 +1,3 @@
+FlowTest/build/**/**
```
jfeingold35, line 1: I added a .gitignore to FlowTest to prevent files that don't belong from getting committed again.

packages/code-analyzer-flowtest-engine/src/engine.ts
```diff
@@ -2,19 +2,20 @@ import {
...
-import {FlowTestConfig} from "./config";
+import {FlowTestCommandWrapper, FlowTestRuleDescriptor} from "./python/FlowTestCommandWrapper";
 
 export class FlowTestEngine extends Engine {
     public static readonly NAME: string = 'flowtest';
-    private readonly config: FlowTestConfig;
+    private readonly commandWrapper: FlowTestCommandWrapper;
 
-    public constructor(config: FlowTestConfig) {
+    public constructor(commandWrapper: FlowTestCommandWrapper) {
```
jfeingold35, line 17: Passing the wrapper into the constructor allows us to replace it with a stub during tests.

packages/code-analyzer-flowtest-engine/src/engine.ts
```diff
@@ -23,24 +24,60 @@ export class FlowTestEngine extends Engine {
...
         return {
             violations: []
         };
     }
+
+    private convertFlowTestRulesToCodeAnalyzerRules(flowTestRules: FlowTestRuleDescriptor[]): RuleDescription[] {
+        return flowTestRules.map(flowTestRule => {
+            return {
+                // The name maps directly over.
+                name: flowTestRule.query_name,
```
jfeingold35: The `query_name` field has spaces in it, and the `query_id` field has periods in it. So I chose to use the one with spaces in it for now. We can have a discussion about what postprocessing might be needed to bring the names more into line with our best practices.

packages/code-analyzer-flowtest-engine/src/python/FlowTestCommandWrapper.ts
```diff
@@ -25,14 +35,14 @@ export class RunTimeFlowTestCommandWrapper implements FlowTestCommandWrapper {
...
+            '-p'
         ];
 
         let stdout: string = '';
         const processStdout = (stdoutMsg: string) => {
             stdout += stdoutMsg;
         }
         await this.pythonCommandExecutor.exec(pythonArgs, processStdout);
-        return stdout;
+        return (JSON.parse(stdout) as FlowTestRuleDescriptor[]).sort((a, b) => a.query_name.localeCompare(b.query_name));
```
jfeingold35: The sorting is necessary because the output printed to stdout isn't sorted at this time. Robert intends to update it with sorting, and when he does I will change it here too.

packages/code-analyzer-flowtest-engine/src/python/FlowTestCommandWrapper.ts
```diff
@@ -3,7 +3,17 @@ import {PythonCommandExecutor} from './PythonCommandExecutor';
 
 
 export interface FlowTestCommandWrapper {
-    getFlowTestHelpText(): Promise<string>;
+    getFlowTestRuleDescriptions(): Promise<FlowTestRuleDescriptor[]>;
+}
+
+export type FlowTestRuleDescriptor = {
```
jag-j, line 13: Nit: The rule for each of the engines is defined in different places. It will be good if they all follow the same convention. I personally like how regex engine has this in config.ts.

jag-j: I think it will also help to add some description on the fields.

jfeingold35: Reasonable. I'll get on that.

jfeingold35: Actually, while I agree that the fields definitely need descriptions, I think it makes more sense for the definition to stay in the Wrapper file, since it's more closely related to interaction with the Python layer than anything else. That's how the PMD integration does it as well, so there's precedent.

jag-j: Sounds good. For something like regexengine that doesn't need a wrapper, it is in config.

jag-j (approved):

packages/code-analyzer-flowtest-engine/test/test-data/goldfiles/FlowTestCommandWrapper.test.ts/catalog.json
stephen-carter-at-sf, line 1: If you name this file catalog.goldfile.json then it'll automatically get the benefit of having linebreaks treated as linux always via: https://github.com/forcedotcom/code-analyzer-core/blob/dev/.gitattributes

packages/code-analyzer-flowtest-engine/test/engine.test.ts
stephen-carter-at-sf, line 1: I see that code coverage for this flowtest package is no longer 100% like all other packages
```
 code-analyzer-flowtest-engine/src        |     100 |      100 |     100 |     100 |                   
  config.ts                               |     100 |      100 |     100 |     100 |                   
  engine.ts                               |     100 |      100 |     100 |     100 |                   
  messages.ts                             |     100 |      100 |     100 |     100 |                   
  plugin.ts                               |     100 |      100 |     100 |     100 |                   
 code-analyzer-flowtest-engine/src/python |   90.56 |    44.44 |   77.77 |   92.15 |                   
  FlowTestCommandWrapper.ts               |     100 |      100 |     100 |     100 |                   
  PythonCommandExecutor.ts                |   78.26 |    28.57 |      70 |   80.95 | 35-45             
  PythonVersionIdentifier.ts              |     100 |      100 |      80 |     100 |                     
```
If possible, can we either add in code coverage for anything to be submitted or add in the ignore comments where code coverage isn't easily able to be covered?
  
 Or will these be hit when the run method is implemented?

packages/code-analyzer-flowtest-engine/src/engine.ts
```diff
@@ -23,24 +24,60 @@ export class FlowTestEngine extends Engine {
 
     public async describeRules(_describeOptions: DescribeOptions): Promise<RuleDescription[]> {
```
stephen-carter-at-sf: The describe command should be looking at the workspace if it is provided to see if the workspace even contains relevant files (in this case, flow based xml files). If there is a workspace and there are no relevant files, then no rules should be returned. That is, we shouldn't be doing any work if there are no flow files in the workspace.

packages/code-analyzer-flowtest-engine/src/engine.ts
```diff
@@ -23,24 +24,60 @@ export class FlowTestEngine extends Engine {
...
 
         this.emitRunRulesProgressEvent(10);
-        this.emitLogEvent(LogLevel.Info, `Temporary message: Python command identified as ${pythonCommand}`);
         this.emitRunRulesProgressEvent(100);
         return {
             violations: []
         };
     }
+
+    private convertFlowTestRulesToCodeAnalyzerRules(flowTestRules: FlowTestRuleDescriptor[]): RuleDescription[] {
```
stephen-carter-at-sf: Nitpick: Can these private methods instead be functions? 

Also maybe call this function `toRuleDescription(flowTestRule: FlowTestRuleDiscriptor)` and then simply make your line 29 to be 
```
const ruleDescriptions: RuleDescription[] = flowTestRules.map(toRuleDescription);
```

packages/code-analyzer-flowtest-engine/src/python/FlowTestCommandWrapper.ts
```diff
@@ -25,14 +59,14 @@ export class RunTimeFlowTestCommandWrapper implements FlowTestCommandWrapper {
...
+            '-p'
         ];
 
         let stdout: string = '';
         const processStdout = (stdoutMsg: string) => {
             stdout += stdoutMsg;
         }
         await this.pythonCommandExecutor.exec(pythonArgs, processStdout);
-        return stdout;
+        return JSON.parse(stdout) as FlowTestRuleDescriptor[];
```
stephen-carter-at-sf: We shouldn't be processing stdout if at all possible because other things can end up in the stdout that we don't want... or stdout can clip if the output is long and the environment caps the buffer. Does flowtest allow us to write the output to a file? If so, we should be reading that file and parsing its output - all in a try catch so that if reading or parsing fails, we know what step it failed at.

If there isn't a way to do it with output files... then at the very minimum - let's put this JSON.parse in a try/catch with a nice error message where we can somehow capture the output (maybe to the logs or somehwere else) so we can diagnose issues in the future.

stephen-carter-at-sf (changes requested):

packages/code-analyzer-flowtest-engine/test/test-data/goldfiles/FlowTestCommandWrapper.test.ts/catalog.json
jfeingold35: Good catch.

packages/code-analyzer-flowtest-engine/src/engine.ts
jfeingold35: It looks like the extensions the FlowTest tool looks for are `.flow-meta.xml` and `.flow`. So I'll make the plugin look for those too.

packages/code-analyzer-flowtest-engine/test/engine.test.ts
stephen-carter-at-sf: Improved: Now I see
```
------------------------------------------|---------|----------|---------|---------|-------------------
File                                      | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s 
------------------------------------------|---------|----------|---------|---------|-------------------
  PythonVersionIdentifier.ts              |     100 |      100 |      80 |     100 |           
```
Maybe there is 1 function that isn't fully hit.

packages/code-analyzer-flowtest-engine/src/engine.ts
```diff
@@ -23,24 +24,56 @@ export class FlowTestEngine extends Engine {
...
     public async describeRules(_describeOptions: DescribeOptions): Promise<RuleDescription[]> {
         this.emitDescribeRulesProgressEvent(0);
-        const pythonCommand: string = this.config.python_command;
-        this.emitDescribeRulesProgressEvent(10);
-        this.emitLogEvent(LogLevel.Info, `Temporary message: Python command identified as ${pythonCommand}`);
-        this.emitRunRulesProgressEvent(100);
-        return [];
+        const flowTestRules: FlowTestRuleDescriptor[] = await this.commandWrapper.getFlowTestRuleDescriptions();
+        this.emitDescribeRulesProgressEvent(75);
+        const convertedRules = flowTestRules.map(r => toRuleDescription(r));
```
stephen-carter-at-sf: Since it is now a function (and not tied to a class instance), I believe you could have also just done `flowTestRules.map(toRuleDescription)`

packages/code-analyzer-flowtest-engine/src/engine.ts
```diff
@@ -23,24 +24,56 @@ export class FlowTestEngine extends Engine {
...
         return {
             violations: []
         };
     }
 }
 
+function toRuleDescription(flowTestRule: FlowTestRuleDescriptor): RuleDescription {
+    return {
+        // The name maps directly over.
+        name: flowTestRule.query_name,
```
stephen-carter-at-sf: It looks like in the catalog.goldfile.json that an example query_name is `Flow: SystemModeWithSharing recordUpdates data` and here it looks like we are using this as the rule name. This does not seem good to me.

Robert closed the Enhancement Request for rules having shorter names (see https://git.soma.salesforce.com/SecurityTools/FlowSecurityLinter/issues/39) ... but it looks like he maintained spaces and colons in the names. For example I see

```query_name = f"Flow Security: User access to {elem_type} {check_labels_val} in mode {run_mode}"```
got turned into
```
query_name = f"Flow: {run_mode} {elem_type} {check_labels_val}"
```
which is still too long.

So either we need to go back to him and make these names even shorter without spaces and colons, or we'll need to create a mapping on our end (either fixed or dynamic).

For example on our end we could probably just remove the prefix "Flow: " and then convert all spaces to dashes for now or something. So instead of
`Flow: SystemModeWithSharing recordUpdates data` maybe the name could for now just be `SystemModeWithSharing-recordUpdates-data`

But I created https://git.soma.salesforce.com/SecurityTools/FlowSecurityLinter/issues/47 to ask that we may receive better names.

**Review by stephen-carter-at-sf — CHANGES_REQUESTED** [review]

File: packages/code-analyzer-flowtest-engine/src/engine.ts [hunk 11]
```diff
@@ -4,43 +4,91 @@ import {
...
-        const pythonCommand: string = this.config.python_command;
-        this.emitDescribeRulesProgressEvent(10);
-        this.emitLogEvent(LogLevel.Info, `Temporary message: Python command identified as ${pythonCommand}`);
-        this.emitRunRulesProgressEvent(100);
-        return [];
+        if (describeOptions.workspace) {
+            const workspaceFiles: string[] = await describeOptions.workspace.getExpandedFiles();
+            // If a workspace is provided but it contains no flow files, then return no rules.
+            if (!workspaceFiles.some(fileIsFlowFile)) {
+                this.emitLogEvent(LogLevel.Fine, 'Workspace contains no Flow files; returning no rules');
```
stephen-carter-at-sf: This seems relevant to possible user debugging... so maybe make this to LogLevel.Debug instead of LogLevel.Fine.

stephen-carter-at-sf (approved):

PR #96, opened 2024-10-01
NEW (RegexEngine) @W-15490095@ Rule to reduce calls to Limits.getHeapSize()

jag-j:


packages/code-analyzer-regex-engine/src/messages.ts
```diff
@@ -53,6 +53,12 @@ const MESSAGE_CATALOG : { [key: string]: string } = {
 
     AvoidOldSalesforceApiVersionsRuleMessage:
         `Found the use of a Salesforce API version that is 3 or more years old. Avoid using an API version that is <= %d.0.`,
+    
+    AvoidGetHeapSizeInLoopRuleDescription:
```
jag-j, line 57: @jshackell-sfdc: Can you please review these two messages? Thank you!

packages/code-analyzer-regex-engine/src/messages.ts
```diff
@@ -53,6 +53,12 @@ const MESSAGE_CATALOG : { [key: string]: string } = {
...
+        `Detects usage of Limits.getHeapSize() in loops`,
```
jshackell-sfdc, line 58:
```suggestion
        `Detects usage of Limits.getHeapSize() in loops. We recommend you avoid this pattern due to performance and resource reasons.`,
```

packages/code-analyzer-regex-engine/src/messages.ts
```diff
@@ -53,6 +53,12 @@ const MESSAGE_CATALOG : { [key: string]: string } = {
...
+    
+    AvoidGetHeapSizeInLoopRuleMessage:
+        `Found the use of Limits.getHeapSize() in a loop`,
```
jshackell-sfdc:
```suggestion
        `Found the use of Limits.getHeapSize() in a loop. We recommend you avoid this pattern due to performance and resource reasons.`,
```

jshackell-sfdc (approved):
See my edits!

packages/code-analyzer-regex-engine/src/messages.ts
stephen-carter-at-sf: The rule description should be somewhat short... but the violation message can be longer. So I think the "We recommend" stuff shouldn't be here in the description but should indeed be in the `AvoidGetHeapSizeInLoopRuleMessage`.

Note that there is no "this" at the rule description level since there is no violation, which is another reason I think I would not add in the "We recommend ..." stuff here.

packages/code-analyzer-regex-engine/src/plugin.ts
```diff
@@ -77,7 +77,14 @@ export function createBaseRegexRules(now: Date): RegexRules {
...
             tags: ['Recommended']
         },
-        AvoidOldSalesforceApiVersions: createAvoidOldSalesforceApiVersionsRule(now)
+        AvoidOldSalesforceApiVersions: createAvoidOldSalesforceApiVersionsRule(now),
+        AvoidGetHeapSizeInLoop: {
+            regex: (/(for|while)\s*\([^)]*\)\s*\{[^}]*\bLimits\.getHeapSize\(\);|do\s*\{[^}]*\}\s*while\s*\([^)]*\);/gi).toString(),
+            description: getMessage('AvoidGetHeapSizeInLoopRuleDescription'),
+            violation_message: getMessage('AvoidGetHeapSizeInLoopRuleMessage'),
+            severity: SeverityLevel.High,
+            tags: ['Recommended', 'Performance']
```
stephen-carter-at-sf, line 86: This is fine... but eventually we'll need to reevaluate all the tags for all of our rules accross all engines to determine (besides Recommended) what the standard tag categories should be.

packages/code-analyzer-regex-engine/src/messages.ts
jshackell-sfdc: Okay!  Whatever you think best.

packages/code-analyzer-regex-engine/src/plugin.ts
```diff
@@ -77,7 +77,14 @@ export function createBaseRegexRules(now: Date): RegexRules {
             severity: SeverityLevel.Info,
             tags: ['Recommended']
         },
-        AvoidOldSalesforceApiVersions: createAvoidOldSalesforceApiVersionsRule(now)
+        AvoidOldSalesforceApiVersions: createAvoidOldSalesforceApiVersionsRule(now),
+        AvoidGetHeapSizeInLoop: {
+            regex: (/(for|while)\s*\([^)]*\)\s*\{[^}]*\bLimits\.getHeapSize\(\);|do\s*\{[^}]*\}\s*while\s*\([^)]*\);/gi).toString(),
```
stephen-carter-at-sf: Concerns:
* First, we aren't checking for whether the code lives in comments. Probably can't do that anyway - just want to make sure this limitation is known
* Second, for the first for|while case... if the user puts an if statement inside of their loop prior to Limits.getHeapSize then this regular expression will assume we already exited the loop when it finds the first closed curly bracket. Regular expressions have been proven to not know the level of depth a specific bracket is for... so there are many edge cases where we might not alert when we should.
* Third, there might be scenarios where we actually have a Limits.getHeapSize inside of a loop that aren't really being invoked by the loop... maybe an anonymous function or class is being defined in the loop and has part of its definition the Limits.getHeapSize() call... not sure what user can or can't do with apex - just calling out that some edge cases there might exist.
* Fourth, the do loop doesn't seem to even be checking for the getHeapSize function at all... so wouldn't this just pick up like every do loop? And if you do add in the Limits\.getHeapSize... do we need to look after the following close bracket while? That is we probably don't care about the trailing `\s*\([^)]*\)`

In general, I think we might only be able to get this working 95% of the time at best. It really should be added to PMD engine instead in my opinion.

stephen-carter-at-sf (changes requested):

packages/code-analyzer-regex-engine/src/plugin.ts
jag-j: As discussed offline, checking if the code is commented is a known limitation. I have taken care of the Fourth issue. Regarding second and third, I agree with the limitations. But PMD also has limitations if the Limit.getHeapSize() is within a function that is called from a loop. Graph Engine would be the most accurate, but it's also the most expensive, both in terms of implementation and time for users to get results. The Regex engine is the least expensive, which is why I slightly preferred it over PMD. It also gave us an opportunity to utilize the new engine we built from the ground up.

stephen-carter-at-sf (approved):

packages/code-analyzer-regex-engine/src/plugin.ts
stephen-carter-at-sf: I approved the PR but I do believe having a if statement before Limits.getHeapSize() is still something PMD would handle and not care about.... but what kills this particular regex based rule.

jag-j: Agreed! I'll bring it up with the Apex team and get in their inputs as well.

PR #99, opened 2024-10-02
NEW(pmd): @W-16113765@: Find and validate java command from environment or config

stephen-carter-at-sf:


packages/code-analyzer-eslint-engine/src/messages.ts
```diff
@@ -38,10 +38,10 @@ const MESSAGE_CATALOG : { [key: string]: string } = {
         `The ESLintEnginePlugin does not support an engine with name '%s'.`,
 
     InvalidLegacyConfigFileName:
-        `The '%s' configuration value was invalid. Expected the file name '%s' to be one of the following: %s`,
+        `The '%s' configuration value is invalid. Expected the file name '%s' to be one of the following: %s`,
```
stephen-carter-at-sf, line 41: Fixed a number of these so that they are consistent with all the other messages which use "is invalid" instead of "was invalid".

packages/code-analyzer-pmd-engine/src/pmd-engine.ts
```diff
@@ -92,7 +92,8 @@ export class PmdEngine extends Engine {
         const cacheKey: string = getCacheKey(workspaceLiaison.getWorkspace());
         if (!this.pmdRuleInfoListCache.has(cacheKey)) {
             const relevantLanguages: PmdLanguage[] = await workspaceLiaison.getRelevantLanguages();
-            const ruleInfoList: PmdRuleInfo[] = await this.pmdWrapperInvoker.invokeDescribeCommand(relevantLanguages, emitProgress);
+            const ruleInfoList: PmdRuleInfo[] = relevantLanguages.length === 0 ? [] :
+                await this.pmdWrapperInvoker.invokeDescribeCommand(relevantLanguages, emitProgress);
```
stephen-carter-at-sf, line 96: This fixes a small bug that existed where the rules command when provided a workspace containing no relevant files still tried to call through to the PMD wrapper. Now we don't attempt to do so unless we should.

packages/code-analyzer-pmd-engine/src/JavaVersionIdentifier.ts
```diff
@@ -0,0 +1,49 @@
...
+                reject(err.message);
+            });
+        });
+    }
+}
+
+export function _extractJavaVersionFrom(javaVersionOutput: string): SemVer|null {
+    // First we'll see if the word "version" exists with the version number and use that first.
+    const matchedParts: RegExpMatchArray | null = javaVersionOutput.match(/version\s+"?(\d+(\.\d+)*)"?/i);
+    if (matchedParts && matchedParts.length > 1) {
```
jag-j: Nit: Checking for matchedParts means there was a match, so no need to check for the length here.

packages/code-analyzer-pmd-engine/src/config.ts
```diff
@@ -0,0 +1,94 @@
...
+        throw new Error(getMessage('InvalidUserSpecifiedJavaCommand',
+            configValueExtractor.getFieldPath('java_command'), (err as Error).message));
+    }
+    return javaCommand;
+}
+
+async function attemptToAutoDetectJavaCommand(configValueExtractor: ConfigValueExtractor, javaVersionIdentifier: JavaVersionIdentifier): Promise<string> {
+    const commandsToAttempt: string[] = [
+        // Environment variables specifying JAVA HOME take precedence (if they exist)
+        ... ['JAVA_HOME', 'JRE_HOME', 'JDK_HOME'].filter(v => process.env[v] !== undefined)
```
jag-j: Just to be sure, can we check for empty environmental variable as well in additional to undefined?

jag-j: Actually, help me understand why are we looking at each of these paths separately instead of just running "java"? Wouldn't that automatically pick the java as set in the environmental variable?

packages/code-analyzer-pmd-engine/src/JavaVersionIdentifier.ts
stephen-carter-at-sf: We are checking that the length > 1 not > 0. This code was taken from our code analyzer 4.0 world... but yeah - let me check the documentation of match to see if checking the length is necessary here.

packages/code-analyzer-pmd-engine/src/config.ts
stephen-carter-at-sf: Yeah I should probably also check for empty strings as well... good catch. In fact I think I can get away with just having .filter(v => process.env[v]) now that I look into it.

But yeah I'm trying to maintain a bit of the same behavior we had with code analyzer 5 where we chose to use JAVA_HOME before looking on the path for java. This actually is in line with our github action as well which sets JAVA_HOME to the version of java that we wish to have (based on the github runner being used) instead of just relying on the default java on the path. I did some research in the area and many tools seem to have this same strategy of favoring the environment variable over the path... so I don't think we are off in doing this.

stephen-carter-at-sf: Done.

packages/code-analyzer-pmd-engine/src/JavaVersionIdentifier.ts
stephen-carter-at-sf: Fixed. Done.

jag-j (approved):

PR #100, opened 2024-10-03
NEW(pmd): @W-16605956@: Add ability to specify rule_languages from pmd engine config

stephen-carter-at-sf:


packages/code-analyzer-pmd-engine/src/config.ts
```diff
@@ -1,94 +1,146 @@
...
+    const pmdConfigValueExtractor: PmdConfigValueExtractor = new PmdConfigValueExtractor(configValueExtractor,
+        javaVersionIdentifier);
     return {
-        java_command: await extractJavaCommand(configValueExtractor, javaVersionIdentifier)
+        java_command: await pmdConfigValueExtractor.extractJavaCommand(),
+        rule_languages: pmdConfigValueExtractor.extractRuleLanguages()
     }
 }
 
-async function extractJavaCommand(configValueExtractor: ConfigValueExtractor, javaVersionIdentifier: JavaVersionIdentifier): Promise<string> {
```
stephen-carter-at-sf, line 36: Sorry for all the noise in this file. The changes were actually pretty minimal... I just moved the functions into a class and then add in the extractRuleLanguages method. The extractJavaCommand and other methods really were untouched.

packages/code-analyzer-pmd-engine/src/pmd-engine.ts
```diff
@@ -32,7 +32,7 @@ export class PmdEngine extends Engine {
         const javaCommandExecutor: JavaCommandExecutor = new JavaCommandExecutor(config.java_command);
         this.pmdWrapperInvoker = new PmdWrapperInvoker(javaCommandExecutor,
             (logLevel: LogLevel, message: string) => this.emitLogEvent(logLevel, message));
-        this.availableLanguages = [PmdLanguage.APEX, PmdLanguage.VISUALFORCE]; // TODO: Will be configurable soon
+        this.selectedLanguages = config.rule_languages as PmdLanguage[];
```
stephen-carter-at-sf, line 35: This basically was the main change beside the switch to use description instead of message from PMD. Again sorry that the review became noisy with a few areas of refactoring.

packages/code-analyzer-flowtest-engine/src/python/PythonCommandExecutor.ts
```diff
@@ -19,13 +19,16 @@ export class PythonCommandExecutor {
...
                     msg.split("\n").map(line => processStdout(line));
                 }
             });
 
             pythonProcess.stderr.on('data', (data: Buffer) => {
-                stderrMessages.push(`${data.toString().trim()}`);
+                const msg: string = data.toString().trim();
+                if(msg.length > 0) { // Not sure why stderr spits out empty lines sometimes, but we ignore them nonetheless
+                    stderrMessages.push(msg);
+                }
```
stephen-carter-at-sf, line 31: Without this, a test that Josh wrote was being flaky. So I added this in to make the output of stderr more robust.

packages/code-analyzer-pmd-engine/src/config.ts
```diff
@@ -1,94 +1,146 @@
...
+
+        // Provide support for 'javascript' which is a supported alias of 'ecmascript'
+        if (ruleLanguagesSet.has('javascript')) {
+            ruleLanguagesSet.delete('javascript');
+            ruleLanguagesSet.add('ecmascript');
+        }
+
+        // Validate each language is supported
+        ruleLanguages = [...ruleLanguagesSet];
+        for (const ruleLanguage of ruleLanguages) {
```
jag-j, line 91: Nit: why convert to array to iterate? Wouldn't set work the same way?

packages/code-analyzer-pmd-engine/pmd-wrapper/src/main/java/com/salesforce/sfca/pmdwrapper/PmdRuleDescriber.java
```diff
@@ -91,4 +91,27 @@ List<PmdRuleInfo> describeRulesFor(Set<String> languages) {
         }
         return ruleInfoList;
     }
+
+    private static String getLimitedDescription(Rule rule) {
```
jag-j, line 95: The logic here is complex enough I would prefer to add a test to this function specifically. It would serve as a good documentation. If that isn't possible (given you test at the command level mostly), can you consolidate the comments here as a doc at the method level?

packages/code-analyzer-pmd-engine/src/config.ts
stephen-carter-at-sf: Converting to an array isn't for iterating purposes... notice the return statement after the loop. I am basically preparing to return the array - I just so happen to iterate over it to do validation checking before I return the normalized version of it.

packages/code-analyzer-pmd-engine/pmd-wrapper/src/main/java/com/salesforce/sfca/pmdwrapper/PmdRuleDescriber.java
stephen-carter-at-sf: This is implementation detail. The resulting output is indeed locked in with the `whenDescribeRulesForJava_thenCorrectRulesAreTurned` test inside of PmdRuleDesriberTest.java. Notice line 82.

Also all the gold files on the typescript side are also locking in the output to confirm it is correct.

I prefer not to create unit tests for implementation detail. I instead prefer to keep my tests at the boundary of the interface I am exposing... in this case it is the on the describeRulesFor method.

But if you really feel strongly I suppose I could add specific implementation detail specific unit tests for this. Let me know and I can do it on Monday if you feel strongly.

jag-j: That's fine. Let's leave it that this way.

jag-j (approved):

PR #101, opened 2024-10-06
NEW (Extension) @W-16833947@ Add rule to require a minimum API version of abstract/virtual classes with private methods

jag-j:


packages/code-analyzer-regex-engine/src/engine.ts
```diff
@@ -108,7 +108,10 @@ export class RegexEngine extends Engine {
 
     private async scanFile(fileName: string, ruleName: string): Promise<Violation[]> {
         const violations: Violation[] = [];
-        const fileContents: string = await fs.promises.readFile(fileName, {encoding: 'utf8'});
+        let fileContents: string = await fs.promises.readFile(fileName, {encoding: 'utf8'})
+        if (this.regexRules[ruleName].include_metadata) {
+            fileContents = fileContents + await fs.promises.readFile(fileName + "-meta.xml", {encoding: 'utf8'}) ;
```
stephen-carter-at-sf: A few things here.
First, this seems to assume that the <fileName>-meta.xml file exists and are attempting to read it without first checking.
Second, if you do check if it exists, you must stay within the workspace provided... so you'll need to see if the file exists among the files provided by the workspace only (and not checking on disk).

packages/code-analyzer-regex-engine/src/messages.ts
```diff
@@ -60,6 +60,12 @@ const MESSAGE_CATALOG : { [key: string]: string } = {
     AvoidGetHeapSizeInLoopRuleMessage:
         `Found the use of Limits.getHeapSize() in a loop. We recommend you avoid this pattern due to performance and resource reasons.`,
 
+    MinVersionForAbstractVirtualClassesWithPrivateMethodRuleDescription:
+        `Detects private methods within abstract/virtual classes`,
```
stephen-carter-at-sf: This should say something about the version.

packages/code-analyzer-regex-engine/src/messages.ts
```diff
@@ -60,6 +60,12 @@ const MESSAGE_CATALOG : { [key: string]: string } = {
...
+    
+    MinVersionForAbstractVirtualClassesWithPrivateMethodRuleMessage:
+        `Found private methods within abstract/virtual classes. Make sure the corresponding API version of the class is at least v61.0. If so, you can ignore this rule.`,
```
stephen-carter-at-sf: Shouldn't this be updated as well... why would someone need to ignore it if your version check ideally handles the correct range?

packages/code-analyzer-regex-engine/src/config.ts
```diff
@@ -45,6 +45,10 @@ export type RegexRule = {
     // Possible values: 1 or 'Critical', 2 or 'High', 3 or 'Moderate', 4 or 'Low', 5 or 'Info'
     // Default: 'Moderate'
     severity: SeverityLevel;
+
+    // [Optional] If the corresponding -meta.xml file should be included for analysis
+    // Default: false
+    include_metadata?: boolean;
```
stephen-carter-at-sf: Please add a comment saying this "INTERNAL USE ONLY (NOT EXPOSED TO USER CONFIG)"

packages/code-analyzer-regex-engine/src/engine.ts
jag-j: Sure, I should check for the existence of -meta.xml file. But that ideally should never be the case in a real world project and in case the meta file isn't found, I want to silently proceed instead of throwing an error.

To your second point, let's assume someone says "--workspace ./force-app/**/*.cls" to scan their apex files, we should still be checking for the corresponding -meta.xml files from the disk rather than just limiting to the workspace. The user doesn't case this rule takes into account -meta.xml files and we shouldn't be expecting them to explicitly define that in their command.

packages/code-analyzer-regex-engine/src/engine.ts
```diff
@@ -108,7 +108,16 @@ export class RegexEngine extends Engine {
 
     private async scanFile(fileName: string, ruleName: string): Promise<Violation[]> {
         const violations: Violation[] = [];
-        const fileContents: string = await fs.promises.readFile(fileName, {encoding: 'utf8'});
+        let fileContents: string = await fs.promises.readFile(fileName, {encoding: 'utf8'})
+        if (this.regexRules[ruleName]?.include_metadata) {
+            const medataFileName = fileName + "-meta.xml";
+            try {
+                await fs.promises.access(medataFileName, fs.constants.F_OK);
+                fileContents = fileContents + await fs.promises.readFile(medataFileName, {encoding: 'utf8'}) ;
```
stephen-carter-at-sf, line 116: The contract of the engine is to not escape the boundaries of the workspace. So you should be checking to see if the file exists in the Workspace.getExpandedFiles() array instead of looking on disk (which the getExpandedFiles does automatically already).

stephen-carter-at-sf (changes requested):

packages/code-analyzer-regex-engine/src/config.ts
```diff
@@ -45,6 +45,11 @@ export type RegexRule = {
     // Possible values: 1 or 'Critical', 2 or 'High', 3 or 'Moderate', 4 or 'Low', 5 or 'Info'
     // Default: 'Moderate'
     severity: SeverityLevel;
+
+    // [Optional] If the corresponding -meta.xml file should be included for analysis
+    // INTERNAL USE ONLY (NOT EXPOSED TO USER CONFIG)
+    // Default: false
+    include_metadata?: boolean;
```
stephen-carter-at-sf, line 51: The default isn't false... its undefined. I don't think we need to specify what the default is... so maybe just remove the commont on line 51.

packages/code-analyzer-regex-engine/src/plugin.ts
```diff
@@ -84,6 +84,15 @@ export function createBaseRegexRules(now: Date): RegexRules {
             violation_message: getMessage('AvoidGetHeapSizeInLoopRuleMessage'),
             severity: SeverityLevel.High,
             tags: ['Recommended', 'Performance']
+        },
+        MinVersionForAbstractVirtualClassesWithPrivateMethod: {
+            regex: (/(?=.*(?<=<apiVersion>)([1-9]|[1-5][0-9]|60)(\.[0-9])?(?=<\/apiVersion>))public\s+(virtual|abstract)\s+class\s+\w+\s*\n*\s*{\s*(?:[^{}]*{[^{}]*})*?\s+private\s+\w+\s+\w+\s*\([^)]*\)\s*{[^}]*}/gis).toString(),
```
stephen-carter-at-sf: If someone writes `<apiVersion>53.0</apiVersion>` then this looks good... but if they do:
```
<apiVersion> 53.0 </apiVersion>
```
or if they do
```
<apiVersion>
  53.0
</apiVersion>
```
Then this would miss it. So shouldn't we add in `\s*` before and after the numbers?

stephen-carter-at-sf: Also... why `\s*\n*\s*`... that seems redundant... Did you mean `\s*\n+\s*`? If not, then why not just do `\s*`

stephen-carter-at-sf: Why is `\s*(?:[^{}]*{[^{}]*})*?` used... why not just do `.*`  here? Seems equivalent. In fact, yours might not work if have a private method that is preceeded with a public method that contains a conditional statement.

packages/code-analyzer-regex-engine/test/engine.test.ts
```diff
@@ -363,6 +374,49 @@ describe('Tests for runRules', () => {
...
+        const expectedViolations: Violation[] = [
+            {
+                ruleName: "MinVersionForAbstractVirtualClassesWithPrivateMethod",
+                message: getMessage('MinVersionForAbstractVirtualClassesWithPrivateMethodRuleMessage'),
+                primaryLocationIndex: 0,
+                codeLocations: [
+                    {
+                        file: path.resolve(__dirname, "test-data", "apexClassWithPrivateMethod", "testAbstractClassWithPrivateMethod.cls"),
+                        startLine: 1,
+                        startColumn: 1,
```
stephen-carter-at-sf: This seems wrong. It appears that the violation doesn't point to the private method... but points to the start of the class up until the end of the private method. I would have expected the startLine to be the start of the private method.

packages/code-analyzer-regex-engine/test/engine.test.ts
```diff
@@ -363,6 +374,49 @@ describe('Tests for runRules', () => {
...
+                ]
+            },
+            {
+                ruleName: "MinVersionForAbstractVirtualClassesWithPrivateMethod",
+                message: getMessage('MinVersionForAbstractVirtualClassesWithPrivateMethodRuleMessage'),
+                primaryLocationIndex: 0,
+                codeLocations: [
+                    {
+                        file: path.resolve(__dirname, "test-data", "apexClassWithPrivateMethod", "testVirtualClassWithPrivateMethod.cls"),
+                        startLine: 1,
```
stephen-carter-at-sf: ditto.

stephen-carter-at-sf (changes requested):

packages/code-analyzer-regex-engine/test/test-data/apexClassWithPrivateMethod/testClassWithoutCorrespondingMetadataFile.cls
```diff
@@ -0,0 +1,8 @@
+public abstract class Parent {
+    private void aPrivateMethod() {
+     System.debug('Parent private method is called.');
+    }
+    public void aMethod() {
+     aPrivateMethod();
+    }
+}
```
stephen-carter-at-sf, line 8: Can you please have at least one example where the public method comes before the private method?

packages/code-analyzer-regex-engine/src/plugin.ts
stephen-carter-at-sf: Lastly... I believe you need to do look aheads and behinds on everything except for the private method so that you get the correct start and end locations.

stephen-carter-at-sf: Try this as a use case please:

```
public abstract class Parent {
    public void aMethod() {
     if(true) {
       aPrivateMethod();
     }
    }


    private void aPrivateMethod() { // This should be the start line
     if(true) {
       System.debug('Parent private method is called.');
     }
    } // This should be the end line
}
```

packages/code-analyzer-regex-engine/src/engine.ts
stephen-carter-at-sf: Just saw your previous comment. Yeah I think we'll have to ask John if he wants us to break the contract of going outside the provided workspace.
Notice we do something similar with the retirejs engine. See engine.ts line 177 which says 
```
return filesSet.has(`${fileInfo.dir}${path.sep}${fileInfo.name}.resource-meta.xml`);
```
where the fileSet is only the files from the workspace.

packages/code-analyzer-regex-engine/src/plugin.ts
jag-j: I agree with all the suggestions, except for the space before and after version number. Since this is an autogenerated file, I am not sure if it is worth catching the other variants. I will leave it as is. I will include the above class in one of the tests.

stephen-carter-at-sf: I believe the regular expression you want is:

```
// Part to match (using look behind) that we are in an abstract/virtual class (using look behind):
//    (^|\s+)(virtual|abstract)[^{]*\s+class\s+.*
// Part to match a private method signature:
//    private [^{]+\([^)]*\)\s*{
// Part to match (using look ahead) that the meta data has apiVersion less than 61.0:
//    .*<apiVersion>\s*([1-9]|[1-5][0-9]|60)(\.[0-9])?\s*<\/apiVersion>
const re = /(?<=(^|\s+)(virtual|abstract)[^{]*\s+class\s+.*)private [^{]+\([^)]*\)\s*{(?=.*<apiVersion>\s*([1-9]|[1-5][0-9]|60)(\.[0-9])?\s*<\/apiVersion>)/gis
```

stephen-carter-at-sf (approved):

PR #102, opened 2024-10-08
FIX(engine-api): @W-16606013@: Fix validateArray to actually normalize values

stephen-carter-at-sf:
* Anytime we touch engine-api we unfortunately need to update all the modules it depends on to keep them up to date.

### DIFF
--- a/packages/code-analyzer-engine-api/src/config.ts
+++ b/packages/code-analyzer-engine-api/src/config.ts
@@ -185,7 +185,7 @@ export class ValueValidator {
             throw new Error(getMessage('ConfigValueMustBeOfType', fieldPath, 'array', getDataType(value)));
         }
         if (elementValidator) {
-            value.every((element, index) => elementValidator(element, `${fieldPath}[${index}]`));
+            value = value.map((element, index) => elementValidator(element, `${fieldPath}[${index}]`));
         }
         return value as T[];
     }
--- a/packages/code-analyzer-engine-api/test/config.test.ts
+++ b/packages/code-analyzer-engine-api/test/config.test.ts
@@ -83,6 +83,11 @@ describe("Tests for ValueValidator", () => {
             getMessage('ConfigValueMustBeOfType', 'someFieldName[1]', 'string', 'number'));
     });
 
+    it("When a custom element validator is given to validateArray, then the resolved elements are returned in the array", () => {
+        expect(ValueValidator.validateArray(['a',2], 'someFieldName', (_e: unknown, _p: string) => 'resolved')).toEqual(
+            ['resolved', 'resolved']);
+    });
+
     it("When an absolute file is given to validateFile, then it is returned", () => {
         const inputFile: string = path.resolve(__dirname, 'config.test.ts');
         expect(ValueValidator.validateFile(inputFile, 'someFieldName')).toEqual(inputFile);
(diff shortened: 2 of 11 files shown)


packages/code-analyzer-engine-api/src/config.ts
```diff
@@ -185,7 +185,7 @@ export class ValueValidator {
             throw new Error(getMessage('ConfigValueMustBeOfType', fieldPath, 'array', getDataType(value)));
         }
         if (elementValidator) {
-            value.every((element, index) => elementValidator(element, `${fieldPath}[${index}]`));
+            value = value.map((element, index) => elementValidator(element, `${fieldPath}[${index}]`));
```
stephen-carter-at-sf, line 188: This is the bug fix.

packages/code-analyzer-engine-api/test/config.test.ts
```diff
@@ -434,6 +439,12 @@ describe("Tests for ConfigValueExtractor", () => {
             getMessage('ConfigValueMustBeOfType','engines.dummy.some_field2[1]', 'string', 'number'));
     });
 
+    it("When calling extractRequiredArray with a custom validated, then the returned array contains resolved elements", () => {
```
jfeingold35, line 442: Tiny nitpick: "With a custom validated"?

jfeingold35: Tiny nitpick: "With a custom validated"?

packages/code-analyzer-engine-api/test/config.test.ts
```diff
@@ -455,6 +466,12 @@ describe("Tests for ConfigValueExtractor", () => {
             getMessage('ConfigValueMustBeOfType','engines.dummy.some_field2[1]', 'string', 'number'));
     });
 
+    it("When calling extractArray with a custom validated, then the returned array contains resolved elements", () => {
```
jfeingold35: Tiny nitpick: "With a custom validated"?

packages/code-analyzer-engine-api/src/config.ts
jfeingold35: Question: Are there any possible side-effects of re-assigning `value` instead of creating a new array?
Also, is "validator" the best name for this kind of thing anymore, since it's now doing something more like processing?

stephen-carter-at-sf: All of the validation methods return the value that they validate. Most validators leave the value untouched... but our pathValidator for example always normalizes the path to be an absolute path for example during validation. I personally have gotten used to this. I don't want to say normalize or process... because validate is the main piece. But ideally folks use the extract* methods primarily which I think already has a good name. But I didn't want to confuse the two... so I think this is good enough.

packages/code-analyzer-engine-api/test/config.test.ts
stephen-carter-at-sf: Fixed.

packages/code-analyzer-engine-api/test/config.test.ts
stephen-carter-at-sf: Fixed.

packages/code-analyzer-engine-api/src/config.ts
stephen-carter-at-sf: No possible side-effects I can think of since we already restrict our configuration values to only be able to be string, boolean, number, or array of those primitives... or an object of those. Can you think of any side effects?

jfeingold35: Nope, but it felt like it was worth raising.

jfeingold35 (approved):

PR #105, opened 2024-10-10
NEW @W-16360921@ FlowTest runRules functionality implemented

jfeingold35:


packages/code-analyzer-flowtest-engine/test/test-data/executable-scripts/error-thrower.py
```diff
@@ -4,7 +4,6 @@
 print('Logging the number "1" to stderr', file=sys.stderr)
 print('Logging the number "2" to stderr', file=sys.stderr)
 print('Logging the number "87" to stdout')
-# Print an empty string to stderr
-print('', file=sys.stderr)
+print('Logging a string\nwith newline characters\n\nin it\nto stderr', file=sys.stderr)
```
jfeingold35, line 7: This should make the associated test less flaky.

packages/code-analyzer-flowtest-engine/test/test-data/example-workspaces/contains-multiple-flows/example.flow-meta.xml
jfeingold35, line 1: Copied from FlowTest's integration test folder.

packages/code-analyzer-flowtest-engine/test/test-data/example-workspaces/contains-multiple-flows/inner_subflow_example.flow-meta.xml
jfeingold35, line 1: Copied from FlowTest's integration test folder.

packages/code-analyzer-flowtest-engine/test/test-data/example-workspaces/contains-multiple-flows/subflow_test1.flow-meta.xml
jfeingold35, line 1: Copied from FlowTest's integration test folder.

packages/code-analyzer-flowtest-engine/test/python/FlowTestCommandWrapper.test.ts
```diff
@@ -28,13 +33,82 @@ describe('FlowTestCommandWrapper implementations', () => {
...
+                const statusProcessorFunction = (completionPercentage: number) => {
+                    completionPercentages.push(completionPercentage);
+                };
+
+                beforeAll(async () => {
+                    results = await wrapper.runFlowTestRules(path.join(PATH_TO_WORKSPACES, 'contains-multiple-flows'), statusProcessorFunction);
+                }, 30000);
+
+                it('Correctly reads and parses results', async () => {
+                    const goldfileName: string = os.platform() === 'win32' ? 'results-windows.goldfile.json' : 'results-non-windows.goldfile.json';
```
jfeingold35, line 60: FlowTest's result ordering is inconsistent between Windows and Unix. Until that's fixed, I've just gone ahead and made Windows and Unix use different goldfiles.

packages/code-analyzer-flowtest-engine/src/utils.ts
```diff
@@ -0,0 +1,90 @@
...
+        violations: []
+    };
+
+    for (const queryName of Object.keys(flowTestExecutionResult.results)) {
+        const flowTestRuleResults = flowTestExecutionResult.results[queryName];
+        for (const flowTestRuleResult of flowTestRuleResults) {
+            const ruleName = toCodeAnalyzerName(flowTestRuleResult.query_name);
+            // FlowTest runs extremely quickly, and its rule selection is fiddly. So it's easier to just run all the rules,
+            // and then throw away results for rules that the user didn't request.
+            if (!requestedRulesSet.has(ruleName)) {
```
jfeingold35: Drawing attention to this part.

packages/code-analyzer-flowtest-engine/src/utils.ts
```diff
@@ -0,0 +1,90 @@
...
+                continue;
+            }
+            // FlowTest's file discovery is known to be overzealous relative to our Workspaces. So we iterate over each
+            // violation's Flow Path, and if it goes through a file that isn't in the Workspace, then we discard it.
+            const flowNodes: FlowNodeDescriptor[] = flowTestRuleResult.flow;
+            if (flowNodes.some(node => !allowedFilesSet.has(node.flow_path))) {
```
jfeingold35: Drawing attention to this part.

packages/code-analyzer-flowtest-engine/src/engine.ts
```diff
@@ -41,14 +52,24 @@ export class FlowTestEngine extends Engine {
...
+    public async runRules(ruleNames: string[], runOptions: RunOptions): Promise<EngineRunResults> {
         this.emitRunRulesProgressEvent(0);
-
-        this.emitRunRulesProgressEvent(10);
+        const workspaceRoot: string | null = runOptions.workspace.getWorkspaceRoot();
+        // If we can't identify a single directory as the root of the workspace, then there's nothing for us to pass into
+        // FlowTest. So just throw an error and be done with it.
+        // NOTE: The only known case where this can occur is if a Windows user is scanning files on two different drives
+        //       (e.g., "C:" and "D:"). Since this is an extreme edge case, it's one we're willing to tolerate.
+        if (workspaceRoot === null) {
```
jfeingold35, line 66: Drawing attention to this part.

packages/code-analyzer-flowtest-engine/src/utils.ts
```diff
@@ -0,0 +1,90 @@
...
+        }
+    }
+    return results;
+}
+
+function toCodeAnalyzerName(queryName: string): string {
+    return queryName.replaceAll('Flow: ', '').replaceAll(' ', '-').toLowerCase();
+}
+
+function toCodeAnalyzerViolationMessage(flowTestRuleResult: FlowTestRuleResult): string {
```
jfeingold35: The message generation part will definitely have to change once we've added `CodeLocation.comment`, and it might also change depending on some pending feedback from Robert. But for now, it is what it is.

packages/code-analyzer-flowtest-engine/src/engine.ts
```diff
@@ -57,42 +78,13 @@ function fileIsFlowFile(fileName: string): boolean {
...
-    }
-}
+/**
+ * Accepts a percentage indicating the completion percentage of the underlying FlowTest tool, and converts it into a
+ * percentage representing the completion percentage of the engine as a whole.
+ * @param flowTestPercentage Completion percentage received from the FlowTest tool.
+ */
+// istanbul ignore next
+function normalizeRelativeCompletionPercentage(flowTestPercentage: number): number {
+    const percentageSpread: number = POST_INVOCATION_RUN_PERCENT - PRE_INVOCATION_RUN_PERCENT;
```
jag-j, line 92: If I understand this correct, PRE_INVOCATION_RUN_PERCENT makes sense because of all the overhead needed to run python. Do we also need the POST_INVOCATION_RUN_PERCENT as well?

packages/code-analyzer-flowtest-engine/src/python/FlowTestCommandWrapper.ts
```diff
@@ -76,4 +105,139 @@ export class RunTimeFlowTestCommandWrapper implements FlowTestCommandWrapper {
...
+            // Note: NaN is the only value in ECMA that is not equal to itself, so this if-expression will only be
+            // satisfiable if the parse attempt failed.
+            if (percentageFlowsScanned === percentageFlowsScanned) {
+                completionPercentageHandler(percentageFlowsScanned);
+            }
+        }
+
+        await this.pythonCommandExecutor.exec(pythonArgs, processStdout);
+
+        const outputFileContents: string = await fsp.readFile(tmpFile, {encoding: 'utf-8'});
```
jag-j, line 138: Can we cleanup/delete the tmpFile ?

packages/code-analyzer-flowtest-engine/src/python/FlowTestCommandWrapper.ts
```diff
@@ -76,4 +105,139 @@ export class RunTimeFlowTestCommandWrapper implements FlowTestCommandWrapper {
...
+            // If the message doesn't start with our status delimiter, then we don't care.
+            if (!stdoutMsg.startsWith(STATUS_DELIMITER)) {
+                return;
+            }
+            // Rip off the status delimiter and then attempt to parse a Float out of the status update.
+            const percentageFlowsScanned: number = parseFloat(stdoutMsg.slice(STATUS_DELIMITER.length));
+            // If we successfully parsed a Float, pass it through the completion percentage processor.
+            // Note: NaN is the only value in ECMA that is not equal to itself, so this if-expression will only be
+            // satisfiable if the parse attempt failed.
+            if (percentageFlowsScanned === percentageFlowsScanned) {
```
jag-j: This seemed confusing first and once I read the comment it makes sense. Would it make sense to use isNaN function instead to make it cleaner?

packages/code-analyzer-flowtest-engine/src/engine.ts
```diff
@@ -41,14 +52,24 @@ export class FlowTestEngine extends Engine {
...
+        // NOTE: The only known case where this can occur is if a Windows user is scanning files on two different drives
+        //       (e.g., "C:" and "D:"). Since this is an extreme edge case, it's one we're willing to tolerate.
+        if (workspaceRoot === null) {
+            throw new Error(getMessage('WorkspaceLacksIdentifiableRoot', runOptions.workspace.getFilesAndFolders().join(', ')));
+        }
+        this.emitRunRulesProgressEvent(PRE_INVOCATION_RUN_PERCENT);
+        const percentageUpdateHandler = /* istanbul ignore next */ (percentage: number) => {
+            this.emitDescribeRulesProgressEvent(normalizeRelativeCompletionPercentage(percentage));
+        }
+        const executionResults: FlowTestExecutionResult = await this.commandWrapper.runFlowTestRules(workspaceRoot, percentageUpdateHandler);
```
stephen-carter-at-sf, line 73: I'm not sure I agree that we should be sending in the workspace root as the directory to be scanned... especially since if the user is suppling flows on Windows from 2 different drives, we error out above.

To avoid all that, can't we just get all the relevant workspace files and use the -f flag of flowtest to scan 1 flow at a time? This would tremendously help with calculating progress as well wouldn't it?

When describing the rules... we are already calculating the relevant flow files... so I would cache those (based on the workspace id) and then reuse that calculation here in the run method.

stephen-carter-at-sf: If you are worried about calling python too many times... then we should have a python script (a wrapper around flowtest) that we can call... that in a single invocation can invoke flowtest multiple times within the same python process that can unify the output for us before sending it back as a results file to the typescript side.

packages/code-analyzer-flowtest-engine/src/engine.ts
stephen-carter-at-sf: This makes me wonder if we should switch to "getWorkspaceRoots" in our engine api that returns an array (assuming we can calculate it). Then the engines can check the length and decide what to do.

packages/code-analyzer-flowtest-engine/src/engine.ts
stephen-carter-at-sf: Yes... he is trying to convert a percentage between 0 and 100 to fit between 10% and 90%... so he needs to do 10 + (fraction*80) so that he doesn't go over 90%.

packages/code-analyzer-flowtest-engine/src/messages.ts
```diff
@@ -16,6 +16,12 @@ const MESSAGE_CATALOG: {[key: string]: string} = {
     CouldNotParseRuleDescriptions:
         `Could not parse rule descriptions from FlowTest output: %s`,
 
+    ResultsFileNotValidJson:
+        `Results file contents are not a valid JSON: %s`,
+
+    CouldNotParseExecutionResults:
+        `Could not parse results from %s`,
```
stephen-carter-at-sf: These are not error messages that the user should ever see... these are internal only error message that ideally the user shouldn't see... to help us debug issues. So I don't think they need to be in the message catalog.

packages/code-analyzer-flowtest-engine/src/python/FlowTestCommandWrapper.ts
```diff
@@ -76,4 +105,139 @@ export class RunTimeFlowTestCommandWrapper implements FlowTestCommandWrapper {
...
+                return false;
+            }
+            for (const ruleResult of result) {
+                if (!this.ruleResultIsValid(ruleResult)) {
+                    return false;
+                }
+            }
+        }
+        return true;
+    }
```
stephen-carter-at-sf, line 182: What is this for? Do we really need to validate what we get from flowtest? Isn't that the job of flowtest to have its own unit tests?

packages/code-analyzer-flowtest-engine/src/python/FlowTestCommandWrapper.ts
stephen-carter-at-sf: Would flowtest ever spit out something to stdout with `**STATUS:` followed by something other than a number?

If not, then why do we need this?

packages/code-analyzer-flowtest-engine/src/python/FlowTestCommandWrapper.ts
```diff
@@ -76,4 +105,139 @@ export class RunTimeFlowTestCommandWrapper implements FlowTestCommandWrapper {
...
+        if (!this.executionResultsAreValid(parsedResults)) {
+            throw new Error(getMessage('CouldNotParseExecutionResults', JSON.stringify(parsedResults)));
+        }
+
+        return parsedResults;
+    }
+
+    private createTmpFileName(): string {
+        return path.join(os.tmpdir(), `flow-test-execution-${Date.now()}.json`);
+    }
```
stephen-carter-at-sf: You should be using the tmp package to create a temporary directory which cleans up after itself when the node process terminates.

See an example of this with pmd-wrapper.ts where we have

```
const tmpDirAsync = promisify((options: tmp.DirOptions, cb: tmp.DirCallback) => tmp.dir(options, cb));

/**
 * Creates a temporary directory that eventually cleans up after itself
 * @param parentTempDir - if supplied, then a temporary folder is placed directly underneath this parent folder.
 */
export async function createTempDir(parentTempDir?: string) : Promise<string> {
    return tmpDirAsync({dir: parentTempDir, keep: false, unsafeCleanup: true});
}
```

packages/code-analyzer-flowtest-engine/src/python/FlowTestCommandWrapper.ts
stephen-carter-at-sf, line 138: Also note that you can just do the equivalent: fsp.readFile(tmpFile, 'utf-8');

packages/code-analyzer-flowtest-engine/src/engine.ts
jfeingold35: Correct.

packages/code-analyzer-flowtest-engine/src/python/FlowTestCommandWrapper.ts
jfeingold35: The `tmp` module does that automatically, with no need for us to set anything.

packages/code-analyzer-flowtest-engine/src/python/FlowTestCommandWrapper.ts
jfeingold35: Yeah, that's probably reasonable. I'll do that.

packages/code-analyzer-flowtest-engine/src/utils.ts
stephen-carter-at-sf, line 1: Nitpick: Typically the types of functions you put into a separate utils file are those that can be reused. But it looks like all the functions in this file are specific to the implementation detail of the flowtest engine. Is there a reason why they are here instead of inside of the engine.ts file?

packages/code-analyzer-flowtest-engine/src/engine.ts
jfeingold35: @stephen-carter-at-sf 
1. The "Windows user scanning flows on different drives" scenario seems like a corner case that would be acceptable to error out in. I'm not sure it's worth bending over backwards for that case alone.
2. With the current implementation, running on each file one-by-one would be disastrous for runtime because of the dependency setup. It also doesn't make progress calculation any easier, since it's already trivial to parse out the Status Update percentages we get from FlowTest as we're currently doing.

packages/code-analyzer-flowtest-engine/src/messages.ts
jfeingold35: The user _should_ never seen them. But especially since things are so in flux with this engine right now, I don't want to discount the possibility that they _do_ see it without us gracefully handling it.

packages/code-analyzer-flowtest-engine/src/python/FlowTestCommandWrapper.ts
jfeingold35: Honestly, I'm a little nervous about the results changing out from under us or something, and I want to lock as much behavior in as I possibly can.

packages/code-analyzer-flowtest-engine/src/python/FlowTestCommandWrapper.ts
jfeingold35: Yes. It will spit out things like "**STATUS: Done scanning flows", or the like.

packages/code-analyzer-flowtest-engine/src/python/FlowTestCommandWrapper.ts
jfeingold35: Good call. Not sure how I missed that.

packages/code-analyzer-flowtest-engine/src/python/FlowTestCommandWrapper.ts
jfeingold35: the second argument can just be the string `utf-8`? Didn't know that. Neat.

packages/code-analyzer-flowtest-engine/src/utils.ts
jfeingold35: Because when I left it in, the conversion code drowned out everything else in the file. I can put it back in if you want.

packages/code-analyzer-flowtest-engine/src/utils.ts
stephen-carter-at-sf: This is ok to discard the parent workflow because all violations show up at the sink (i.e. in the child workflow)? Is that true? If isn't true... then I'm not sure how I feel about discarding the parent.

packages/code-analyzer-flowtest-engine/src/utils.ts
stephen-carter-at-sf: I personally would prefer implementation detail left as non-exported functions in the file it is being invoked from so that the implementation detail has no chance of leaking. But I'll leave this up to you at the end of the day. Just wanted to call it out to drive the thought.

packages/code-analyzer-flowtest-engine/src/utils.ts
jfeingold35: If the child workflow isn't in the workspace, then in a reasonable world FlowTest wouldn't have been able to see it and thus wouldn't have thrown violations because of it, right? So this step is necessary to preserve the expectations for how it should behave.

packages/code-analyzer-flowtest-engine/src/python/FlowTestCommandWrapper.ts
stephen-carter-at-sf: That's unfortunately. I am not sure how I feel about attempting to convert it to a floating point number first and then aftewards checking. I'd prefer to use a regular expression or something to confirm the text is indeed a number first before converting.

packages/code-analyzer-flowtest-engine/src/utils.ts
jfeingold35: Reasonable. I'll move it back.

packages/code-analyzer-flowtest-engine/src/python/FlowTestCommandWrapper.ts
stephen-carter-at-sf: Well I would think our tests would fail if something changed out from under us. But I would hope that if flowtest produces a results file... that we should be assured that it is of valid format. Otherwise, things would blow up in some other way I'm sure. I just don't think we need all this validation code. Doing a cast to the data structure and then having our tests (or production code) consume that datastructure would be sufficient in my mind. Thoughts?

packages/code-analyzer-flowtest-engine/src/python/FlowTestCommandWrapper.ts
jfeingold35: I'm not too worried about it. I spent a while testing how `parseFloat` behaves if a number is at the beginning, middle, or end of the input string. It seems like if the beginning of the string is a number then it will pull that out, and otherwise it'll return NaN.

packages/code-analyzer-flowtest-engine/src/messages.ts
stephen-carter-at-sf: Fair enough... but we should probably say something like: "Unexected error:" at the very least since the user can't do anything about it. It is really for us and not the end user. 

It is equivalent to line 80 in utils.ts in my mind... in that if the user ever say it - they should know that it isn't their fault and they should just forward the message to us.

packages/code-analyzer-flowtest-engine/src/python/FlowTestCommandWrapper.ts
jfeingold35: Fair enough. I can probably remove it and it would be fine.

packages/code-analyzer-flowtest-engine/src/engine.ts
stephen-carter-at-sf: I see your point. A python wrapper would help eleviate most of your concerns though. But with this solution, does flowtest recurively scan all subdirectories looking for flows?

jfeingold35: Just tested, and yes it does.

packages/code-analyzer-flowtest-engine/src/engine.ts
jag-j: So my understanding is correct. But still wondering about the last 10%. Why is it required? When the flowtest engine is done, does the rest take 10%? I understand the initial 10% is required.

jfeingold35: The last 10% is the post-processing we do on the FlowTest results. i.e., discarding the ones we don't want to include, and converting them into our format.

packages/code-analyzer-flowtest-engine/test/engine.test.ts
```diff
@@ -136,21 +141,161 @@ describe('Tests for the FlowTestEngine', () => {
...
+
+        it.each([
+            {case: 'Empty results', inputFile: 'empty-results.json', goldfile: 'empty-results.goldfile.json'},
+            {case: 'Non-empty results', inputFile: 'non-empty-results.json', goldfile: 'non-empty-results.goldfile.json'}
+        ])('Converts FlowTest format to Code Analyzer format. Case: $case', async ({inputFile, goldfile}) => {
+            // ==== SETUP ====
+            const fakeFlowTestResults: FlowTestExecutionResult = await readAndParseInputFile(path.join(PATH_TO_SAMPLE_RESULTS, inputFile));
+
+            // Instantiate the engine with a stubbed wrapper that will return the desired FlowTestExecutionResult.
+            const engine: FlowTestEngine = new FlowTestEngine(new StubCommandWrapper([], fakeFlowTestResults));
```
stephen-carter-at-sf: I understand the desire to speed up testing by introducing fake flowtest results... but this makes the tests less robust and subject to change in the future. You'll notice with my pmd engine... although I have a pmd-wrapper - I don't have any tests for it. All my tests do the full round trip in order to guarantee that the tests will never have to change as implementation detail changes. But again - this only works if execution is fast enough so that our tests don't take too long... so I get it.

But my real question is, do we have at least one end to end test or are we relying on A --> B  and B --> C tests only?

jfeingold35: I'll add at least one E2E test.

packages/code-analyzer-flowtest-engine/src/python/FlowTestCommandWrapper.ts
jfeingold35: @stephen-carter-at-sf ,
OH. I remember why I didn't do this. FlowTest currently refuses to overwrite an existing file. So we need to generate a name for a temporary file, and _then_ feed that name into FlowTest so it _creates_ the file.

jag-j (approved):

packages/code-analyzer-flowtest-engine/src/python/FlowTestCommandWrapper.ts
```diff
@@ -55,7 +84,6 @@ export class RunTimeFlowTestCommandWrapper implements FlowTestCommandWrapper {
         const pythonArgs: string[] = [
             PATH_TO_PIPX_PYZ,
             'run',
-            '--no-cache',
```
jfeingold35, line 58: The end-to-end test was broken when I first added it, and I was able to determine that running Python twice in extremely quick succession with the `--no-cache` flag was causing the error about building the Wheel. Removing it seems to have fixed that, at least on my machine.

packages/code-analyzer-flowtest-engine/test/python/FlowTestCommandWrapper.test.ts
stephen-carter-at-sf: Ideally we shouldn't be needed to test the direct output of flowtest. That is, instead of doing A (user input) --> B (flowtest) --> C (code analyzer output) testing... we would just have A --> C testing. Things like ordering is something we can fix with sorting routines from B-->C as an implementation detail.

packages/code-analyzer-flowtest-engine/src/utils.ts
stephen-carter-at-sf: Thanks for viewing this with me together... please add a comment showing what flow does when the subflow is removed which validates that we are safe to review this (since we run from the workspace root).

packages/code-analyzer-flowtest-engine/src/python/FlowTestCommandWrapper.ts
stephen-carter-at-sf: Final thought... look to see if pipx allows us to either control where the cache lives... and/or the permissions that is given to the files when written.

stephen-carter-at-sf (approved):

PR #107, opened 2024-10-16
FIX (RegexEngine) @W-15490095@ Fix regex that catches Limits.getHeapSize() in a loop

jag-j:


packages/code-analyzer-regex-engine/src/plugin.ts
```diff
@@ -79,7 +79,7 @@ export function createBaseRegexRules(now: Date): RegexRules {
         },
         AvoidOldSalesforceApiVersions: createAvoidOldSalesforceApiVersionsRule(now),
         AvoidGetHeapSizeInLoop: {
-            regex: (/(for|while)\s*\([^)]*\)\s*\{[^}]*\bLimits\.getHeapSize\(\);|do\s*\{[^}]*\bLimits\.getHeapSize\(\);[^}]*\}\s*while\s*\([^)]*\);/gi).toString(),
+            regex: (/(?<=(for|while)\s*\([^)]*\)\s*\{[^}]*?)\bLimits\.getHeapSize\(\)|(?<=(do)\s*\{[^}]*?)\bLimits\.getHeapSize\(\)/gi).toString(),
```
stephen-carter-at-sf: Any reason why \b is on the outside and not on the inside. Also, is there a reason why do is in parenthesis now?

Slight modification to yours would be:
```
/(?<=(for|while)\s*\([^)]*\)\s*\{[^}]*?\b)Limits\.getHeapSize\(\)|(?<=do\s*\{[^}]*?\b)Limits\.getHeapSize\(\)/gi
```
or just
```
/(?<=(for|while)\s*\([^)]*\)\s*\{[^}]*\b)Limits\.getHeapSize\(\)|(?<=do\s*\{[^}]*\b)Limits\.getHeapSize\(\)/gi
```

stephen-carter-at-sf (changes requested):

packages/code-analyzer-regex-engine/src/plugin.ts
jag-j: Done!

stephen-carter-at-sf (approved):

PR #110, opened 2024-10-29
CHANGE @W-16867469@ Added comment property to code location in outfile

jfeingold35:


packages/code-analyzer-core/test/rule-selection.test.ts
```diff
@@ -311,15 +315,17 @@ describe('Tests for selecting rules', () => {
...
+        expect(ruleSelectionProgressEvents).toHaveLength(11);
         const expectedPercentages: number[] = [
             0, // initial progress always reported
-            10, // average of 20 from engine1 and 0 from engine2
-            40, // average of 80 from engine1 and 0 from engine2
-            55, // average of 80 from engine1 and 30 from engine2
-            85, // average of 80 from engine1 and 90 from engine2
-            95, // average of 100 from engine1 and 90 from engine2
-            100, // average of 100 from engine1 and 100 from engine 2
+            6.666666666666667, // average of 20 from engine1, 0 from engine2, 0 from engine3
```
jfeingold35, line 321: @stephen-carter-at-sf , why can we assume a consistent ordering of the emit messages? I'm willing to believe that we can, but I'd just like to understand why.

stephen-carter-at-sf: We could if you want... another option is to change the test to just check the first (0) and last values (100) ... and then check that all other values increase each time... so update line 332 to just check that the type is equal to EventType.RuleSelectionProgressEvent as its own assertion, the timestamp is equal to sampleTimestamp with its own assertion, and then just check the precentComplete[i] > percentComplete[i-1]. 

But note that this change theoretically only gets updated if you update the shared stubs. In the future we might want to just create new stubs instead of modifying the existing ones that these tests depend on.

packages/code-analyzer-core/src/output-format.ts
```diff
@@ -57,19 +57,67 @@ type ViolationOutput = {
...
+
 class CsvOutputFormatter implements OutputFormatter {
     format(results: RunResults): string {
         const violationOutputs: ViolationOutput[] = toViolationOutputs(results.getViolations(), results.getRunDirectory());
         const options: CsvOptions = {
             header: true,
             quoted_string: true,
             columns: ['rule', 'engine', 'severity', 'type', 'tags', 'file', 'line', 'column',
-                'endLine', 'endColumn', 'pathLocations', 'message', 'resources'],
+                'endLine', 'endColumn', 'locations', 'message', 'resources'],
```
stephen-carter-at-sf, line 132: I was expecting to also see primaryLocationIndex here... but I'm not sure if we need it for CSV format or not. What are your thoughts? Does the index make sense if it is a stringified list anyway? If you don't think it's needed - I'm cool with that. Up to you.

packages/code-analyzer-core/src/output-format.ts
```diff
@@ -57,19 +57,67 @@ type ViolationOutput = {
...
+    }
+
+    public getComment(): string | undefined {
+        return this.comment;
+    }
+
+    public toString(): string {
+        const commentPortion: string = this.comment ? ` (${this.comment})` : '';
+        return `${this.file}:${this.line}:${this.column}${commentPortion}`;
+    }
```
stephen-carter-at-sf: I see you removed the createLocationString function which has conditionals around the file, startline, startColumn, etc.

Because the interface has | undefined:
```export interface CodeLocation {
    getFile(): string | undefined
    getStartLine(): number | undefined
    getStartColumn(): number | undefined
    getEndLine(): number | undefined
    getEndColumn(): number | undefined
}
```

Then even though we control the only engine that is creating path locations... the interface allows a 3rd party engine to create code locations that don't have the file, line, column, endLine, endColumn possibly filled in.

At the top level this comes up with the UnexpectedEngineErrorViolation... but currently it doesn't come up within the getCodeLocationsOutput for the Flow and DataFlow types.

But I still think we should program to the interface and not to our internal expectations - because a 3rd party engine could very well set the  RuleType to Flow or DataFlow and then not contractually have to set these values - or they could just use something equivalent to our UndefinedCodeLocation.

So I recommend fixing this. You'll have to also update the getFile, getLIne, getColumn to also have `| undefined`.

Alternatively if you want to stick with this assumption... where you are putting in `!` marks on lines 75,76,77 ... maybe instead you should make these formal assertions with a human readable error message that says something like:
```
Unexpected error: Violation of type '%s' from engine '%s' produce a code location with an undefined file, line, or column field.
```
but this would require you putting the check outside of the constructor where you still have access to the violation information.

packages/code-analyzer-core/test/stubs.ts
```diff
@@ -46,6 +46,8 @@ export class StubEnginePlugin extends engApi.EnginePluginV1 {
             this.createdEngines.set(engineName, new StubEngine1(config));
         } else if (engineName == "stubEngine2") {
             this.createdEngines.set(engineName, new StubEngine2(config));
+        } else if (engineName == "stubEngine3") {
+            this.createdEngines.set(engineName, new StubEngine3(config));
```
stephen-carter-at-sf, line 50: In the future... feel free to create another stub plugin with its own stub engine local to the test file you want to update. This way you won't need to update all the tests everywhere that use this shared set of stubs. But this is all fine of course.

packages/code-analyzer-engine-api/src/results.ts
```diff
@@ -2,6 +2,7 @@ export type CodeLocation = {
     file: string
     startLine: number
     startColumn: number
+    comment?: string
```
stephen-carter-at-sf, line 5: As a fast follow on... we'll need to create a PR to update the version number of the engine-api and core... which will have a trickle effect on all other packages that depend on the engine-api.

We at some point should discuss a release strategy soon with regards to keeping these version numbers correct... i'm thinking we can update them to say something like
0.12.0-SNAPSHOT
for example when we make a change. Then prior to deployment we would update it to 0.12.0. To be discussed later... but let's not forget that the engine api has changed and thus we need to increment its version this month and release.

packages/code-analyzer-flowtest-engine/important.txt
```diff
@@ -0,0 +1,3 @@
+PIPX_DEFAULT_PYTHON=/Library/Frameworks/Python.framework/Versions/3.12/bin/python3 pipx run --no-cache --spec ./flowtest -- flowtest -h
+python3 ../../Downloads/pipx.pyz run --no-cache --spec . -- flowtest -j beep.json -f integration_test/test_data/Twenty_Questions_Troubleshooter.flow-meta.xml
+python3 /Users/jfeingold/Code/code-analyzer-core/packages/code-analyzer-flowtest-engine/pipx.pyz run --no-cache --spec . -- flowtest -h
```
stephen-carter-at-sf: What is this? Did you accidentally have this in your changelist. If so - please remove so that it doesn't get checked in.

stephen-carter-at-sf (changes requested):

packages/code-analyzer-core/src/output-format.ts
jfeingold35: My view was that the index didn't make sense for a stringified list.

packages/code-analyzer-core/src/output-format.ts
jfeingold35: Sure, I can change this so it no longer assumes that `startLine`, `startColumn`, and `file` are defined.

packages/code-analyzer-core/test/stubs.ts
jfeingold35: I actually considered doing that when I saw how many files I'd have to change, but then I decided it was actually probably worth making the changes, just so that the other tests use engines with rules that are Flow-typed in addition to the existing Standard and DataFlow ones.

packages/code-analyzer-flowtest-engine/important.txt
jfeingold35: Good catch, that was my bad.

packages/code-analyzer-core/src/output-format.ts
```diff
@@ -57,19 +57,67 @@ type ViolationOutput = {
...
+        return this.endColumn;
+    }
+
+    public getComment(): string | undefined {
+        return this.comment;
+    }
+
+    public toString(): string {
+        const commentPortion: string = this.comment ? ` (${this.comment})` : '';
+        return `${this.file}:${this.line}:${this.column}${commentPortion}`;
```
stephen-carter-at-sf: I still see that there is no conditionals on the file. If you look at how createLocationString was implemented... we wouldn't return an empty string if there was no file. And we wouldn't return the column info if there was no line info, etc.

Did you want to change this to what you have? If so.. what does it look like... would it say "undefined:3:undefined" if it has a line for example?

stephen-carter-at-sf (approved):

PR #113, opened 2024-10-31
NEW(cpd): @W-16113482@: Introduce CPD rules (describe only - no run yet)  and small PMD refactor

stephen-carter-at-sf:
In this review:
* Refactor PmdLanguage and PmdWorkspaceLiason to be agnostic of PMD (so it can be reused with CPD as well)
* Change 'ecmascript' visibility... to instead make 'javascript' show up instead with code analyzer
* Add in the CPD rules (assuming 1 rule per language so they can have language specific tags)

Later we will introduce the run implementation - for now we simply throw a warning message if someone runs a CPD rule.


packages/code-analyzer-pmd-engine/src/config.ts
```diff
@@ -104,10 +108,10 @@ class PmdConfigValueExtractor {
...
         const ruleLanguagesSet: Set<string> = new Set(ruleLanguages.map(l => l.toLowerCase()));
 
-        // Provide support for 'javascript' which is a supported alias of 'ecmascript'
-        if (ruleLanguagesSet.has('javascript')) {
-            ruleLanguagesSet.delete('javascript');
-            ruleLanguagesSet.add('ecmascript');
+        // Provide support for 'ecmascript' which is a supported alias of 'javascript'
+        if (ruleLanguagesSet.has('ecmascript')) {
+            ruleLanguagesSet.delete('ecmascript');
+            ruleLanguagesSet.add('javascript');
```
stephen-carter-at-sf, line 114: I think we should just hide the fact that PMD uses 'ecmascript' as the identifier behind the scenes. So I refactored to make 'javascript' show up everywhere in code analyzer instead of 'ecmascript'. But users can still select 'ecmascript' as a language to be aliased to 'javascript'.

packages/code-analyzer-pmd-engine/src/constants.ts
```diff
@@ -2,67 +2,73 @@
...
-    JAVA = 'java',
-    JAVASCRIPT = 'ecmascript',
-    VISUALFORCE = 'visualforce',
-    XML = 'xml'
+export enum LanguageId {
+    APEX = 'apex', // [CPD+PMD]
+    HTML = 'html', // [CPD+PMD]
+    JAVA = 'java', // [CPD+PMD] - We might remove support for this in the near future since we aren't seeing any valid use cases for it within Salesforce projects
+    JAVASCRIPT = 'javascript', // [CPD+PMD] Note that this id gets converted to 'ecmascript' when communicating to PMD since it uses 'ecmascript' instead as the identifier
+    TYPESCRIPT = 'typescript', // [CPD only - No PMD support] Note that TSLanguageModule is defined within the pmd-javascript jar file so no additional jar to add in here.
```
stephen-carter-at-sf, line 10: Renamed this to be not PMD specific anymore. These are just language identifiers for code analyzer for both pmd and cpd engines.

So instead of listing the languages that PMD recognizes, this turned into the languages that code analyzer knows about for both the cpd and pmd engines.

packages/code-analyzer-pmd-engine/src/pmd-engine.ts
```diff
@@ -22,10 +22,10 @@ export class PmdEngine extends Engine {
     static readonly NAME: string = "pmd";
 
     private readonly pmdWrapperInvoker: PmdWrapperInvoker;
-    private readonly selectedLanguages: PmdLanguage[];
+    private readonly selectedLanguages: LanguageId[];
     private readonly customRulesets: string[];
 
-    private pmdWorkspaceLiaisonCache: Map<string, PmdWorkspaceLiaison> = new Map();
+    private workspaceLiaisonCache: Map<string, WorkspaceLiaison> = new Map();
```
stephen-carter-at-sf, line 28: Sorry for all the noise. I refactored the PmdWorkspaceLiaison to just be the WorkspaceLiaison because it is something I can use for both pmd and cpd engines.

packages/code-analyzer-pmd-engine/src/pmd-engine.ts
```diff
@@ -226,9 +176,20 @@ function fetchRuleInfoByRuleName(ruleInfoList: PmdRuleInfo[], uniqueRuleName: st
...
+}
+
+function toPmdRuleLanguage(languageId: LanguageId): string {
+    // We must convert 'javascript' to 'ecmascript' since PMD actually uses 'ecmascript' as the identifier instead of 'javascript'
+    return languageId == LanguageId.JAVASCRIPT ? 'ecmascript' : languageId;
+}
+
+function toLanguageId(pmdRuleLanguage: string): LanguageId {
+    // We must convert 'ecmascript' back to 'javascrijpt'
+    return pmdRuleLanguage == 'ecmascript' ? LanguageId.JAVASCRIPT : pmdRuleLanguage as LanguageId;
```
stephen-carter-at-sf, line 194: So now instead of using 'ecmascript' everywhere... I use 'javascript' everywhere and at the boundaries of calling through to the pmd-wrapper... i switch it to 'ecmascript' and back when needed.

This way, the tags say "javascriptLanguage" and the conficted rule names have "RuleName-javascript" instead of saying "ecmascriptLanguage" an "RuleName-ecmascript".

So again - this was the refactoring of PMD to make the lanuage handling better.

packages/code-analyzer-pmd-engine/src/utils.ts
```diff
@@ -59,6 +61,57 @@ export class JavaCommandExecutor {
     }
 }
 
+// noinspection JSMismatchedCollectionQueryUpdate (IntelliJ is confused about how I am setting the private values, suppressing warnings)
+export class WorkspaceLiaison {
```
stephen-carter-at-sf, line 65: This isn't new at all... it just got renamed and moved. It lived in the pmd-engine file before... so nothing to see here.

packages/code-analyzer-pmd-engine/test/cpd-engine.test.ts
```diff
@@ -1,19 +1,56 @@
...
+describe('Tests for the getName method of CpdEngine', () => {
     it('When getName is called, then cpd is returned', () => {
         const engine: CpdEngine = new CpdEngine();
         expect(engine.getName()).toEqual('cpd');
     });
+});
+
+describe('Tests for the describeRules method of PmdEngine', () => {
+    it('When using defaults without workspace, then all rules are returned', async () => {
+        // TODO: BEFORE GA, we need to eventually decide on what the default languages should be. For now we just return all.
```
stephen-carter-at-sf, line 20: Since we are still waiting on some decisions - I left in some TODOs. Need to have quick team huddle to decide if we want to keep java as a supported language for both PMD and CPD... and we need to decide what the default languages available for CPD should be. And do we even want to make this configurable.

packages/code-analyzer-eslint-engine/src/engine.ts
```diff
@@ -97,7 +97,7 @@ export class ESLintEngine extends Engine {
 
                 /* istanbul ignore else */
                 if (ruleNames.has(ruleName)) {
-                    violations.push(toViolation(eslintResult.filePath, resultMsg));
+                    violations.push(violation);
```
stephen-carter-at-sf, line 100: Cleaning this up while I'm in here.

packages/code-analyzer-engine-api/src/rules.ts
```diff
@@ -7,16 +7,30 @@ export enum SeverityLevel {
 }
 
 export enum RuleType {
-    Standard = "Standard",
-    DataFlow = "DataFlow",
+    // For rules that produce violations where each violation is associated with a single code location.
+    Standard= "Standard",
+
+    // For rules that produce violations where each violation is associated with multiple code locations.
+    MultiLocation = "MultiLocation",
```
stephen-carter-at-sf, line 14: Since I am touching the engine-api... before the end of the month we'll need to increment all the package versions. So I'll do that separately - at the time of us being ready to integrate the CPD engine with the CLI later this month.

packages/code-analyzer-pmd-engine/test/cpd-engine.test.ts
```diff
@@ -1,19 +1,56 @@
...
+
+        const engine: CpdEngine = new CpdEngine();
+        const ruleDescriptions: RuleDescription[] = await engine.describeRules({});
+
+        await expectRulesToMatchGoldFile(ruleDescriptions, 'rules_allLanguages.goldfile.json');
+    });
+
+    it('When using defaults with workspace that only apex code, then only apex rule is returned', async () => {
```
jfeingold35, line 28: Nitpick: "That _contains_ only apex code"?

packages/code-analyzer-pmd-engine/src/pmd-engine.ts
jfeingold35: Good call to have the rules present as being named "X-javascript". Having the more specific name is going to cut down on confusion later.

packages/code-analyzer-pmd-engine/src/config.ts
jfeingold35: Do PMD's ECMAScript rules even work against typescript? If they're JS-only, then we should absolutely just alias them as JS.

jfeingold35 (approved):

packages/code-analyzer-pmd-engine/src/config.ts
stephen-carter-at-sf: That's correct... Typescript isn't even supported by PMD... it is only supported by CPD and has a separate language model. I'm not sure why PMD uses ecmascript instead of javascript as their identifier.

packages/code-analyzer-pmd-engine/test/cpd-engine.test.ts
stephen-carter-at-sf: Good catch... will fix in my next PR.

PR #114, opened 2024-10-31
FIX(regex): @W-17108343@: Improve performance of regex rules

stephen-carter-at-sf:


packages/code-analyzer-regex-engine/src/engine.ts
```diff
@@ -122,10 +122,17 @@ export class RegexEngine extends Engine {
...
-            const startLine: number = getLineNumber(match.index, newlineIndexes);
-            const startColumn: number = getColumnNumber(match.index, newlineIndexes);
-            const endLine: number = getLineNumber(match.index + match[0].length, newlineIndexes);
-            const endColumn: number = getColumnNumber(match.index + match[0].length, newlineIndexes);
+            let startIndex: number = match.index;
+            let matchLength: number = match[0].length;
+            if (match.groups?.target) {
+                startIndex = startIndex + match[0].indexOf(match.groups.target);
+                matchLength = match.groups.target.length;
+            }
```
stephen-carter-at-sf, line 133: The secret sauce

packages/code-analyzer-regex-engine/test/engine.test.ts
```diff
@@ -141,11 +141,10 @@ describe("Tests for RegexEngine's getName and describeRules methods", () => {
                 path.resolve(__dirname, 'test-data', 'sampleWorkspace', 'dummy3.js')
             ])});
 
-        expect(rulesDescriptions).toHaveLength(4);
+        expect(rulesDescriptions).toHaveLength(3);
         expect(rulesDescriptions[0]).toMatchObject(EXPECTED_AvoidTermsWithImplicitBias_RULE_DESCRIPTION);
-        expect(rulesDescriptions[1]).toMatchObject(EXPECTED_NoGetHeapSizeInLoop_RULE_DESCRIPTION);
```
stephen-carter-at-sf, line 146: This shouldn't have been there since this isn't a rule specific to js files. With my update to limit the rule to apex files - I had to update this test.

packages/code-analyzer-regex-engine/src/engine.ts
jag-j: Nit pick: Can the first section be extracted into a method, say "fetchAdjustedBoundary" or something similar. Will help with readability. Looks great otherwise!

jag-j (approved):

packages/code-analyzer-regex-engine/src/engine.ts
stephen-carter-at-sf: Added a comment. The startIndex and matchLength are inline - so extracting to separate method wouldn't be so simple. I hope comment makes it clear.

PR #117, opened 2024-11-07
NEW @W-17100356@ Add sarif formatter to output formats

jag-j:


packages/code-analyzer-core/test/test-data/expectedOutputFiles/multipleViolations.goldfile.sarif
```diff
@@ -0,0 +1,230 @@
...
+                "region": {
+                  "startLine": 90,
+                  "startColumn": 1,
+                  "endLine": 95,
+                  "endColumn": 10
+                }
+              }
+            }
+          ],
+          "level": "warning"
```
jfeingold35, line 291: What's the deal with this `level` property? Are the only values `warning` and `error`?

jag-j: yes, there is only warning and error. So decided to go with 2 and below as error and others as warning.

packages/code-analyzer-core/package.json
```diff
@@ -16,6 +16,7 @@
     "@salesforce/code-analyzer-engine-api": "0.12.0",
     "@types/js-yaml": "^4.0.9",
     "@types/node": "^20.0.0",
+    "@types/sarif": "^2.1.4",
```
stephen-carter-at-sf: Any reason why you chose 2.1.4 instead of 2.1.7?

packages/code-analyzer-core/src/output-format.ts
```diff
@@ -6,12 +6,14 @@ import * as xmlbuilder from "xmlbuilder";
 import * as fs from 'fs';
 import path from "node:path";
 import {Clock, RealClock} from "./utils";
+import { Log, Run, Result, ReportingDescriptor, Location, Region, Notification } from "sarif";
```
stephen-carter-at-sf: These names are very common. So either let's move the SarifOutputFormatter into its own file... or ... just import this as import * as sarif from "sarif" so that you can then refer to these common names as sarif.Log, sarif.Run, etc.

stephen-carter-at-sf (changes requested):

packages/code-analyzer-core/src/output-format.ts
```diff
@@ -231,6 +235,112 @@ class XmlOutputFormatter implements OutputFormatter {
...
+                    helpUri: ''
+                };
+                if (v.resources) {
+                    rule['helpUri'] = v.resources[0];
+                }
+                rules.push(rule);
+            }
+        }
+
+		return rules;
```
stephen-carter-at-sf: Please remove tab characters from your files... only spaces please for this repo.

packages/code-analyzer-core/test/test-data/expectedOutputFiles/multipleViolations.goldfile.sarif
```diff
@@ -0,0 +1,230 @@
...
+                ],
+                "severity": 3
+              }
+            }
+          ]
+        }
+      },
+      "results": [
+        {
+          "ruleId": "stub3RuleA",
```
stephen-carter-at-sf, line 232: This rule is of type "Flow" and it has multiple locations which make up the path.

Soon we'll also have type "MultiLocation" for CPD results which have multiple locations. 

So shouldn't we be populating the result with multiple locations in these cases?

packages/code-analyzer-core/package.json
jag-j: Since 4.x uses 2.1.4, I used the same so this doesn't bring in any variation during comparison.

packages/code-analyzer-core/src/output-format.ts
jag-j: Good point. I will do this.

packages/code-analyzer-core/src/output-format.ts
```diff
@@ -263,24 +264,46 @@ class SarifOutputFormatter implements OutputFormatter {
...
+                                    startColumn: violationLocation.getColumn(),
+                                    endLine: violationLocation.getEndLine(),
+                                    endColumn: violationLocation.getEndColumn(),
+                                } as sarif.Region,
+                            },
+                        };
+                        relatedLocations.push(relatedLocation);
+                        locIndex++;
+                    }   
+                });
```
stephen-carter-at-sf: This isn't what I was expecting.

I thought we agreed that:
* Only if the data type is MultiLocation, Flow, and DataFlow (which can be determined with a new subfunction - to also help update the other locations which have things like
```
        primaryLocationIndex: [RuleType.DataFlow, RuleType.Flow].includes(rule.getType()) ? violation.getPrimaryLocationIndex() : undefined,
        locations: [RuleType.DataFlow, RuleType.Flow].includes(rule.getType()) ? createCodeLocationOutputs(codeLocations, runDir) : undefined,
```
) would we populate the relatedLocations
* And 2nd... we would all of the locations - including the primary... using the relatedLocations as the full path or full set of locations. 

That is, I thought the goal was to use the locations field to just hold the primary location... and then for types MultiLocation, Flow, and DataFlow to have the relatedLocations hold the all the locations (which represent the path, etc). This is essentially how the json, xml, etc format works - so I was hoping to be consistent.

stephen-carter-at-sf (changes requested):

packages/code-analyzer-core/src/output-format.ts
```diff
@@ -231,6 +235,132 @@ class XmlOutputFormatter implements OutputFormatter {
...
+                        region: {
+                            startLine: violation.line,
+                            startColumn: violation.column,
+                            endLine: violation.endLine,
+                            endColumn: violation.endColumn
+                        } as sarif.Region
+                    }
+                };
+                const relatedLocations:sarif.Location[] = [];
+                if(violation.primaryLocationIndex && violation.locations) {
```
stephen-carter-at-sf: You should be using `typeSupportsMultipleLocations(violation.getRule().getType())` here

where violation isn't the ViolationOutput data structure... but the actual Violation data structure that core provides from the results (which you should be getting from `runResults.getEngineRunResults(engineName)`.

packages/code-analyzer-core/src/output-format.ts
```diff
@@ -231,6 +235,132 @@ class XmlOutputFormatter implements OutputFormatter {
...
+        const resultsOutput: ResultsOutput = toResultsOutput(results);
+        const resultsByEngine: Map<string, ViolationOutput[]> = new Map<string, ViolationOutput[]>();
+
+        for (const engine of results.getEngineNames()) {
+            resultsByEngine.set(engine, []);
+        }
+
+        for (const violation of resultsOutput.violations) {
+            resultsByEngine.get(violation.engine)?.push(violation);
+        }
```
stephen-carter-at-sf: Oh now I see why you are having difficulties. It looks like you are using the ViolationOutput data structure which was the data structure used to help generate json and xml.

You should be just iterating over the results.geEngineNames() and then using runResults.getEngineRunResults(engineName).

stephen-carter-at-sf (changes requested):

packages/code-analyzer-core/src/output-format.ts
```diff
@@ -293,13 +423,17 @@ function createViolationOutput(violation: Violation, runDir: string, sanitizeFcn
...
+        primaryLocationIndex: typeSupportsMultipleLocations(rule) ? violation.getPrimaryLocationIndex() : undefined,
+        locations: typeSupportsMultipleLocations(rule) ? createCodeLocationOutputs(codeLocations, runDir) : undefined,
         message: sanitizeFcn(violation.getMessage()),
         resources: violation.getResourceUrls()
     };
 }
 
+function typeSupportsMultipleLocations(rule: Rule) {
+    return [RuleType.DataFlow, RuleType.Flow, RuleType.MultiLocation].includes(rule.getType());
+}
```
stephen-carter-at-sf: Maybe just pass in the RuleType instead of the rule.

packages/code-analyzer-core/src/output-format.ts
```diff
@@ -231,6 +235,131 @@ class XmlOutputFormatter implements OutputFormatter {
...
+    format(results: RunResults): string {
+        const resultsByEngine: Map<string, Violation[]> = new Map<string, Violation[]>();
+
+        for (const engine of results.getEngineNames()) {
+            resultsByEngine.set(engine, []);
+        }
+
+        for (const violation of results.getViolations()) {
+            resultsByEngine.get(violation.getRule().getEngineName())?.push(violation);
+        }
```
stephen-carter-at-sf: This isn't needed. The results are already available per engine via `results.getEngineRunResults(engineName)`

So just loop over the engine names and then in the loop get the results for that engine.

packages/code-analyzer-core/src/output-format.ts
```diff
@@ -231,6 +235,131 @@ class XmlOutputFormatter implements OutputFormatter {
...
+                        region: {
+                            startLine: primaryViolation.getStartLine(),
+                            startColumn: primaryViolation.getStartColumn(),
+                            endLine: primaryViolation.getEndLine(),
+                            endColumn: primaryViolation.getEndColumn()
+                        } as sarif.Region
+                    }
+                };
+                const relatedLocations:sarif.Location[] = [];
+                if(typeSupportsMultipleLocations(violation.getRule().getType())) {
```
stephen-carter-at-sf: The cyclometic complexity is quite high. 

One trick to address this, is to separate each transformation into its smallest unit.

The resulting code could look like:

```
class SarifOutputFormatter implements OutputFormatter {
        const runDir = results.getRunDirectory();

        const sarifRuns: sarif.Run[] = results.getEngineNames()
            .map(engineName => results.getEngineRunResults(engineName))
            .filter(engineRunResults => engineRunResults.getViolationCount() > 0)
            .map(engineRunResults => toSarifRun(engineRunResults, runDir));

        // Construct SARIF log
        const sarifLog: sarif.Log = {
            version: "2.1.0",
            $schema: 'http://json.schemastore.org/sarif-2.1.0',
            runs: sarifRuns,
        };

        // Return formatted SARIF JSON string
        return JSON.stringify(sarifLog, null, 2);
    }
}

function toSarifRun(engineRunResults: EngineRunResults, runDir: string): sarif.Run {
    const violations: Violation[] = engineRunResults.getViolations();
    const rules: Rule[] = [... new Set(violations.map(v => v.getRule()))]; 
    const ruleNames: string[] = rules.map(r => r.getName());

    return {
        tool: {
            driver: {
                name: engineRunResults.getEngineName(),
                informationUri: "https://developer.salesforce.com/docs/platform/salesforce-code-analyzer/guide/version-5.html",
                rules: rules.map(toSarifReportingDescriptor),
            }
        },
        results: violations.map(v => toSarifResult(v, ruleNames.indexOf(v.getRule().getName()))),
        invocations: [
            {
                executionSuccessful: true,
                workingDirectory: {
                    uri: runDir,
                },
            },
        ],
    };
}

function toSarifResult(violation: Violation, ruleIndex: number) : sarif.Result {
    const primaryCodeLocation = violation.getCodeLocations()[violation.getPrimaryLocationIndex()];
    const result: sarif.Result = {
        ruleId: violation.getRule().getName(),
        ruleIndex: ruleIndex,
        message: { text: violation.getMessage() },
        locations: [toSarifLocation(primaryCodeLocation)],
        level: toSarifNotificationLevel(violation.getRule().getSeverityLevel()),
    };
    if(typeSupportsMultipleLocations(violation.getRule().getType())) {
        result.relatedLocations = violation.getCodeLocations().map(toSarifLocation);
    }
    return result;
}

function toSarifLocation(codeLocation: CodeLocation): sarif.Location {
    return {
        physicalLocation: {
            artifactLocation: {
                uri: codeLocation.getFile(),
            },
            region: {
                startLine: codeLocation.getStartLine(),
                startColumn: codeLocation.getStartColumn(),
                endLine: codeLocation.getEndLine(),
                endColumn: codeLocation.getEndColumn()
            } as sarif.Region
        }
    }
}

function toSarifReportingDescriptor(rule: Rule): sarif.ReportingDescriptor {
    return sarif.ReportingDescriptor = {
        id: rule.getName(),
        properties: {
            category: rule.getTags(),
            severity: rule.getSeverityLevel()
        },
        helpUri: rule.getResourceUrls().length > 0 ? rule.getResourceUrls()[0] : '';
    }
}

function toSarifNotificationLevel(severity: SeverityLevel): sarif.Notification.level {
    return severity < 3 ? 'error' : 'warning'; // IF satif.Notification.level is an enum then please return the num instead of the string.
}
```

stephen-carter-at-sf (approved):

packages/code-analyzer-core/src/output-format.ts
```diff
@@ -231,6 +235,130 @@ class XmlOutputFormatter implements OutputFormatter {
...
+    private groupViolationsByEngine(results: RunResults): Map<string, Violation[]> {
+        const resultsByEngine = new Map<string, Violation[]>();
+        for (const engine of results.getEngineNames()) {
+            resultsByEngine.set(engine, []);
+        }
+        for (const violation of results.getViolations()) {
+            resultsByEngine.get(violation.getRule().getEngineName())?.push(violation);
+        }
+        return resultsByEngine;
+    }
```
stephen-carter-at-sf: Again this isn't needed. The violations are already grouped by engine with results.getEngineRunResults(engineName)

stephen-carter-at-sf (approved):

PR #118, opened 2024-11-07
NEW(cpd): @W-16866828@: Implemented the runRules method for CPD (with no configuration yet)

stephen-carter-at-sf:
In this PR:
* I refactored to give the old pmd-wrapper a new name (pmd-cpd-wrappers) so that it didn't feel so PMD specific.
* I refactored the cpd wrapper to output a data structure that contains ProcessingError instances (just like PMD does)
* I did a small refactor of the tests - moving sampleWorkspace to samplePmdWorkspace
* I then implemented and tested the runRules method on CPD engine.

Note that no configuration has been added yet - so the java command, minumumTokens, etc are all just hard coded for now.


packages/code-analyzer-pmd-engine/test/cpd-engine.test.ts
```diff
@@ -44,9 +44,120 @@ async function expectRulesToMatchGoldFile(actualRuleDescriptions: RuleDescriptio
...
 
 describe('Tests for the runRules method of CpdEngine', () => {
-    it('TEMPORARY TEST FOR CODE COVERAGE', async () => {
-        // Will delete this test as soon as we implement the CpdEngine.
+    it('When zero rules names are provided then return zero violations', async () => {
         const engine: CpdEngine = new CpdEngine();
         expect(await engine.runRules([],{workspace: new Workspace([__dirname])})).toEqual({violations: []});
     });
+
+    it('When rule name is not associated with a language that CPD knows about', async () => {
```
jfeingold35: This is missing the "then" part of the "when...then...".

stephen-carter-at-sf: Fixed. Thanks.

packages/code-analyzer-pmd-engine/pmd-cpd-wrappers/src/main/java/com/salesforce/sfca/cpdwrapper/CpdLanguageRunResults.java
```diff
@@ -0,0 +1,34 @@
...
+        public int numNonemptyLinesInBlock;
+        public int numBlocks;
+        public List<BlockLocation> blockLocations = new ArrayList<>();
+
+        public static class BlockLocation {
+            public String file;
+            public int startLine;
+            public int startCol;
+            public int endLine;
+            public int endCol;
```
jag-j, line 25: I remember seeing a comment field on the typescript side. Do you want to be consistent with the datastructures on the java and typescript side?

packages/code-analyzer-pmd-engine/test/test-data/sampleCpdWorkspace/sampleJavascript2.js
```diff
@@ -0,0 +1,10 @@
+function abc3() {
```
jag-j: Do you have any non-html files with minimum token greater than the default? Also, can we rename these files a bit more descriptive? Like sampleJavascriptWithLessThanMinTokens.js ?

jag-j: Ok, I do see some happy path Apex Classes. So, never mind about my first comment. Can we still have the filenames be more descriptive?

packages/code-analyzer-pmd-engine/pmd-cpd-wrappers/src/main/java/com/salesforce/sfca/cpdwrapper/CpdLanguageRunResults.java
jag-j: Based on the conversation on slack, this is a flowtest only thing. So resolving this.

stephen-carter-at-sf: The comment field isn't something the CPD engine contains. 
The goal of the wrapper classes of PMD and CPD is to simply call through to java to let PMD/CPD do their thing. The PMD world already has their predefined output data structure in json... but CPD only has xml. So I created this datastructure to contain the data that CPD does have using the language that they essentially use. It is on the typscript side that the transformation from PMD or CPD specific output data structures into code analyzer results data structure takes place.

packages/code-analyzer-pmd-engine/test/test-data/sampleCpdWorkspace/sampleJavascript2.js
stephen-carter-at-sf: Yeah, let me see if I can make the files more descriptive...

stephen-carter-at-sf: Done.

jag-j (approved):

PR #123, opened 2024-11-12
NEW(cpd): @W-16866836@: Add java_command and rule_langage configurability to cpd engine

stephen-carter-at-sf:
In this review:
* Added java_command and rule_languages fields to the cpd configuration.
* The java code updates are to add in some better handling of exceptions that CPD could throw so that we can continue processing other languages even if one language fails.
* Removed 'java' language support since John (our PM) agrees that we don't need it for CA5.
* Made cpd rules 'Recommended' by default since it is super easy to turn them off - either disable engine, remove rule_languages, or just remove 'Recommended' tags from the cpd rules.


packages/code-analyzer-pmd-engine/pmd-cpd-wrappers/src/main/java/com/salesforce/sfca/cpdwrapper/CpdRunner.java
```diff
@@ -105,6 +110,16 @@ private CpdLanguageRunResults runLanguage(String language, List<Path> pathsToSca
                     languageRunResults.matches.add(cpdMatch);
                 }
             });
+
+            // Instead of throwing exceptions and causing the entire run to fail, instead we report exceptions as
+            // if they are processing errors so that they can better be handled on the typescript side
+            for (Exception ex : errorListener.exceptionsCaught) {
```
stephen-carter-at-sf, line 116: I noticed a bug in CPD and so I created https://github.com/pmd/pmd/issues/5322. This bug actually prevented the engine from finishing across all languages since we just had java throw an exception which causes the engine to report the exception as a unexpected exception based violation.

But to improve the experience, so that CPD errors don't prevent us from finishing with other languages, I went ahead here and made the exceptions (messages + stack traces) show up as a special kind of processing error which then appears an ErrorEvent back to the CLI so that things continue and the error still gets reported.

This was a fix as you go item as I was self-QAing my own changes here.

packages/code-analyzer-pmd-engine/src/config.ts
```diff
@@ -4,37 +4,42 @@ import {JavaVersionIdentifier} from "./JavaVersionIdentifier";
...
-
-export const CPD_AVAILABLE_LANGUAGES: string[] = ALL_LANGUAGES;
+const DEFAULT_JAVA_COMMAND: string = 'java';
 
 export const PMD_ENGINE_CONFIG_DESCRIPTION: ConfigDescription = {
     overview: getMessage('PmdConfigOverview'),
     fieldDescriptions: {
-        java_command: getMessage('PmdConfigFieldDescription_java_command'),
-        rule_languages: getMessage('PmdConfigFieldDescription_rule_languages', PMD_AVAILABLE_LANGUAGES_TEXT),
+        java_command: getMessage('SharedConfigFieldDescription_java_command', PMD_ENGINE_NAME),
```
stephen-carter-at-sf, line 18: I first tried to use PmdEngine.NAME and CpdEngine.NAME but the compiler had issues with the order of things being loaded (since then config.ts would depend on pmd-engine.ts and vice versa). 

So to fix this, I just moved the names of these engines into the constants.ts file.

packages/code-analyzer-pmd-engine/src/config.ts
```diff
@@ -98,11 +138,56 @@ class PmdConfigValueExtractor {
         return javaCommand;
     }
 
+    private async attemptToAutoDetectJavaCommand(): Promise<string> {
```
stephen-carter-at-sf, line 141: Nothing new here... just moved this to a shared abstract class so that it could be reused for both cpd and pmd.

packages/code-analyzer-pmd-engine/src/cpd-engine.ts
```diff
@@ -124,7 +126,7 @@ function createRuleForLanguage(languageId: LanguageId): RuleDescription {
         name: getRuleNameFromLanguage(languageId),
         severityLevel: SeverityLevel.Info,
         type: RuleType.MultiLocation,
-        tags: [`${languageId}Language`],
+        tags: ['Recommended', `${languageId}Language`],
```
stephen-carter-at-sf, line 129: Note: that we'll confirm this with John before going GA... but I am leaning towards Josh's preference here to just make these turned on by default with the 'Recommended' flag. Makes sense to me.

packages/code-analyzer-pmd-engine/test/plugin.test.ts
```diff
@@ -76,10 +87,10 @@ describe('Tests for the PmdCpdEnginesPlugin', () => {
         }
     });
 
-    it(`When createEngineConfig for 'pmd' is given an empty raw config and if our attempt to look-up java on the users machine produces old java version, then error`, async () => {
+    it.each(['cpd','pmd'])(`When createEngineConfig for '%s' is given an empty raw config and if our attempt to look-up java on the users machine produces old java version, then error`, async (engineName) => {
```
stephen-carter-at-sf, line 90: Reusing many tests using it.each - since it is the Abstract class that has most of the logic behind what is being tested with the java_command and rule_languages fields for both pmd and cpd.

packages/code-analyzer-pmd-engine/test/test-data/cpdGoldfiles/rules_allDefaultLanguages.goldfile.json
```diff
@@ -16,30 +17,20 @@
...
+      "Recommended",
       "htmlLanguage"
     ],
     "description": "Identify duplicate code blocks within your workspace files associated with the 'html' language.",
     "resourceUrls": [
       "https://docs.pmd-code.org/latest/pmd_userdocs_cpd.html#refactoring-duplicates"
     ]
   },
-  {
-    "name": "DetectCopyPasteForJava",
```
stephen-carter-at-sf, line 27: We might remove java support alltogether (still waiting to confirm with John). But for now, although still available, it isn't one of the default languages anymore.

jfeingold35 (approved):

PR #125, opened 2024-11-13
NEW @W-17042378@ Supplemental PR with HTML updates

jfeingold35:


packages/code-analyzer-core/package.json
```diff
@@ -13,7 +13,7 @@
   "main": "dist/index.js",
   "types": "dist/index.d.ts",
   "dependencies": {
-    "@salesforce/code-analyzer-engine-api": "0.13.0",
+    "@salesforce/code-analyzer-engine-api": "0.13.1",
```
stephen-carter-at-sf: We don't need to update the engine api. Only thing that needs to be updated is code-analyzer-core to 0.16.1

So this PR should have only changes in html-template-0.0.#.txt (<-- in which I would hope we can have marcelino update his version numbers each time he is finished making changes.) and the one package.json file only.

stephen-carter-at-sf (changes requested):

stephen-carter-at-sf (approved):

PR #126, opened 2024-11-13
NEW(cpd): @W-16866916@: Add minimum_tokens and skip_duplicate_files to cpd engine config

stephen-carter-at-sf:


packages/code-analyzer-pmd-engine/test/cpd-engine.test.ts
```diff
@@ -218,4 +218,96 @@ describe('Tests for the runRules method of CpdEngine', () => {
         expect(progressEvents.map(e => e.percentComplete)).toEqual([2, 5, 9.65, 14.3, 17.4, 20.5,
             26.7, 32.9, 39.1, 42.2, 45.3, 51.5, 57.7, 63.9, 67, 70.1, 76.3, 82.5, 88.7, 93.35, 98, 100]);
     });
+
+    it('When specifying a minimum_tokens length that is small enough to pick up smaller code blocks, then violations are returned', async () => {
```
jag-j, line 225: Is there a negative test case for this? i.e. minimum_tokens large enough nothing gets picked?

stephen-carter-at-sf: Yes... notice the above test (uses default 100 tokens) doesn't pick up the javascript files. Only when we reduce down to lower minimum_tokens amount does it get picked up. See line 135 which includes DetectCopyPasteForJavascript. I'll add a comment though to make this more clear.

packages/code-analyzer-pmd-engine/src/messages.ts
```diff
@@ -38,6 +38,15 @@ const MESSAGE_CATALOG : { [key: string]: string } = {
         `List of languages associated with CPD to be made available for 'cpd' engine rule selection.\n` +
         `The languages that you may choose from are: %s.`,
 
+    CpdConfigFieldDescription_minimum_tokens:
```
jfeingold35, line 41: These lines feel a little long. I'm not sure if there's anything to be done about that, though.

packages/code-analyzer-pmd-engine/src/messages.ts
```diff
@@ -38,6 +38,15 @@ const MESSAGE_CATALOG : { [key: string]: string } = {
...
+        `The minimum number of tokens required to be in a duplicate block of code in order to be reported as a violation.\n` +
+        `The concept of a token may be defined differently per language, but in general it a distinct basic element of source code.\n` +
+        `For example, this could be language specific keywords, identifiers, operators, literals, and more.\n` +
+        `See https://docs.pmd-code.org/latest/pmd_userdocs_cpd.html to learn more.`,
+
+    CpdConfigFieldDescription_skip_duplicate_files:
+        `Indicates whether to ignore multiple copies of files of the same name and length.`,
```
jfeingold35, line 48: "And length" is confusing. Should it be somethign like "files with the same name and contents"?

packages/code-analyzer-pmd-engine/test/cpd-engine.test.ts
stephen-carter-at-sf: OK -- added a comment in the previous test to make this clear.

packages/code-analyzer-pmd-engine/test/plugin.test.ts
```diff
@@ -294,6 +296,41 @@ describe('Tests for the PmdCpdEnginesPlugin', () => {
...
     });
 
+    it(`When createEngineConfig for 'cpd' is given a minimum_tokens value that is not a number, then error`, async () => {
+        const rawConfig: ConfigObject = {minimum_tokens: 'oops'};
+        const configValueExtractor: ConfigValueExtractor = new ConfigValueExtractor(rawConfig, 'engines.cpd');
+        await expect(plugin.createEngineConfig('cpd', configValueExtractor)).rejects.toThrow(
+            getMessageFromCatalog(SHARED_MESSAGE_CATALOG, 'ConfigValueMustBeOfType', 'engines.cpd.minimum_tokens', 'number', 'string'));
+    });
+
+    it.each([-5,0,2.5])(`When createEngineConfig for 'cpd' is given a minumum_tokens number that is %s, then error`, async (invalidValue) => {
```
jfeingold35: Quibble: Could the test name explain why the values are invalid?
Maybe `{value: -5, reason: 'is negative'})('When value for minimum_tokens $reason, then error')`?

packages/code-analyzer-pmd-engine/test/plugin.test.ts
```diff
@@ -294,6 +296,41 @@ describe('Tests for the PmdCpdEnginesPlugin', () => {
...
+    });
+
+    it(`When createEngineConfig for 'cpd' is given a skip_duplicate_files value that is not a boolean, then error`, async () => {
+        const rawConfig: ConfigObject = {skip_duplicate_files: 3};
+        const configValueExtractor: ConfigValueExtractor = new ConfigValueExtractor(rawConfig, 'engines.cpd');
+        await expect(plugin.createEngineConfig('cpd', configValueExtractor)).rejects.toThrow(
+            getMessageFromCatalog(SHARED_MESSAGE_CATALOG, 'ConfigValueMustBeOfType', 'engines.cpd.skip_duplicate_files', 'boolean', 'number'));
+    });
+
+    it(`When createEngineConfig for 'cpd' is given a valid minimum_tokens value, then it is used`, async() => {
```
jfeingold35: Should be `skip_duplicate_files` in name, not `minimum_tokens`, right?

jfeingold35 (changes requested):

packages/code-analyzer-pmd-engine/src/messages.ts
stephen-carter-at-sf: This is the same wording as in CPD... so I wanted to reflect what they say. See:
https://docs.pmd-code.org/latest/pmd_userdocs_cpd.html 
where it says

--skip-duplicate-files | Ignore multiple copies of files of the same name and length in comparison.

But yeah all these should go through doc review of course. Still waiting on all of core descriptions to be reviewed by Juliet.

packages/code-analyzer-pmd-engine/src/messages.ts
stephen-carter-at-sf: It is less than the PmdConfigFieldDescription_custom_rulesets for example. So I think it is good. But if we need to later do a full review of all descriptions - then yeah we can shorten things at that time of doc review.

packages/code-analyzer-pmd-engine/test/plugin.test.ts
stephen-carter-at-sf: Sure...

packages/code-analyzer-pmd-engine/test/plugin.test.ts
stephen-carter-at-sf: Good catch...

stephen-carter-at-sf: Fixed.

packages/code-analyzer-pmd-engine/test/plugin.test.ts
stephen-carter-at-sf: Fixed.

packages/code-analyzer-pmd-engine/src/messages.ts
stephen-carter-at-sf: Yeah I checked https://github.com/pmd/pmd/blob/3659a2cb7d5f65d689daf45e861e593ad24d34a5/pmd-core/src/main/java/net/sourceforge/pmd/internal/util/FileCollectionUtil.java#L57 and it appears that they simply just omit adding a file if another file with the same length and file size was added. They do so by creating a new file signature based off of 
```
String signature = path.getFileName() + "_" + Files.size(path);
```
so I think in order to be technically correct, we need to stick with this "same name and length" phrase ... or maybe reword it to "same name and file size" if we think that'll help.

jfeingold35 (approved):

PR #135, opened 2024-11-20
CHANGE @W-16891041@ FlowTest rules consolidated

jfeingold35:


packages/code-analyzer-flowtest-engine/src/engine.ts
```diff
@@ -51,9 +51,9 @@ export class FlowTestEngine extends Engine {
                 return [];
             }
         }
-        const flowTestRules: FlowTestRuleDescriptor[] = await this.commandWrapper.getFlowTestRuleDescriptions();
         this.emitDescribeRulesProgressEvent(75);
-        const convertedRules = flowTestRules.map(toRuleDescription);
+        const consolidatedNames: Set<string> = new Set(getConsolidatedRuleNames());
+        const convertedRules: RuleDescription[] = [...consolidatedNames.values()].map(getConsolidatedRuleByName);
```
stephen-carter-at-sf: Why use a set? Why not just do `getConsolidatedRuleNames().map(getConsolidatedRuleByName)`?

packages/code-analyzer-flowtest-engine/src/engine.ts
```diff
@@ -15,6 +14,7 @@ import {
     FlowTestExecutionResult,
     FlowTestRuleDescriptor
```
stephen-carter-at-sf: Is this needed?

packages/code-analyzer-flowtest-engine/src/engine.ts
stephen-carter-at-sf, line 1: It looks like we have tabs mixed with spaces. Can you make it all spaces?

jfeingold35: Where am I using tabs?

packages/code-analyzer-flowtest-engine/src/hardcoded-catalog.ts
```diff
@@ -0,0 +1,57 @@
...
+        description: getMessage('ConsolidatedRuleDescription', 'With Sharing'),
+        severityLevel: SeverityLevel.Low,
+        tags: ['Recommended', 'Security'],
+        resourceUrls: []
+    }]
+]);
+
+export function getConsolidatedRuleNames(): string[] {
+    return [...QUERY_NAMES_TO_CONSOLIDATED_NAMES.values()];
+}
```
stephen-carter-at-sf: Oh I see... you use a set in the other spot. I think it should be here to make things unique. So you could here do
`[...new Set(QUERY_NAMES_TO_CONSOLIDATED_NAMES.values())]`

stephen-carter-at-sf (approved):
Consider moving the location where you make the consolidatedRuleNames unique (i.e. apply the set inside of the getConsolidatedRuleNames function instead. 

Otherwise, approve.

PR #137, opened 2024-11-26
CHANGE @W-17311830@ FlowTest no longer makes log files

jfeingold35:


packages/code-analyzer-flowtest-engine/package.json
```diff
@@ -17,6 +17,7 @@
     "@types/node": "^20.0.0",
     "@types/semver": "^7.5.8",
     "@types/tmp": "^0.2.6",
+    "fast-glob": "^3.3.2",
```
stephen-carter-at-sf: This shouldn't be a dependency if it is only used in test. It should be a devDepenency at best. But is there an alternative instead of adding more dependencies to our project?

packages/code-analyzer-flowtest-engine/test/python/FlowTestCommandWrapper.test.ts
```diff
@@ -55,6 +57,11 @@ describe('FlowTestCommandWrapper implementations', () => {
                 it('Correctly parses status updates from stdout', () => {
                     expect(completionPercentages).toEqual([0, 25, 50, 75]);
                 });
+
+                it('Generates no log file', async () => {
+                    const flowLogFiles = await fg.glob('./**/.flowtest_log_*.log');
+                    expect(flowLogFiles).toHaveLength(0);
+                });
```
stephen-carter-at-sf: Seems heavy handed to pull in all of fast-glob just too look up if a .log file was generated somewhere underneath the current working directory. I wonder if there is some other node built in tool that we could use instead. If not then I wonder if the test is even needed or worth it.

packages/code-analyzer-flowtest-engine/package.json
jfeingold35: Huh, I was positive I had it as a `devDependency` already. Guess not.
I'll see if there's an alternative, but the problem is that the logfile's name is randomly generated, so we don't already know where it is.
I'll see what the other options are to programmatically find a file.

packages/code-analyzer-flowtest-engine/test/python/FlowTestCommandWrapper.test.ts
jfeingold35: The test is worth it IMO because it locks in the behavior of not generating the log file.

packages/code-analyzer-flowtest-engine/test/python/FlowTestCommandWrapper.test.ts
```diff
@@ -55,6 +55,12 @@ describe('FlowTestCommandWrapper implementations', () => {
                 it('Correctly parses status updates from stdout', () => {
                     expect(completionPercentages).toEqual([0, 25, 50, 75]);
                 });
+
+                it('Generates no log file', async () => {
+                    const logFileMatcher = /\.flowtest_log_.+\.log/;
+                    const logFiles = (await fs.readdir('.')).filter(f => f.match(logFileMatcher));
+                    expect(logFiles).toHaveLength(0);
+                });
```
stephen-carter-at-sf, line 63: This looks good now. Thanks for removing the glob dependency.

stephen-carter-at-sf (approved):

PR #143, opened 2024-12-03
NEW: @W-17293079@: Update ConfigDescription objects to include type and default value information

stephen-carter-at-sf:
This work will allow us to then update our cli client making it far easier and more robust to detect default values, types, etc.

### DIFF
--- a/packages/code-analyzer-core/src/code-analyzer.ts
+++ b/packages/code-analyzer-core/src/code-analyzer.ts
@@ -238,15 +238,16 @@ export class CodeAnalyzer {
             return;
         }
 
+        const engineOverrides: EngineOverrides = this.config.getEngineOverridesFor(engineName);
+
         try {
             const engineConfigDescription: engApi.ConfigDescription = enginePluginV1.describeEngineConfig(engineName);
-            normalizeEngineConfigDescription(engineConfigDescription, engineName);
-            this.engineConfigDescriptions.set(engineName, engineConfigDescription);
+            const configDescription: ConfigDescription = normalizeEngineConfigDescription(engineConfigDescription, engineName, engineOverrides);
+            this.engineConfigDescriptions.set(engineName, configDescription);
         } catch (err) {
             throw new Error(getMessage('PluginErrorWhenCreatingEngine', engineName, (err as Error).message));
         }
 
-        const engineOverrides: EngineOverrides = this.config.getEngineOverridesFor(engineName);
         if (engineOverrides[FIELDS.DISABLE_ENGINE]) {
             this.emitLogEvent(LogLevel.Debug, getMessage('EngineDisabled', engineName,
                 `${FIELDS.ENGINES}.${engineName}.${FIELDS.DISABLE_ENGINE}`))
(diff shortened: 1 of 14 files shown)


packages/code-analyzer-engine-api/src/config.ts
```diff
@@ -18,8 +18,26 @@ export type ConfigDescription = {
...
+}
+
+export type ConfigFieldDescription = {
+    // Text that describes the field to be documented
+    descriptionText: string
+
+    // The type for the value associated with this field (as a string).
+    //   Note that this as a string instead of an enum in order to give flexibility to describe the type of the field as
+    //   we want it to show up in our public docs, but it is recommended to return one of the following when possible:
+    //     'string', 'number', 'boolean', 'object', 'array'
```
jfeingold35, line 32: Could we provide an enum with these common options and type it as `valueType: TypeNum | string;`? On the one hand, it would make it easier to maintain consistency. On the other, it's another thing that cascades. What we have is fine, I just want to throw an alternative out there.

packages/code-analyzer-core/src/code-analyzer.ts
```diff
@@ -524,15 +525,28 @@ function isIntegerBetween(value: number, leftBound: number, rightBound: number):
     return value >= leftBound && value <= rightBound && Number.isInteger(value);
 }
 
-function normalizeEngineConfigDescription(engineConfigDescription: ConfigDescription, engineName: string): void {
-    // Every engine config should have an overview, so if missing, then we add in a generic one
-    if (!engineConfigDescription.overview) {
-        engineConfigDescription.overview = getMessage('GenericEngineConfigOverview', engineName.toUpperCase());
+function normalizeEngineConfigDescription(engineConfigDescription: engApi.ConfigDescription, engineName: string,
```
jfeingold35: Tiny nitpick: Since this method returns a new object now instead of modifying the provided one, could we rename it `toNormalized...` instead of `normalize...?`

packages/code-analyzer-engine-api/src/config.ts
stephen-carter-at-sf: We can add this in the future if we find it to be valuable. But I'd prefer to not add yet another thing to our engine api if at all possible. Since the only place this field will be used is for our public doc-generate tools right now... then I don't think we need to focus too much of our efforts here right now. Also I like the idea of giving flexibility for us to put in anything we want... not just "array" but "array of strings" if we want to in the future - if we think it makes the doc pages look better.

packages/code-analyzer-core/src/code-analyzer.ts
stephen-carter-at-sf: How about fromEngineConfigDescription or toConfigDescription instead since we are now converting an engApi.ConfigDescription to a ConfigDescription?

packages/code-analyzer-engine-api/src/config.ts
jfeingold35: Reasonable.

packages/code-analyzer-core/src/code-analyzer.ts
stephen-carter-at-sf: Pushed a commit to call this toConfigDescription and added in a comment to make it clear.

jfeingold35 (approved):

PR #144, opened 2024-12-03
NEW @W-17159591@ SARIF output now passes official validation

jfeingold35:


packages/code-analyzer-core/src/output-formats/sarif-output-format.ts
```diff
@@ -60,26 +61,35 @@ function toSarifResult(violation: Violation, ruleIndex: number) : sarif.Result {
...
+                region: {
+                    startLine: codeLocation.getStartLine(),
+                    startColumn: codeLocation.getStartColumn(),
+                    endLine: codeLocation.getEndLine(),
+                    endColumn: codeLocation.getEndColumn()
+                } as sarif.Region
+            }
         }
+    } else {
+        return {
```
jfeingold35, line 88: The validator didn't like having an empty `region` object, so I'm just using an empty `artifactLocation` object outright when there's no code location.
Technically I suppose it's possible we could see a case where there's a code location with no start line or anything else, and the output of that would be invalid SARIF. I can account for that scenario if consensus is that it's worth the trouble.

stephen-carter-at-sf (approved):

PR #145, opened 2024-12-04
CHANGE: @W-17310689@: Standardize our tags across engines and add PMD rule mappings

stephen-carter-at-sf:


packages/code-analyzer-pmd-engine/src/pmd-rule-mappings.ts
stephen-carter-at-sf, line 1: This is one of the main files I added.

packages/code-analyzer-regex-engine/src/plugin.ts
```diff
@@ -72,18 +73,18 @@ export function createBaseRegexRules(now: Date): RegexRules {
     return {
         NoTrailingWhitespace: {
             regex: (/(?<target>[ \t]+)((\r?\n)|$)/g).toString(),
-            file_extensions: ['.cls', '.trigger'],
+            file_extensions: ['.cls', '.trigger'], // Currently restricted to apex files only... but we might want to extend where this rule applies in the future
             description: getMessage('TrailingWhitespaceRuleDescription'),
             violation_message: getMessage('TrailingWhitespaceRuleMessage'),
             severity: SeverityLevel.Info,
-            tags: ['Recommended', 'CodeStyle']
+            tags: [COMMON_TAGS.RECOMMENDED, COMMON_TAGS.CATEGORIES.CODE_STYLE]
```
jfeingold35: Should this not have the `COMMON_TAGS.LANGUAGES.APEX` tag?

packages/code-analyzer-regex-engine/src/config.ts
```diff
@@ -82,7 +82,7 @@ export function validateAndNormalizeConfig(valueExtractor: ConfigValueExtractor)
             violation_message: ruleExtractor.extractString('violation_message',
                 getDefaultRuleViolationMessage(regexString, ruleName, description))!,
             severity: ruleExtractor.extractSeverityLevel('severity', DEFAULT_SEVERITY_LEVEL)!,
-            tags: ruleExtractor.extractArray('tags', ValueValidator.validateString, DEFAULT_TAGS)!,
+            tags: ruleExtractor.extractArray('tags', ValueValidator.validateString, [COMMON_TAGS.RECOMMENDED, COMMON_TAGS.CUSTOM])!,
```
jfeingold35, line 85: A thought occurs: Is there anything that can go wrong if a rule has the same tag multiple times? Asking on the off chance that someone tags their custom rule with `CUSTOM` manually.

packages/code-analyzer-pmd-engine/test/test-data/pmdGoldfiles/rules_visualforceOnly.goldfile.json
```diff
@@ -1,11 +1,11 @@
 [
   {
     "name": "VfCsrf",
-    "severityLevel": 3,
+    "severityLevel": 2,
```
jfeingold35, line 4: Why did the severity of these rules change?

packages/code-analyzer-regex-engine/src/plugin.ts
stephen-carter-at-sf: I'm hesitant to do so because we might open up this regex rule to work on more than apex code in the future. It's easier to add in a tag than remove it. I don't see a reason why this rule needs to be apex specific (even though currently it is).

packages/code-analyzer-regex-engine/src/config.ts
stephen-carter-at-sf: If the user supplies their own tags... they would have to add in "Custom" themselves. We don't add these onto the end. But even if they do add in multiple tags, I believe core automatically uniquifies the list. But even if we didnt make it unique then selection mechanism wouldn't change at all. So no issues here.

packages/code-analyzer-pmd-engine/test/test-data/pmdGoldfiles/rules_visualforceOnly.goldfile.json
stephen-carter-at-sf: See the work item which points (I think in the chatter thread) to the slack thread where John shared all the PMD rule severity and tags.

stephen-carter-at-sf: Again see the pmd-rule-mappings.ts file.

packages/code-analyzer-regex-engine/src/plugin.ts
stephen-carter-at-sf: But yeah - the more I think about it... I guess it is fine if we remove tags in a future release. Or maybe we wouldn't need to even if this rule worked on other file types as well.

I'm just conflicted due to the other rule for implicit bias which works on all text files - not just apex files.

What do you prefer? Add in Apex tag or not? (which might be fine even if we expand it to other file types because it does indeed deal with apex code).

stephen-carter-at-sf: OK... i convinced myself. Added in the `COMMON_TAGS.LANGUAGES.APEX` tag per your request.

jfeingold35 (approved):

PR #146, opened 2024-12-04
NEW @W-17229126@ HTML now includes version info

jfeingold35:


packages/code-analyzer-core/src/output-formats/html-output-format.ts
```diff
@@ -22,19 +22,19 @@ export class HtmlOutputFormatter implements OutputFormatter {
 
     format(results: RunResults): string {
         const runDir = results.getRunDirectory();
-        // It is easiest to put the violations in json format (while escaping html characters) since it is easily
+        // It is easiest to put the results in json format (while escaping html characters) since it is easily
         // consumed by our html template.
-        const jsonViolations: JsonViolationOutput[] = toJsonViolationOutputArray(results.getViolations(), runDir,escapeHtml);
+        const jsonOutput: JsonResultsOutput = toJsonResultsOutput(results, escapeHtml);
```
jfeingold35, line 27: Marcelino's changes to the HTML template rely on the entire object being passed in, as opposed to just a violation array. So that's what we're doing now.

stephen-carter-at-sf: Oh right. Good catch here. Thanks for calling this out.

stephen-carter-at-sf (approved):

jag-j (approved):

PR #151, opened 2024-12-06
CHANGE: @W-17386421@: Hide rule_languages fields from pmd and cpd eng...

stephen-carter-at-sf:
…ine configs in favor of just having them all enabled all the time


packages/code-analyzer-pmd-engine/src/pmd-engine.ts
```diff
@@ -79,6 +79,11 @@ export class PmdEngine extends Engine {
             .map(ruleName => fetchRuleInfoByRuleName(ruleInfoList, ruleName))
             .filter(ruleInfo => ruleInfo !== null);
 
+        if (selectedRuleInfoList.length === 0) {
```
stephen-carter-at-sf, line 82: Small optimization that I noticed while I was in here... doing a quick return when there are zero rules to run.

jfeingold35 (approved):

PR #158, opened 2024-12-10
NEW(pmd): @W-17310907@: Update our build to be able to build/test our own PMD rules

stephen-carter-at-sf:


packages/code-analyzer-pmd-engine/pmd-cpd-wrappers/build.gradle.kts
```diff
@@ -1,8 +1,8 @@
 import java.awt.Desktop
 
 plugins {
-    // Apply the application plugin to add support for building a CLI application in Java.
-    application
+    // plugin to build java code
+    java
```
stephen-carter-at-sf, line 5: We really didn't need to application plugin since we don't need cli wrappers at all. So using java plugin directly turns out to be cleaner.

This also allows us to put jar files in java-libs folder directly underneath the dist folder instead of nested under pmd-cpd-wrappers/lib folder.

jfeingold35 (approved):

PR #159, opened 2024-12-11
NEW @W-17397558@ CodeAnalyzerCore now requires Node v20 or later (1 of 2)

jfeingold35:
This is the first of two PRs that collectively constitute User Story W-17397558.
This PR adds a validation to the Code Analyzer Core constructor, requiring that the current version of Node is at least 20.


packages/code-analyzer-core/src/messages.ts
```diff
@@ -39,6 +39,9 @@ const MESSAGE_CATALOG : MessageCatalog = {
     GenericEngineConfigOverview:
         `%s ENGINE CONFIGURATION`,
 
+    UnsupportedNodeVersion:
+        `Code Analyzer requires Node v%s or later. Current Node version: %s`,
```
stephen-carter-at-sf: Since some users may not know what Node is, can we maybe say something that indicates that they should install/upgrade to 20+. Maybe something like:
```
Code Analyzer must run using Node v%s, but is currently running from and older version of Node version: %s. Please upgrade the version of 'node' on your machine.
```

packages/code-analyzer-core/src/code-analyzer.ts
```diff
@@ -48,7 +51,11 @@ export class CodeAnalyzer {
     private readonly rulesCache: Map<string, RuleImpl[]> = new Map();
     private readonly engineRuleDiscoveryProgressAggregator: EngineProgressAggregator = new EngineProgressAggregator();
 
-    constructor(config: CodeAnalyzerConfig) {
+    constructor(config: CodeAnalyzerConfig, version: string = process.version) {
+        const semver: SemVer = new SemVer(version);
+        if (semver.major < MINIMUM_SUPPORTED_NODE) {
+            throw new Error(getMessage('UnsupportedNodeVersion', MINIMUM_SUPPORTED_NODE, version));
+        }
```
stephen-carter-at-sf, line 63: Nitpick (favor to ask to to satisfy my preference): Just thinking about future expansion. Can we call through to a subfunction from the constructor called validateEnvironment or something... so we can keep the constructor looking as clean as possible and so we know where to add in another validations/assertions in the future.

packages/code-analyzer-core/src/code-analyzer.ts
```diff
@@ -48,7 +51,11 @@ export class CodeAnalyzer {
     private readonly rulesCache: Map<string, RuleImpl[]> = new Map();
     private readonly engineRuleDiscoveryProgressAggregator: EngineProgressAggregator = new EngineProgressAggregator();
 
-    constructor(config: CodeAnalyzerConfig) {
+    constructor(config: CodeAnalyzerConfig, version: string = process.version) {
```
stephen-carter-at-sf, line 54: Ah nice - i forgot about default values. Perfect for testability.

packages/code-analyzer-core/test/code-analyzer.test.ts
```diff
@@ -37,6 +55,7 @@ describe("Tests for the run method of CodeAnalyzer", () => {
     const expectedStubEngine2RuleNames: string[] = ['stub2RuleA', 'stub2RuleC'];
 
     beforeEach(async () => {
+        jest.mock('process')
```
stephen-carter-at-sf: Do we need really this?

jfeingold35: No, that was something I tried early on and forgot to remove. Good catch!

packages/code-analyzer-core/src/code-analyzer.ts
jfeingold35: Reasonable. Stand by.

.github/workflows/verify-pr/validate-changed-package-versions.js
```diff
@@ -60,6 +60,11 @@ function convertFileNameToPackageNameIfPossible(changedFile) {
     }
 }
 
+function isFileInTestFolder(changedFile) {
+    const changedFilePathSegments = path.dirname(changedFile).split('/');
+    return changedFilePathSegments.includes('test');
```
stephen-carter-at-sf: Hmm not sure how I feel about this. What if a piece of source code happens to live in some folder called test in the future. Instead, can we be a little more strict to look for the specific test folder with something like `"packages/<nameOfPackageFolder>/test/"` instead?

packages/code-analyzer-core/src/messages.ts
jfeingold35: `Code Analyzer requires Node v%s or later, and is currently running Node %s. Please upgrade the version of 'node' on your machine`?

.github/workflows/verify-pr/validate-changed-package-versions.js
jfeingold35: My thought was that this way, it also skips Java test files. But I'm fine with a more strict version, and we can loosen it if we find ourselves in the position of needing to do so. Stand by for change.

packages/code-analyzer-core/src/code-analyzer.ts
```diff
@@ -48,10 +51,18 @@ export class CodeAnalyzer {
...
+        this.validateEnviroment(version);
         this.config = config;
     }
 
+    private validateEnviroment(version: string): void {
```
stephen-carter-at-sf: typo Enviroment --> Environment

stephen-carter-at-sf (changes requested):

packages/code-analyzer-core/src/code-analyzer.ts
jfeingold35: Oops. Good catch!

stephen-carter-at-sf (approved):

PR #162, opened 2024-12-12
NEW: @W-16371174@: Make all configuration settings case insensitive

stephen-carter-at-sf:
In addition to making everything case insensitive, I also made all unknown config values that a user puts in their config file produce an error. So now no more cases where issues accidentally have a typo in their config file and it gets ignored.

### DIFF
--- a/packages/code-analyzer-core/src/code-analyzer.ts
+++ b/packages/code-analyzer-core/src/code-analyzer.ts
@@ -273,6 +273,10 @@ export class CodeAnalyzer {
         const engineConfigValueExtractor: engApi.ConfigValueExtractor = new engApi.ConfigValueExtractor(
             engineOverrides as engApi.ConfigObject, `${FIELDS.ENGINES}.${engineName}`, this.config.getConfigRoot());
 
+        // Although not truly a hidden field (it is well documented), we mark 'disable_engine' as a hidden field on the
+        // extractor so that each engine doesn't need to list it when using the validateOnlyContainsKeys method.
+        engineConfigValueExtractor._addHiddenKeys([FIELDS.DISABLE_ENGINE]);
+
         try {
             const engineConfig: engApi.ConfigObject = await enginePluginV1.createEngineConfig(engineName, engineConfigValueExtractor);
             const engine: engApi.Engine = await enginePluginV1.createEngine(engineName, engineConfig);
--- a/packages/code-analyzer-core/src/code-analyzer.ts
+++ b/packages/code-analyzer-core/src/code-analyzer.ts
@@ -273,9 +273,9 @@ export class CodeAnalyzer {
         const engineConfigValueExtractor: engApi.ConfigValueExtractor = new engApi.ConfigValueExtractor(
             engineOverrides as engApi.ConfigObject, `${FIELDS.ENGINES}.${engineName}`, this.config.getConfigRoot());
 
-        // Although not truly a hidden field (it is well documented), we mark 'disable_engine' as a hidden field on the
-        // extractor so that each engine doesn't need to list it when using the validateOnlyContainsKeys method.
-        engineConfigValueExtractor._addHiddenKeys([FIELDS.DISABLE_ENGINE]);
+        // We mark 'disable_engine' as a key that the extractor should not worry about with validation so that each
+        // engine doesn't need to list it when using the validateContainsOnlySpecifiedKeys method.
+        engineConfigValueExtractor.addKeysThatBypassValidation([FIELDS.DISABLE_ENGINE]);
 
         try {
             const engineConfig: engApi.ConfigObject = await enginePluginV1.createEngineConfig(engineName, engineConfigValueExtractor);
--- a/packages/code-analyzer-engine-api/src/config.ts
+++ b/packages/code-analyzer-engine-api/src/config.ts
@@ -44,11 +44,16 @@ export class ConfigValueExtractor {
     private readonly configObj: ConfigObject;
     private readonly fieldPathRoot: string;
     private readonly configRoot: string;
+    private hiddenKeys: string[] = [];
 
     constructor(configObject: ConfigObject, fieldPathRoot: string = '', configRoot: string = process.cwd()) {
         this.configObj = configObject;
         this.fieldPathRoot = fieldPathRoot;
         this.configRoot = configRoot;
+    }
+
+    _addHiddenKeys(hiddenKeys: string[]) {
+        this.hiddenKeys = this.hiddenKeys.concat(hiddenKeys);
     }
 
     getObject(): ConfigObject {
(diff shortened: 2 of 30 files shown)


packages/code-analyzer-core/src/code-analyzer.ts
```diff
@@ -269,6 +269,7 @@ export class CodeAnalyzer {
 
         const engineConfigValueExtractor: engApi.ConfigValueExtractor = new engApi.ConfigValueExtractor(
             engineOverrides as engApi.ConfigObject, `${FIELDS.ENGINES}.${engineName}`, this.config.getConfigRoot());
+        engineConfigValueExtractor._addHiddenKeys([FIELDS.DISABLE_ENGINE]);
```
stephen-carter-at-sf, line 272: Although not truly a hidden field... doing this here allows each engine to use the validateOnlyContainsKeys without having to specify 'disable_engine' in their list. They should only have to care about what they have added to the config.

packages/code-analyzer-engine-api/src/config.ts
```diff
@@ -63,99 +68,135 @@ export class ConfigValueExtractor {
         return this.configRoot;
     }
 
-    extractRequiredBoolean(fieldName: string): boolean {
-        return ValueValidator.validateBoolean(this.configObj[fieldName], this.getFieldPath(fieldName));
+    validateOnlyContainsKeys(keys: string[]) {
```
jfeingold35: I don't love this name (it's weirdly worded IMO), but I can't think of an alternative. `validateContainsOnlySpecifiedKeys()` maybe?

packages/code-analyzer-engine-api/src/config.ts
```diff
@@ -63,99 +68,135 @@ export class ConfigValueExtractor {
...
+        const actualKeys: string[] = this.getKeys();
```
jfeingold35, line 72: Could we name this variable `actualNonHiddenKeys`, just to make it even more clear that hidden keys are excluded from this check?

packages/code-analyzer-engine-api/src/config.ts
```diff
@@ -63,99 +68,135 @@ export class ConfigValueExtractor {
...
+    hasValueDefinedFor(key: string): boolean {
+        const value: ConfigValue = getValueUsingCaseInsensitiveKey(this.configObj, key);
+        return value !== null;
     }
 
-    extractNumber(fieldName: string, defaultValue?: number): number | undefined {
-        return !this.hasValueDefinedFor(fieldName) ? defaultValue : this.extractRequiredNumber(fieldName);
+
+    /** ==== BOOLEAN VALUE EXTRACTION ==== **/
+    extractRequiredBoolean(key: string): boolean {
```
jfeingold35, line 96: Part of me wonders if we could simplify this by using generics, but I'm not sure how feasible that is.

packages/code-analyzer-engine-api/src/config.ts
```diff
@@ -44,13 +44,18 @@ export class ConfigValueExtractor {
...
     private readonly configRoot: string;
+    private hiddenKeys: string[] = [];
 
     constructor(configObject: ConfigObject, fieldPathRoot: string = '', configRoot: string = process.cwd()) {
         this.configObj = configObject;
         this.fieldPathRoot = fieldPathRoot;
         this.configRoot = configRoot;
     }
 
+    _addHiddenKeys(hiddenKeys: string[]) {
```
jfeingold35: Why is this method underscore-prefixed? Having a public underscore-prefixed method for testing purposes makes sense, but it looks like this is being called in other plugins, which rubs me wrong.
Also, since its most prominent use is for `disableEngine`, which isn't a hidden field, is there maybe a better name for it?

packages/code-analyzer-core/src/code-analyzer.ts
```diff
@@ -273,6 +273,10 @@ export class CodeAnalyzer {
         const engineConfigValueExtractor: engApi.ConfigValueExtractor = new engApi.ConfigValueExtractor(
             engineOverrides as engApi.ConfigObject, `${FIELDS.ENGINES}.${engineName}`, this.config.getConfigRoot());
 
+        // Although not truly a hidden field (it is well documented), we mark 'disable_engine' as a hidden field on the
+        // extractor so that each engine doesn't need to list it when using the validateOnlyContainsKeys method.
+        engineConfigValueExtractor._addHiddenKeys([FIELDS.DISABLE_ENGINE]);
```
jfeingold35: I can't quite articulate why, but it feels weird to me that we do this in Core instead of the API. Am I missing something?

packages/code-analyzer-regex-engine/test/plugin.test.ts
```diff
@@ -283,7 +290,7 @@ describe('RegexEnginePlugin Custom Config Tests', () => {
         const valueExtractor: ConfigValueExtractor = new ConfigValueExtractor(rawConfig, 'engines.regex');
         await expect(plugin.createEngineConfig("regex", valueExtractor)).rejects.toThrow(
             getMessageFromCatalog(SHARED_MESSAGE_CATALOG, 'ConfigValueMustBeOfType',
-                'engines.regex.custom_rules.NoTodos.regex', 'string', 'undefined'));
+                'engines.regex.custom_rules.NoTodos.regex', 'string', 'null'));
```
jfeingold35, line 293: Why did this change?

packages/code-analyzer-retirejs-engine/src/plugin.ts
```diff
@@ -0,0 +1,26 @@
...
+    getAvailableEngineNames(): string[] {
+        return [RetireJsEngine.NAME];
+    }
+
+    async createEngine(engineName: string, _config: ConfigObject): Promise<Engine> {
+        validateEngineName(engineName);
+        return new RetireJsEngine();
+    }
+
+    async createEngineConfig(engineName: string, configValueExtractor: ConfigValueExtractor): Promise<ConfigObject> {
```
jfeingold35, line 15: If I'm understanding this correctly, we needed to implement this method because now we need to actively validate that the config is empty instead of simply ignoring the config altogether?

packages/code-analyzer-retirejs-engine/test/plugin.test.ts
```diff
@@ -0,0 +1,51 @@
...
+    it('When createEngine is passed retire-js then an RetireJsEngine instance is returned', async () => {
+        expect(await plugin.createEngine('retire-js', {})).toBeInstanceOf(RetireJsEngine);
+    });
+
+    it('When createEngine is passed anything else then an error is thrown', async () => {
+        await expect(plugin.createEngine('oops', {})).rejects.toThrow(
+            getMessage('UnsupportedEngineName' ,'oops'));
+    });
+
+    it('When createEngineConfig is an invalid engine name, then error', async () => {
```
jfeingold35: "When createEngineConfig is _passed_ an invalid engine name"?

packages/code-analyzer-engine-api/src/index.ts
```diff
@@ -4,6 +4,7 @@ export {
     ConfigObject,
     ConfigValue,
     ConfigValueExtractor,
+    getValueUsingCaseInsensitiveKey,
```
jfeingold35, line 7: Why is this method being exported now? I didn't see it used anywhere.

packages/code-analyzer-engine-api/src/config.ts
stephen-carter-at-sf: But these are indeed the actual keys for the object that the user provided. Notice line 75.

packages/code-analyzer-engine-api/src/config.ts
stephen-carter-at-sf: Because ConfigObject has finite number of types, I wanted to reflect that and not go with generics. The validator uses generics with validateType.

packages/code-analyzer-engine-api/src/config.ts
stephen-carter-at-sf: Actually the original intention was for custom_engine_plugin_modules field.

Let me rename it to addKeysThatBypassValidation.

packages/code-analyzer-core/src/code-analyzer.ts
stephen-carter-at-sf: Core is the place where we extract out of the CodeAnalyzerConfig the engine specific config. And it is the consumer of the disable_engine field. It just so happens to put this value not at the top level but inside of each of the engine config sections. So the engines don't need to each manually add this themselves.

packages/code-analyzer-regex-engine/test/plugin.test.ts
stephen-carter-at-sf: Since I now normalize to values listed underneath the ConfigValue datatype, there is no longer such thing as undefined. See the getValueUsingCaseInsensitiveKey method which we use not instead of directly returning config[key].

packages/code-analyzer-retirejs-engine/src/plugin.ts
stephen-carter-at-sf: Yes exactly.

packages/code-analyzer-engine-api/src/index.ts
stephen-carter-at-sf: The core module also uses it in order to make the engine names and rule names case insensitive. See core's config.ts file at lines 164, 168, and 172.

packages/code-analyzer-engine-api/src/config.ts
stephen-carter-at-sf: Fixed.

packages/code-analyzer-engine-api/src/config.ts
stephen-carter-at-sf: Fixed.

packages/code-analyzer-retirejs-engine/test/plugin.test.ts
stephen-carter-at-sf: Thanks. Fixed.

jfeingold35 (approved):

PR #163, opened 2024-12-13
NEW(pmd): @W-17386449@: Add file_extensions config field for both cpd and pmd engines

stephen-carter-at-sf:


packages/code-analyzer-core/test/config.test.ts
```diff
@@ -171,7 +171,7 @@ describe("Tests for creating and accessing configuration values", () => {
     it("When top level config has an unknown key, then we error", () => {
         expect(() => CodeAnalyzerConfig.fromObject({doesNotExist: 3})).toThrow(
             getMessageFromCatalog(SHARED_MESSAGE_CATALOG,'ConfigObjectContainsInvalidKey','<TopLevel>', 'doesNotExist',
-                '["config_root","log_folder","rules","engines"]'));
+                '["config_root","engines","log_folder","rules"]'));
```
stephen-carter-at-sf, line 174: I went ahead and updated engine-api to always sort the fields for the error message in the validation method. Thus these got re-ordered.

packages/code-analyzer-eslint-engine/src/config.ts
```diff
@@ -158,54 +158,35 @@ class ESLintEngineConfigValueExtractor {
...
-        const fileExtsObjExtractor: ConfigValueExtractor = this.delegateExtractor.extractObjectAsExtractor('file_extensions');
-
-        // Validate languages
-        const validLanguages: string[] = Object.keys(DEFAULT_CONFIG.file_extensions);
-        for (const key of fileExtsObjExtractor.getKeys()) {
-            // Note: In the future we may want to make the languages case-insensitive. Right now it is a little tricky
-            //       because the extract* methods (like extractArray) look for the exact key name.
-            if (!(validLanguages.includes(key))) {
-                throw new Error(getMessage('InvalidFieldKeyForObject', fileExtsObjExtractor.getFieldPath(), key, validLanguages.join(', ')))
+        const fileExtensionsExtractor: ConfigValueExtractor = this.delegateExtractor.extractObjectAsExtractor(
```
stephen-carter-at-sf, line 161: I wanted to make the code more simple and robust for file_extensions field extraction for all three: pmd, cpd, and eslint. This is why eslint got touched as well. You'll notice the algorithm is nearly the same now between them.

packages/code-analyzer-pmd-engine/pmd-cpd-wrappers/src/main/java/com/salesforce/sfca/cpdwrapper/CpdRunner.java
```diff
@@ -53,16 +54,23 @@ private CpdLanguageRunResults runLanguage(String language, List<Path> pathsToSca
...
+        // Force the language so that pmd doesn't look at file extensions. Note: we already associated the files based
+        // on their file extensions to the correct languages the typescript side.
         Language cpdLanguageId = config.getLanguageRegistry().getLanguageById(language);
         if (cpdLanguageId == null) {
             throw new RuntimeException("The language \"" + language + "\" is not recognized by CPD.");
         }
-        config.setOnlyRecognizeLanguage(cpdLanguageId);
+        LanguageVersion forcedLangVer = config.getLanguageVersionDiscoverer()
+                .getDefaultLanguageVersion(cpdLanguageId);
+        config.setForceLanguageVersion(forcedLangVer);
```
stephen-carter-at-sf, line 67: It turns out that setOnlyRecognizeLanguage wasn't doing what I thought it was doing. It was limiting what languages should run to only xml... but didn't force it to run on all files provided. I did this correctly already with PMD - so I just had to fix it here with CPD. 

It didn't show up as a bug before because before we were only passing in xml files for xml language for example. But now we can pass in .txt files for xml language if we want to - thus fixing this to correctly use setForceLanguageVersion now.

packages/code-analyzer-pmd-engine/src/cpd-engine.ts
```diff
@@ -41,7 +41,7 @@ export class CpdEngine extends Engine {
         this.cpdWrapperInvoker = new CpdWrapperInvoker(javaCommandExecutor,
             (logLevel: LogLevel, message: string) => this.emitLogEvent(logLevel, message));
         this.config = config;
-        this.extensionToLanguageMap = toExtensionsToLanguageMap(DEFAULT_FILE_EXTENSIONS); // TODO: Make this configurable
+        this.extensionToLanguageMap = toExtensionsToLanguageMap(config.file_extensions);
```
stephen-carter-at-sf, line 44: This is basically the big win for both cpd and pmd.

jfeingold35 (approved):

PR #172, opened 2024-12-13
CHANGE @W-17397505@ Added core version to certain outputs

jfeingold35:
Adds the `code-analyzer-core` version to certain output formats, with the name `"code-analyzer"` and the value being pulled from Core's `package.json` version property.
The outputs affected are:
- XML
- JSON
- HTML (indirectly via JSON)

I chose not to include it in SARIF because I couldn't figure out a good place to slot it into the schema, since It's not technically a tool that has violations or rules associated with it.


packages/code-analyzer-core/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@salesforce/code-analyzer-core",
   "description": "Core Package for the Salesforce Code Analyzer",
-  "version": "0.19.1",
+  "version": "0.20.0-SNAPSHOT",
```
jfeingold35, line 4: I wasn't sure whether we wanted to do an `0.20.0` for this or an `0.19.2`. I'm fine with either.

packages/code-analyzer-core/src/output-format.ts
```diff
@@ -14,6 +14,8 @@ export enum OutputFormat {
     SARIF = "SARIF"
 }
 
+export const CODE_ANALYZER_CORE_NAME: string = 'code-analyzer';
```
jfeingold35, line 17: If memory serves, this was the name we agreed on, and it means all our conventions for engine names (lowercase, skewer-case, etc).

packages/code-analyzer-core/src/results.ts
```diff
@@ -238,17 +240,27 @@ export class UnexpectedErrorEngineRunResults implements EngineRunResults {
...
 
-    getRunDirectory() {
+    getRunDirectory(): string {
         return this.runDir;
     }
 
+    getCoreVersion(): string {
+        if (!this.coreVersion) {
+            const pathToPackageJson: string = path.join(__dirname, '..', 'package.json');
+            const packageJson: { version: string } = JSON.parse(fs.readFileSync(pathToPackageJson, 'utf-8'));
```
jfeingold35, line 258: I'm using `fs.readFileSync()` here instead of a Promise because I wanted to keep the results formatting a synchronous process instead of introducing `await`s everywhere.
We cache the result to make sure we only do the read once, so there shouldn't be a noticeable impact on performance.

stephen-carter-at-sf (approved):

PR #173, opened 2024-12-16
NEW: @W-17310939@: Add in our first AppExchange security rule...

stephen-carter-at-sf:
... to serve as a template for adding in more

### DIFF
--- a/packages/code-analyzer-pmd-engine/package.json
+++ b/packages/code-analyzer-pmd-engine/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@salesforce/code-analyzer-pmd-engine",
   "description": "Plugin package that adds 'pmd' and 'cpd' as engines into Salesforce Code Analyzer",
-  "version": "0.16.2-SNAPSHOT",
+  "version": "0.17.0-SNAPSHOT",
   "author": "The Salesforce Code Analyzer Team",
   "license": "BSD-3-Clause",
   "homepage": "https://developer.salesforce.com/docs/platform/salesforce-code-analyzer/overview",
--- a/packages/code-analyzer-pmd-engine/package.json
+++ b/packages/code-analyzer-pmd-engine/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@salesforce/code-analyzer-pmd-engine",
   "description": "Plugin package that adds 'pmd' and 'cpd' as engines into Salesforce Code Analyzer",
-  "version": "0.16.1",
+  "version": "0.16.2-SNAPSHOT",
   "author": "The Salesforce Code Analyzer Team",
   "license": "BSD-3-Clause",
   "homepage": "https://developer.salesforce.com/docs/platform/salesforce-code-analyzer/overview",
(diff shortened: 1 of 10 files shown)


packages/code-analyzer-pmd-engine/src/config.ts
```diff
@@ -312,14 +312,14 @@ abstract class SharedConfigValueExtractor {
         const extToLangMap: Map<string, Language> = new Map(); // To keep track if file extension shows up with more than one language
         const fileExtensionsMap: Record<Language, string[]> = {... DEFAULT_FILE_EXTENSIONS}; // Start with copy
         for (const language of Object.keys(fileExtensionsMap) as Language[]) {
-            const fileExts: string[] = makeUnique(fileExtensionsExtractor.extractArray(language,
+            const fileExts: string[] = makeUniqueCaseInsensitive(fileExtensionsExtractor.extractArray(language,
```
stephen-carter-at-sf, line 315: Since the AppExchange rules will need to add in more default file_extensions like `.remoteSite`... i'd prefer to not force the conversion of our defaults to all be lowercase since `.remoteSite` looks nicer than `.remotesite` and is what is all over the public documentation.

But things still work case insensitive as we expect. But it did require me to change the makeUnique function to be a makeUniqueCaseInsensitive routine instead.

packages/code-analyzer-pmd-engine/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@salesforce/code-analyzer-pmd-engine",
   "description": "Plugin package that adds 'pmd' and 'cpd' as engines into Salesforce Code Analyzer",
-  "version": "0.16.1",
+  "version": "0.16.2-SNAPSHOT",
```
jfeingold35: Question: I know we typically only do new Minor Versions when Core changes, but does it make sense for us to add these new rules in a patch release?

packages/code-analyzer-pmd-engine/pmd-rules/src/test/resources/sfca/rulesets/appexchange_xml/xml/AvoidInsecureHttpRemoteSiteSetting.xml
```diff
@@ -0,0 +1,39 @@
...
+        <description>When url contains https then violation should not be reported</description>
+        <expected-problems>0</expected-problems>
+        <code><![CDATA[
+<?xml version="1.0" encoding="UTF-8"?>
+<RemoteSiteSetting xmlns="http://soap.sforce.com/2006/04/metadata">
+    <description>Used for Apex callout to mapping web service</description>
+    <disableProtocolSecurity>false</disableProtocolSecurity>
+    <isActive>true</isActive>
+    <url>https://www.maptestsite.net/mapping1</url>
+</RemoteSiteSetting>
```
stephen-carter-at-sf, line 35: I got this example right from https://developer.salesforce.com/docs/atlas.en-us.api_meta.meta/api_meta/meta_remotesitesetting.htm

jfeingold35 (approved):

packages/code-analyzer-pmd-engine/package.json
stephen-carter-at-sf: Yeah - I agree... now that we have a system in place to keep track of version numbers better... I don't think we need to keep all the engines in sync with their minor version anymore... So i'll make this 0.17.0-SNAPSHOT.

stephen-carter-at-sf: Done.

PR #179, opened 2024-12-20
NEW(eslint): @W-17378695@: Add in rule mappings to lock in severities and tags for eslint

stephen-carter-at-sf:


packages/code-analyzer-eslint-engine/test/rule-mappings.test.ts
```diff
@@ -0,0 +1,35 @@
...
+describe('Tests for the rule-mappings', () => {
+    it('Test that the list of all bundled rules matches our RULE_MAPPINGS list', async () => {
+        const engine: ESLintEngine = new ESLintEngine(DEFAULT_CONFIG);
+        const ruleDescriptions: RuleDescription[] = await engine.describeRules({});
+        const actualRuleNames: Set<string> = new Set(ruleDescriptions.map(rd => rd.name));
+        const ruleNamesInRuleMappings: Set<string> = new Set(Object.keys(RULE_MAPPINGS));
+
+        const unusedRuleNamesFromRuleMappings: Set<string> = setDiff(ruleNamesInRuleMappings, actualRuleNames);
+        if (unusedRuleNamesFromRuleMappings.size > 0) {
+            throw new Error("The following rule names are found in RULE_MAPPINGS but do not associated to any bundled eslint rules that come with our base configs:\n    " +
```
jfeingold35, line 15: Is there a reason that you're using `if`s and throwing errors instead of using `expect()`?

jfeingold35 (approved):

packages/code-analyzer-eslint-engine/test/rule-mappings.test.ts
stephen-carter-at-sf: Yes... so that I can have a really nice helpful error message displayed with everything I need. Many of the `expect` methods don't actually have diagnostic message capabilities surprisingly enough. That is
```
toHaveLength(expected: number): R;
```
is the only signature here.

PR #181, opened 2024-12-23
NEW(eslint): @W-17495039@: Allow users to specify other file extensions tha…

stephen-carter-at-sf:
…t aren't associated with base plugins but only custom plugins

### DIFF
--- a/packages/code-analyzer-eslint-engine/src/base-config.ts
+++ b/packages/code-analyzer-eslint-engine/src/base-config.ts
@@ -53,36 +53,24 @@ export class LegacyBaseConfigFactory {
     }
 
     private createJavascriptConfig(baseRuleset: BaseRuleset): Linter.ConfigOverride {
-        return {
+        const jsConfig: Linter.ConfigOverride = {
             files: this.config.file_extensions.javascript.map(ext => `*${ext}`),
             extends: [`eslint:${baseRuleset}`]
         }
+        return this.addJavascriptParser(jsConfig);
     }
 
     private createLwcConfig(): Linter.ConfigOverride {
-        return {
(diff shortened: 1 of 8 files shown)


packages/code-analyzer-eslint-engine/src/base-config.ts
```diff
@@ -55,7 +55,21 @@ export class LegacyBaseConfigFactory {
...
+                    babelrc: false,
+                    configFile: false,
+                    parserOpts: {
+                        plugins: [
+                            "classProperties",
+                            ["decorators", { "decoratorsBeforeExport": false }]
+                        ]
+                    }
+                }
+            }
```
stephen-carter-at-sf, line 72: I realized that the base config needed a parser in some testing. Just fixed this.

packages/code-analyzer-eslint-engine/src/base-config.ts
```diff
@@ -53,36 +53,24 @@ export class LegacyBaseConfigFactory {
     }
 
     private createJavascriptConfig(baseRuleset: BaseRuleset): Linter.ConfigOverride {
-        return {
+        const jsConfig: Linter.ConfigOverride = {
             files: this.config.file_extensions.javascript.map(ext => `*${ext}`),
             extends: [`eslint:${baseRuleset}`]
         }
+        return this.addJavascriptParser(jsConfig);
```
stephen-carter-at-sf, line 60: [Fix as you go item] I realized that the base config needed a parser while doing some addition testing on just the base rules. I just assumed that the default would just work out of the box. I never caught this before because the LWC rules were always pulling in the parser. So this only comes up when you disable the lwc rules but keep the base rules enabled.

jfeingold35 (approved):

PR #189, opened 2025-01-09
CHANGE(eslint): @W-17517304@: Add LWC and category tags to lwc rules

stephen-carter-at-sf:


packages/code-analyzer-eslint-engine/src/rule-mappings.ts
```diff
@@ -817,31 +820,31 @@ export const RULE_MAPPINGS: Record<string, {severity: SeverityLevel, tags: strin
     // =================================================================================================================
     "@lwc/lwc/no-deprecated": {
         severity: SeverityLevel.Moderate,
-        tags: [COMMON_TAGS.RECOMMENDED, COMMON_TAGS.LANGUAGES.JAVASCRIPT]
+        tags: [COMMON_TAGS.RECOMMENDED, LWC, COMMON_TAGS.CATEGORIES.BEST_PRACTICES, COMMON_TAGS.LANGUAGES.JAVASCRIPT]
```
jag-j, line 823: Where do we get the category information for these lwc rules? I searched up the eslint-plugin-lwc docs, but couldn't find them.

stephen-carter-at-sf: Our team decides the category tags that we want to assign these rules. We work with the teams who own these plugins to decide what is best. It isn't in their docs, not would it be because eslint by itself doesn't have a notion of categories or tags really.

jag-j (approved):

PR #190, opened 2025-01-13
FIX(core): @W-16960743@: Fix edge cases for dynamic importing on Windows and a few other items

stephen-carter-at-sf:


packages/code-analyzer-core/test/stubs.ts
```diff
@@ -1,14 +1,4 @@
 import * as engApi from "@salesforce/code-analyzer-engine-api"
-import {
```
stephen-carter-at-sf, line 2: While I was in here, I thought it best to clean up these imports to make it clear they are from the engine api

packages/code-analyzer-core/src/code-analyzer.ts
```diff
@@ -99,9 +99,18 @@ export class CodeAnalyzer {
...
+            try {
+                resolvedModulePath = require.resolve(enginePluginModulePath, {paths: [this.config.getConfigRoot()]});
+            } catch (ignored) {
+                // On windows, there is an edge case where a standalone file in the same directory as the user's config
+                // file may not be resolved by require.resolve if given as just the file name. So we attempt to resolve
+                // this using path.resolve for this edge case.
+                resolvedModulePath = path.resolve(this.config.getConfigRoot(), enginePluginModulePath);
+            }
+            // To avoid issues with dynamically importing absolute paths on Windows, we need to use 'file://' URI format
+            pluginModule = (await import(`file://${resolvedModulePath.replaceAll('\\', '/')}`));
```
stephen-carter-at-sf: I thoroughly tested a bunch of edge cases on my windows machine. There isn't an easy way to add in unit tests for these edge cases without somehow setting up a javascript specific structure that has a node modules folder that somehow symlinks back to the engine api.

The existing tests cover the typical cases... so I think we are good without having to add in a whole testing scenario for these windows edge cases... especially since this functionality is not even documented/exposed to end users at all and is still for internal testing purposes only.

packages/code-analyzer-flowtest-engine/src/python/FlowTestCommandWrapper.ts
```diff
@@ -95,7 +95,7 @@ export class RunTimeFlowTestCommandWrapper implements FlowTestCommandWrapper {
     // istanbul ignore next
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     private executionResultsAreValid(executionResults: any): executionResults is FlowTestExecutionResult {
-        if (!('results' in executionResults) || typeof executionResults.results !== 'object') {
+        if (!executionResults.results || typeof executionResults.results !== 'object') {
```
stephen-carter-at-sf, line 98: Fix as you go item - thought I'd fix small bug. Note that if executionResults.results is equal to null we also want to return false - thus updating this if statement to be more robust.

packages/code-analyzer-core/src/results.ts
```diff
@@ -159,7 +159,8 @@ export class UnexpectedEngineErrorViolation implements Violation {
     }
 
     getMessage(): string {
-        return getMessage('UnexpectedEngineErrorViolationMessage', this.engineName, this.error.message);
+        return getMessage('UnexpectedEngineErrorViolationMessage', this.engineName,
+            this.error.stack ? this.error.stack : this.error.message); // Prefer to get the whole stack when possible
```
stephen-carter-at-sf, line 163: Tiny improvement to help us get the full message+stack trace when possible if an engine throws an error.... this will help us diagnose issues from users much more easily.

jag-j (approved):

PR #192, opened 2025-01-15
FIX(flowtest): @W-17599599@: Fix flowtest bugs on no-violation scenarios and fix tests

stephen-carter-at-sf:
Note that I found a number of bugs that I actually fixed. All of them were because the tests were either missing or was written incorrectly. Three of them include:
* The run progress wasn’t actually working because the wrote emit function was being used.
* The case where the workspace didn’t have any flow files in the directory wasn’t working because the flowtest utility doesn’t handle empty directories properly (but we can handle this on our end)
* The case where the flow files don’t actually have any violations in them - we were throwing null pointer exception.

I’ve completely updated all the tests and now all the known bugs are fixed.


packages/code-analyzer-flowtest-engine/src/python/FlowTestCommandWrapper.ts
```diff
@@ -116,9 +118,8 @@ export class RunTimeFlowTestCommandWrapper implements FlowTestCommandWrapper {
         return true;
     }
 
-    // istanbul ignore next
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
stephen-carter-at-sf, line 120: I also fixed all the cases where "any" was being used so that we can remove the disabled eslint lines.

packages/code-analyzer-flowtest-engine/test/python/FlowTestCommandWrapper.test.ts
```diff
@@ -20,22 +22,20 @@ describe('FlowTestCommandWrapper implementations', () => {
...
+                            delete queryResults.counter;
                         }
                     }
-                }, 30000);
+                });
 
                 it('Correctly reads and parses results', async () => {
                     const goldfileName: string = 'results.goldfile.json';
                     const goldFileContents: string = (await fs.readFile(path.join(PATH_TO_GOLDFILES, goldfileName), {encoding: 'utf-8'}))
-                        .replaceAll('"__PATH_TO_SUBFLOW_TEST1__"', JSON.stringify(path.join(PATH_TO_WORKSPACES, 'contains-multiple-flows', 'subflow_test1.flow-meta.xml')))
```
stephen-carter-at-sf, line 35: These parent and subflow files don't contain violations. So I am waiting for Robert Sussland to give us actual working examples. These weren't doing anything for us in terms of testing.

packages/code-analyzer-flowtest-engine/src/engine.ts
```diff
@@ -77,13 +80,22 @@ export class FlowTestEngine extends Engine {
         }
         this.emitRunRulesProgressEvent(PRE_INVOCATION_RUN_PERCENT);
         const percentageUpdateHandler = /* istanbul ignore next */ (percentage: number) => {
-            this.emitDescribeRulesProgressEvent(normalizeRelativeCompletionPercentage(percentage));
+            this.emitRunRulesProgressEvent(normalizeRelativeCompletionPercentage(percentage));
```
stephen-carter-at-sf, line 83: This bug would have easily been caught if there was even 1 test for the run progress. This just shows how important testing is.

stephen-carter-at-sf:
fixes https://github.com/forcedotcom/sfdx-scanner/issues/1724

packages/code-analyzer-flowtest-engine/src/engine.ts
```diff
@@ -77,13 +80,22 @@ export class FlowTestEngine extends Engine {
...
         const executionResults: FlowTestExecutionResult = await this.commandWrapper.runFlowTestRules(workspaceRoot, percentageUpdateHandler);
-        const convertedResults: EngineRunResults = toEngineRunResults(executionResults, ruleNames, await runOptions.workspace.getExpandedFiles());
+        const convertedResults: EngineRunResults = toEngineRunResults(executionResults, ruleNames, relevantFiles);
         this.emitRunRulesProgressEvent(100);
         return convertedResults;
     }
+
+    private async getRelevantFiles(workspace: Workspace): Promise<string[]> {
+        const cacheKey: string = workspace.getWorkspaceId();
+        if (!this.relevantFilesCache.has(cacheKey)) {
```
jag-j, line 93: What happens in the case of new files being added/deleted?

stephen-carter-at-sf: When code analyzer runs... there are no files being added or deleted. This cache is helpful for storing information between the describeRules and runRules commands. Nearly all engines use this technique.

If you are talking about having a persistent CodeAnalyzer class on some client, which we don't currently have... then the client would want to recreate the workspace instance before calling runRules again (at which point the workspace instance would have a new workspaceId).

packages/code-analyzer-flowtest-engine/src/engine.ts
```diff
@@ -49,14 +51,10 @@ export class FlowTestEngine extends Engine {
...
-        if (describeOptions.workspace) {
-            const workspaceFiles: string[] = await describeOptions.workspace.getExpandedFiles();
-            // If a workspace is provided but it contains no flow files, then return no rules.
-            if (!workspaceFiles.some(fileIsFlowFile)) {
-                this.emitLogEvent(LogLevel.Debug, 'Workspace contains no Flow files; returning no rules');
-                this.emitDescribeRulesProgressEvent(100);
-                return [];
-            }
+        if (describeOptions.workspace && (await this.getRelevantFiles(describeOptions.workspace)).length == 0) {
+            this.emitLogEvent(LogLevel.Debug, 'Workspace contains no Flow files. Returning no flowtest rules.');
```
jag-j: Can you remind me what is the doc review process for these messages? Do you get it reviewed from Juliet at the time of PR review?

stephen-carter-at-sf: These are debug logs which only show up in the log file - Josh added it in... I personally don't think it's needed. But your point is valid, it probably should be moved to the message catalog unless it is a "Fine" level (in which fine is really for our own diagnostic purposes). Let me switch this to "Fine" real quick.

stephen-carter-at-sf: Done.

packages/code-analyzer-flowtest-engine/src/engine.ts
jag-j: Ok, that was the missing piece. I wasn't sure when this cache is generated. We are good if this is between describeRules and runRules. In vscode, this is different as we persist some caches across runs and hence the question.

jag-j (approved):

PR #195, opened 2025-01-21
FIX(eslint): @W-17530186@: Incorporate fixes for LWC dependencies

stephen-carter-at-sf:
In addition to updating our LWC dependencies, I added a commit since it bothered me that I had so many gold files with all the same rules in it. So I reworked the unit tests so to reduce how many gold files we need so that we only have to update one or 2 gold files at most when rules change instead of a bunch of gold files each time


packages/code-analyzer-eslint-engine/test/test-data/legacyConfigCases/workspace_HasCustomConfigModifyingExistingRules/.eslintrc.json
```diff
@@ -1,7 +1,4 @@
 {
-  "extends": [
```
jfeingold35, line 2: What motivated this?

packages/code-analyzer-eslint-engine/test/test-data/legacyConfigCases/rules_OnlyCustomConfigModifyingExistingRules.goldfile.json
```diff
@@ -1,809 +1,42 @@
 [
   {
-    "name": "array-callback-return",
```
jfeingold35, line 3: I'm not sure I understand why you can get rid of all the rule declarations in here. Could you explain that please?

packages/code-analyzer-eslint-engine/test/test-data/legacyConfigCases/workspace_HasCustomConfigModifyingExistingRules/.eslintrc.json
stephen-carter-at-sf: Trying to remove redundancy of the same rules in multiple gold files. Having all the recommended rules here didn't provide any value in testing. Removing it on the other hand does provide value - easier to maintain gold files.

packages/code-analyzer-eslint-engine/test/test-data/legacyConfigCases/rules_OnlyCustomConfigModifyingExistingRules.goldfile.json
stephen-carter-at-sf: Tied you your other question. Now the custom config doesn't pull in all the recommended rules... instead it just pulls in just 3 with modifications.

jfeingold35: So it sounds like this was a behavioral change in LWC that we are changing our tests to account for?

stephen-carter-at-sf: I removed `"extends": [ ...` from the custom config... to no longer pull in these rules in this gold file. This has nothing to do with LWC. See commit 1 for the primary change and 2 for the test updates to make maintaining the gold files easier.

jfeingold35 (approved):

PR #200, opened 2025-01-30
CHANGE: @W-17662327@: Update dependencies where possible

stephen-carter-at-sf:


.github/workflows/publish-package-to-npm.yml
```diff
@@ -229,11 +229,14 @@ jobs:
...
-              npm publish --tag latest-alpha --access public --verbose --dry-run
+              npm publish --tag latest-beta --access public --verbose --dry-run
               echo "Fake-published ${PACKAGE_NAME}@${PUBLISHED_VERSION}"
             else
-              npm publish --tag latest-alpha --access public --verbose
+              npm publish --tag latest-beta --access public --verbose
               echo "Published ${PACKAGE_NAME}@${PUBLISHED_VERSION}"
+          
+              npm dist-tag add @salesforce/${PACKAGE_NAME}@${PUBLISHED_VERSION} latest
+              echo "Added latest tag to @salesforce/${PACKAGE_NAME}@${PUBLISHED_VERSION}"
```
stephen-carter-at-sf, line 239: Also while I was in here, I thought we start actually updating the latest tag for these packages when we publish as well. The latest for all these packages is still on 0.1.0 so this will help us start being able to use @latest in things like our doc generation tools soon.

packages/T-E-M-P-L-A-T-E/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@salesforce/t-e-m-p-l-a-t-e",
   "description": "T-E-M-P-L-A-T-E",
-  "version": "0.17.0-SNAPSHOT",
+  "version": "0.1.0-SNAPSHOT",
```
stephen-carter-at-sf, line 4: No need to keep T-E-M-P-L-A-T-E package.json anything other than 0.1.0-SNAPSHOT ... so just going to fix this now so we don't need to touch it again... since we don't publish this package.

jfeingold35 (approved):

PR #201, opened 2025-01-30
FIX(retire-js): @W-17711048@: Use spawn instead of exec to fix issue with spaces in command path

stephen-carter-at-sf:
This isn't something we can easily unit test because the command is run from the directory in which the code analyzer plugin is installed. So no unit tests were added.
But I was able to reproduce the issue with:
```
git clone git@github.com:forcedotcom/code-analyzer-core.git "folder with spaces"
cd "folder with spaces"
npm install
npm run build
npm run test
```
which made the retire-js tests fail.

With this change, the tests pass again.


stephen-carter-at-sf:
fixes https://github.com/forcedotcom/sfdx-scanner/issues/1737

jfeingold35 (approved):

PR #205, opened 2025-02-14
FIX(pmd): @W-17829775@: Use correct java command and log full command

stephen-carter-at-sf:


packages/code-analyzer-pmd-engine/src/utils.ts
```diff
@@ -22,17 +22,23 @@ const NO_OP = () => {};
...
         return new Promise<void>((resolve, reject) => {
             const stderrMessages: string[] = [];
             const allJavaArgs: string[] = javaClassPaths.length == 0 ? javaCmdArgs :
                 ['-cp', javaClassPaths.join(path.delimiter), ... javaCmdArgs];
-            const javaProcess: ChildProcessWithoutNullStreams = spawn('java', allJavaArgs);
+
+            this.emitLogEvent(LogLevel.Fine, `Calling command: ${this.javaCommand} ` +
+                allJavaArgs.map(arg => arg.startsWith('-') ? arg : `"${arg}"`).join(' '));
+
+            const javaProcess: ChildProcessWithoutNullStreams = spawn(this.javaCommand, allJavaArgs);
```
stephen-carter-at-sf, line 41: This is the main fix.

The other updates were purely so we could log the full call to the logs so we can more easily see the java command being used.

jfeingold35 (approved):

PR #209, opened 2025-02-20
NEW(flowtest): @W-17701084@: Write flowtest debug log file to log folder

stephen-carter-at-sf:


packages/code-analyzer-flowtest-engine/src/engine.ts
```diff
@@ -70,11 +69,16 @@ export class FlowTestEngine extends Engine {
             return { violations: [] };
         }
 
+        const dateTimeStr: string = formatToDateTimeString(this.clock.now());
+        const logFile: string = path.join(runOptions.logFolder, `sfca-flowtest-${dateTimeStr}.log`);
```
stephen-carter-at-sf, line 73: I noticed the CLI uses an epoc timestamp. I think we might want to update things over there to create a file with a more readable date time string like I have here at some point so that we can be consistent.

jfeingold35: Yeah, that's a good idea.

jfeingold35 (approved):

PR #212, opened 2025-02-24
CHANGE @W-17865538@ Graceful recovery from uninstantiable engines

jfeingold35:


packages/code-analyzer-core/src/code-analyzer.ts
```diff
@@ -258,6 +259,9 @@ export class CodeAnalyzer {
         for (const engineRunResults of engineRunResultsList) {
             runResults.addEngineRunResults(engineRunResults);
         }
+        for (const [uninstantiableEngine, error] of this.uninstantiableEnginesMap.entries()) {
```
jfeingold35, line 263: It seemed like overkill to add a separate method for this, so I just added the for-loop inline. I'm happy to pull it into its own method if desired, though.

packages/code-analyzer-core/src/code-analyzer.ts
```diff
@@ -357,7 +361,9 @@ export class CodeAnalyzer {
             const configDescription: ConfigDescription = toConfigDescription(engineConfigDescription, engineName, engineOverrides);
             this.engineConfigDescriptions.set(engineName, configDescription);
         } catch (err) {
-            throw new Error(getMessage('PluginErrorWhenCreatingEngine', engineName, (err as Error).message));
+            this.uninstantiableEnginesMap.set(engineName, err as Error);
+            this.emitLogEvent(LogLevel.Error, getMessage('PluginErrorWhenCreatingEngine', engineName, (err as Error).message));
```
jfeingold35: The message is exactly the same; it's just emitted as an Error-level log (and mapped to the engine name) instead of thrown as an error.

packages/code-analyzer-core/src/code-analyzer.ts
```diff
@@ -387,8 +393,10 @@ export class CodeAnalyzer {
             this.listenToEngineEvents(engine);
 
         } catch (err) {
-            throw new Error(getMessage('PluginErrorWhenCreatingEngine', engineName, (err as Error).message) + '\n\n' +
-                getMessage('InstructionsToIgnoreErrorAndDisableEngine', engineName));
+            this.uninstantiableEnginesMap.set(engineName, err as Error);
```
jfeingold35, line 398: The message is exactly the same; it's just emitted as an Error-level log (and mapped to the engine name) instead of thrown as an error.

packages/code-analyzer-core/src/messages.ts
```diff
@@ -63,6 +63,12 @@ const MESSAGE_CATALOG : MessageCatalog = {
     UnexpectedEngineErrorViolationMessage:
         `The engine with name '%s' threw an unexpected error: %s`,
 
+    UninstantiableEngineErrorRuleDescription:
+        `This rule reports a violation when engine '%s' cannot be instantiated.`,
+
+    UninstantiableEngineErrorViolationMessage:
+        `The engine with name '%s' could not be instantiated. Error: %s`,
+
```
jfeingold35, line 71: @jshackell-sfdc , I'll need your input on these error messages (though obviously everything related to tomorrow's release is a higher priority than this).

packages/code-analyzer-core/src/results.ts
```diff
@@ -213,26 +213,22 @@ export class ViolationImpl implements Violation {
     }
 }
 
-export class UnexpectedEngineErrorViolation implements Violation {
-    private readonly engineName: string;
-    private readonly error: Error;
-    private readonly rule: Rule;
+abstract class AbstractLocationlessViolation implements Violation {
```
jfeingold35, line 216: Since both `UnexpectedEngineErrorViolation` and the new `UninstantiableEngineErrorViolation` are locationless, it seemed worth pulling the shared behavior into an abstract class that both (and any future similar synthetic violations) can extend.

packages/code-analyzer-core/src/results.ts
```diff
@@ -251,6 +247,30 @@ export class UnexpectedEngineErrorViolation implements Violation {
     }
 }
 
+export class UninstantiableEngineErrorViolation extends AbstractLocationlessViolation {
```
jfeingold35, line 250: Spellcheck does not recognize "uninstantiable" as a word, so I'm prepared to rename this if anyone has a better idea.

packages/code-analyzer-core/src/results.ts
```diff
@@ -286,38 +306,50 @@ export class EngineRunResultsImpl implements EngineRunResults {
     }
 }
 
-export class UnexpectedErrorEngineRunResults implements EngineRunResults {
+abstract class AbstractErroneousEngineRunResults implements EngineRunResults {
```
jfeingold35, line 309: Same as above, it made sense to put all shared behavior in an abstract parent class.

packages/code-analyzer-core/src/results.ts
```diff
@@ -286,38 +306,50 @@ export class EngineRunResultsImpl implements EngineRunResults {
...
 
-    getViolations(): Violation[] {
+    public getViolations(): Violation[] {
         return [this.violation];
     }
 }
 
+export class UninstantiableEngineRunResults extends AbstractErroneousEngineRunResults {
+    constructor(engineName: string, error: Error) {
+        super(engineName, 'unknown', new UninstantiableEngineErrorViolation(engineName, error));
```
jfeingold35, line 343: Since the engine can't even be instantiated, we don't actually know what its version is, so I'm just hardcoding "unknown".

packages/code-analyzer-core/src/rules.ts
```diff
@@ -116,38 +116,62 @@ export class RuleImpl implements Rule {
     }
 }
 
-export class UnexpectedEngineErrorRule implements Rule {
-    private readonly engineName: string;
+abstract class AbstractEngineErrorRule implements Rule {
```
jfeingold35, line 119: Reusable abstract parent class.

packages/code-analyzer-core/src/code-analyzer.ts
stephen-carter-at-sf: Although it is rare for this error to ever be thrown... since it is coming from the describeEngineConfig method... can we actually append the `InstructionsToIgnoreErrorAndDisableEngine` message as well just like we do on line 397 below? They both should be the same in my opinion.

packages/code-analyzer-core/src/messages.ts
```diff
@@ -63,6 +63,12 @@ const MESSAGE_CATALOG : MessageCatalog = {
     UnexpectedEngineErrorViolationMessage:
         `The engine with name '%s' threw an unexpected error: %s`,
 
+    UninstantiableEngineErrorRuleDescription:
+        `This rule reports a violation when engine '%s' cannot be instantiated.`,
+
+    UninstantiableEngineErrorViolationMessage:
+        `The engine with name '%s' could not be instantiated. Error: %s`,
```
stephen-carter-at-sf, line 70: This seems very similar to the PluginErrorWhenCreatingEngine message which this seems to wrap.

I understand that this is the violation message and that is the error... but I wonder if there is a way to consolidate the two... we should think about this a bit - otherwise the violation message will be unnecessarily verbose/redundant I believe.

packages/code-analyzer-core/src/code-analyzer.ts
```diff
@@ -357,7 +361,9 @@ export class CodeAnalyzer {
             const configDescription: ConfigDescription = toConfigDescription(engineConfigDescription, engineName, engineOverrides);
             this.engineConfigDescriptions.set(engineName, configDescription);
         } catch (err) {
-            throw new Error(getMessage('PluginErrorWhenCreatingEngine', engineName, (err as Error).message));
+            this.uninstantiableEnginesMap.set(engineName, err as Error);
+            this.emitLogEvent(LogLevel.Error, getMessage('PluginErrorWhenCreatingEngine', engineName, (err as Error).message));
+            return;
```
stephen-carter-at-sf, line 368: I'm wondering about this return with respect to the "config" use case.

If a user has configuration for the engine in their code-analyzer.yml file but they have a typo for a setting in that one engine... what happens? And, doesn't this wipe out their config instead of preserving it?

Instead, I wonder if we should before returning, basically do we what we do on line 372-373 where we simply return whatever the user gave us for that engine (untouched)... so as to not erase it in the `config -f` case.

packages/code-analyzer-core/src/results.ts
stephen-carter-at-sf, line 216: I understand that the errors are from 2 different sources (plugin while constructing the engine vs the engine itself)... but I have a gut feeling it might be more elegant to just have a single class (possibly renamed) with additional input arguments instead of using OOP Abstract classes in all these cases (wherever you are introducing a 2nd class. But I'm on the fence to be honest... so feel free to do as you wish here. Just think about it.

packages/code-analyzer-core/src/messages.ts
jfeingold35: This doesn't actually wrap the other one. `PluginErrorWhenCreatingEngine` is applied to the log, and this one is applied to the violation.

packages/code-analyzer-core/src/code-analyzer.ts
jfeingold35: Yeah, we could do that if we want to, though it would be a behavioral change.

packages/code-analyzer-core/src/code-analyzer.ts
jfeingold35: I'll check what happens in this case, but do we want to handle it as part of this work item or should we log it as a separate bug?

jfeingold35: Added the following to my config `engines` section:
```
retire-js:
   disable-engine: false
   asdfasdfasdfasdf : 'asdfasdfasdfasdf'
```
And when I ran `code-analyzer config` with that config file, I got:
```
 ›   Error: Core [15:28:57]: Failed to create engine with name 'retire-js' due to the following error:
 ›   The 'engines.retire-js' configuration object contains an invalid key 'asdfasdfasdf'. Valid keys are: []
 ›
 ›   If you wish to ignore this error and disable this engine, then update your Code Analyzer configuration with:
 ›   engines:
 ›     retire-js:
 ›       disable_engine: true
Selecting rules... done.
Error (1): Failed to get configuration for engine with name 'retire-js' since this engine has not been added to Code Analyzer.
```
Which actually seems pretty desirable. Do you still want to change the behavior to what you described?

stephen-carter-at-sf: Oh it still errors. Yeah this works. Is this locked in with a test by chance? I think it should be. Basically I don't want to accidentally remove their configuration in a config command output.

packages/code-analyzer-core/src/code-analyzer.ts
```diff
@@ -4,7 +4,7 @@ import {
     EngineRunResultsImpl,
     RunResults,
     RunResultsImpl,
-    UnexpectedErrorEngineRunResults
+    UnexpectedErrorEngineRunResults, UninstantiableEngineRunResults
```
stephen-carter-at-sf: Just a nitpick thing... I wish IntelliJ were smarter here - but can you manually put this on its own line to keep with the formatting.

stephen-carter-at-sf (approved):
Approved

packages/code-analyzer-core/src/code-analyzer.ts
jfeingold35: Yeah, will do.

jfeingold35:
@stephen-carter-at-sf , sure, I'll look into an alternate implementation that uses a single class instead of an abstract hierarchy; see if that's any cleaner.

packages/code-analyzer-core/src/code-analyzer.ts
jfeingold35: I'm not immediately positive. I'll check, and if it's not then I'll add something to lock it in.

PR #215, opened 2025-03-06
CHANGE @W-17916375@: make the results output formatting more specific

randi274:
Description: 

Refactors the output format code to be more specific to the results formatting use case. This will make it easier to come back and create a `RuleSelectionFormatter ` and a `JsonRulesFormatter`.

WI: W-17916375

There should be no behavior change from this PR.


packages/code-analyzer-core/src/output-format.ts
```diff
@@ -21,32 +22,32 @@ export enum OutputFormat {
 export const CODE_ANALYZER_CORE_NAME: string = 'code-analyzer';
 
 /**
- * Abstract class to convert RunResults to formatted output text
+ * Abstract class to convert data objects to formatted output text
  */
 export abstract class OutputFormatter {
```
randi274: `OutputFormatter` and `OutputFormat` are both [exported](https://github.com/forcedotcom/code-analyzer-core/blob/b43fef9f171af73746d3f33ae44a9f6272657d31/packages/code-analyzer-core/src/index.ts#L29) at the top level, so I was hesitant to go the route of simply changing the name of this to a more specific "ResultsOutputFormatter". From an outside client perspective, I would think shouldn't need anything more specific than "OutputFormatter", but I'm open to approaching this differently.

packages/code-analyzer-core/src/output-format.ts
```diff
@@ -21,32 +22,32 @@ export enum OutputFormat {
...
     /**
-     * Formats run results into output text as a string
-     * @param runResults RunResults to be formatted
+     * Formats a given data structure into output text as a string
+     * @param data the type of data to be formatted
      */
-    abstract format(runResults: RunResults): string
+    abstract format(data: RunResults | RuleSelection): string
```
randi274: Something I considered instead of extending the types here was to change the abstract type from `format` to `formatRules` and `formatResults` specifically. While that allowed the parameter types to stay more specific, I recalled in recent conversation that SARIF will never have a rules format, so I assumed would be an ill-fit to have a `formatRules` method on something that should never have it. Again, open to having my mind changed here 😄

stephen-carter-at-sf: Using the `| RuleSelection` here on the abstract class but not on the concrete classes is a bit of an abuse of the typescript/javascript language and seems like a code smell.

In another language like Java, we would have the ability to have multiple format functions - one for each input signature. But we are not using Java, so typically you avoid conditional code and bugs by keeping the classes or methods separate.

packages/code-analyzer-core/src/output-format.ts
stephen-carter-at-sf: I understand, but thankfully our CLI is the only consumer so far and thus we are allowed right now and should make breaking changes to get this all right for the long term when we expose this.

So I think the better solution here is to rename this OutputFormatter to instead be a RunResultsFormatter since it's format function works on RunResults.

Then we would have a RuleSelectionFormatter whose format function will take inRuleSelection instances.

packages/code-analyzer-core/src/output-formats/results/json-output-format.ts
```diff
@@ -86,9 +86,9 @@ export type JsonCodeLocationOutput = {
 }
 
 /**
- * Formatter for JSON Output Format
+ * Formatter for Results JSON Output Format
  */
-export class JsonOutputFormatter implements OutputFormatter {
+export class ResultsJsonOutputFormatter implements OutputFormatter {
```
stephen-carter-at-sf: Then for each of these classes, I'd call them something like JsonRunResultsFormatter or RunResultsToJsonFormatter or whatever.

packages/code-analyzer-core/src/results.ts
```diff
@@ -409,7 +409,7 @@ export class RunResultsImpl implements RunResults {
     }
 
     toFormattedOutput(format: OutputFormat): string {
-        return OutputFormatter.forFormat(format, this.clock).format(this);
+        return OutputFormatter.forResultsFormat(format, this.clock).format(this);
```
stephen-carter-at-sf: This would be then simply be RunResultsFormatter.forFormat(format, this.clock).format(this);

stephen-carter-at-sf (changes requested):
The index.ts file would need to be updated as well to expose the new RunResultsFormatter. But of course we can keep OutputFormat the same since that's just an enum.

packages/code-analyzer-core/src/output-format.ts
jfeingold35: +1

randi274: Ok cool! In that case I'm good to update that naming convention

packages/code-analyzer-core/src/output-formats/results/csv-output-format.ts
stephen-carter-at-sf, line 1: Just a suggestion... what do you think about renaming these files to be csv-run-results-format.ts, etc.
I know it is verbose. But as we the other files for the rule selection world, it will be nice to have unique filenames simple so that if we need to do a file search, we don't find 2 instead - but 1.

No big deal if you disagree.

packages/code-analyzer-core/src/output-formats/results/xml-output-format.ts
```diff
@@ -75,5 +75,6 @@ export class XmlOutputFormatter implements OutputFormatter {
         }
 
         return violationsNode.end({ pretty: true, allowEmpty: true });
+
```
stephen-carter-at-sf: Nitpick... extra line probably not needed.

stephen-carter-at-sf (approved):

packages/code-analyzer-core/src/output-formats/results/xml-output-format.ts
randi274: I think this might be something with my linter - happy to clean that up 👍

packages/code-analyzer-core/src/output-formats/results/csv-output-format.ts
randi274: I was on the fence about it, but I see the virtue in that as well! I'll make that change too

packages/code-analyzer-core/test/output-format.test.ts
jfeingold35, line 1: You might want to rename this file to `run-output-format.test.ts` or similar, unless you intend for it to also contain the rule output format tests.

randi274: I was thinking to keep them in the same spot if that works! It made the code feel a little neater to keep them both in `output-format.ts` too, unless we have a 1 class: 1 file relationship

randi274: It also meant I could keep the enum `OutputFormat` there without feeling like it needed to move to a `types` or other spot

jfeingold35: Fair enough.

PR #216, opened 2025-03-07
NEW (SFGE) @W-17915698@ Part 1 - Added simple copy of SFGE java layer

jfeingold35:
This PR the first of multiple (probably two or three) PRs that collectively comprise W-17915698. This PR is as close as possible to a straightforward copy of SFGE and cli-messaging pasted directly into `packages/code-analyzer-sfge-engine/sfge`.
I made no changes whatsoever to the files, aside from the bare minimum that was necessary for certain tests to pass.
This commit is its own pull request to make it easier to review, since only three of the ~780 changed files actually have anything net-new in them.


packages/code-analyzer-sfge-engine/.gitignore
```diff
@@ -3,3 +3,6 @@
 
 # Ignore Gradle build output directory
 build
+
+# Ignore log files
```
jfeingold35, line 7: Don't want any of the log files created by JUnit or whatever to accidentally be added.

packages/code-analyzer-sfge-engine/sfge/build.gradle.kts
```diff
@@ -20,10 +20,28 @@ repositories {
 // These dependencies use our version catalog. Versions are listed in the ../gradle/libs.versions.toml file.
 // See https://docs.gradle.org/current/userguide/platforms.html#sub::toml-dependencies-format
 dependencies {
+    implementation("commons-cli:commons-cli:1.4")
```
jfeingold35, line 23: This whole block was copy-pasted from the `build.gradle.tks` file from v4.

packages/code-analyzer-sfge-engine/sfge/build.gradle.kts
```diff
@@ -20,10 +20,28 @@ repositories {
...
+        exclude("junit")
+    }
+    implementation("org.reflections:reflections:0.9.12")
+    implementation("org.ow2.asm:asm:9.2")
+    implementation(files("lib/apex-jorje-lsp-sfge.jar"))
     // --- TEST ONLY DEPENDENCIES -----------------------------------------------
     testImplementation(libs.hamcrest)
     testImplementation(libs.junit.jupiter) // Maps to junit-jupiter
     testRuntimeOnly("org.junit.platform:junit-platform-launcher")
+    testImplementation("org.mockito:mockito-core:5.2.0")
```
jfeingold35, line 43: The mockito ones are the only ones I believe needed to make the jump here, since the other test implementation mentions are all hamcrest/junit stuff.

packages/code-analyzer-sfge-engine/sfge/build.gradle.kts
```diff
@@ -62,11 +80,16 @@ tasks.test {
     // Use JUnit 5 for unit tests.
     useJUnitPlatform()
 
+    systemProperty("junit.jupiter.extensions.autodetection.enabled", true)
```
jfeingold35, line 83: I don't fully understand why, but I had to copy-paste this from v4's gradle file before the tests would pass.

packages/code-analyzer-sfge-engine/sfge/build.gradle.kts
```diff
@@ -62,11 +80,16 @@ tasks.test {
...
+
     testLogging {
         events("passed", "skipped", "failed")
-        showStandardStreams = true
+        showStandardStreams = false
```
jfeingold35, line 87: Turning off unnecessary logging, because otherwise _holy smokes_ these tests create a lot of console noise.

packages/code-analyzer-sfge-engine/sfge/build.gradle.kts
```diff
@@ -62,11 +80,16 @@ tasks.test {
...
         exceptionFormat = org.gradle.api.tasks.testing.logging.TestExceptionFormat.FULL
     }
+
+    // Run tests in multiple threads
+    maxParallelForks = Runtime.getRuntime().availableProcessors()/2 + 1
```
jfeingold35, line 92: Parallelism copied from v4 file.

packages/code-analyzer-sfge-engine/sfge/src/main/java/com/salesforce/messaging/CliMessager.java
jfeingold35, line 1: CLI messager was packaged as `com.salesforce.messaging`, so it was easy enough for this first phase to just drop it directly into SFGE so that everything compiles. If we need to, a subsequent phase of this story can involve replacing/refactoring it.

packages/code-analyzer-sfge-engine/sfge/src/test/java/com/salesforce/messaging/EventKeyTest.java
jfeingold35, line 1: This testfile utilizes the corresponding `.md` file, which no longer exists. Thus, these tests are currently broken. To unblock the PR, I've simply opted to disable this testfile entirely. It will be enabled, refactored, or replaced in a subsequent phase of this story.

packages/code-analyzer-sfge-engine/.gitignore
```diff
@@ -2,4 +2,7 @@
 .gradle
 
 # Ignore Gradle build output directory
-build
+sfge/build
```
jfeingold35, line 5: There's a package called `/build/` in the sfge code, so we need to specifically ignore only the Gradle build folder or else we lose that package.

packages/code-analyzer-sfge-engine/.gitignore
stephen-carter-at-sf: Ideally for, we shouldn't in local directories. So this is something we'll want to change eventually for both testing and prod. Note that engines now receive the configured log directory (which by default goes into a temp folder) so we should be using that eventually (obviously not in this PR though).

packages/code-analyzer-sfge-engine/sfge/build.gradle.kts
stephen-carter-at-sf: Did you upgrade anything yet? Also, we'll want to move these libraries into the toml file to better manage the versions over time.

Also jorje itself... is that updated? Or will you do that in the next PR?

packages/code-analyzer-sfge-engine/sfge/build.gradle.kts
stephen-carter-at-sf: https://junit.org/junit5/docs/current/user-guide/#running-tests-build-gradle-config-params says:

When auto-detection is enabled, extensions discovered via the [ServiceLoader](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/ServiceLoader.html) mechanism will be added to the extension registry after JUnit Jupiter’s global extensions (e.g., support for TestInfo, TestReporter, etc.).

Which tests failed? I wonder if they have a non-standard test suite.

packages/code-analyzer-sfge-engine/sfge/src/main/java/com/salesforce/CliMessagerAppender.java
```diff
@@ -0,0 +1,82 @@
+package com.salesforce;
+
+import com.salesforce.config.SfgeConfigProvider;
+import com.salesforce.messaging.CliMessager;
```
stephen-carter-at-sf, line 4: I thought we were removing all dependencies on this CliMessager and was going to just replace it with another stdout based progress reporting mechanism? Or did you just copy this one in and we'll just use this as is?

stephen-carter-at-sf (approved):
Approved, but I'm assuming that eventually we'll be updating dependencies. Also, consider my other suggestions.

packages/code-analyzer-sfge-engine/sfge/build.gradle.kts
jfeingold35: Haven't updated any dependencies yet. That's going to be the next PR.

packages/code-analyzer-sfge-engine/sfge/build.gradle.kts
jfeingold35: There were hundreds, if not thousands, of failing to tests before I added that line.

packages/code-analyzer-sfge-engine/sfge/src/main/java/com/salesforce/CliMessagerAppender.java
jfeingold35: Yeah, I intend to rework this solution in the second or third PR, but I'm just copy-pasting for now to keep code changes minimal in this first PR.

PR #217, opened 2025-03-10
NEW @W-17916375@: add JSON rules formatter

randi274:
Description:
Adds a way to format the RuleSelection class into a JSON output string and returns that string for a client to use.


packages/code-analyzer-core/src/output-formats/format-utils.ts
```diff
@@ -0,0 +1,4 @@
+/**
+ * Sanitizing function to recast all strings
+ */
+export const sanitize: (text: string) => string = t => t;
```
randi274: I thought this would be good to pull out rather than carry as a passed function - I can make a separate PR to refactor the results if this is helpful here

stephen-carter-at-sf: The rule of thumb regarding utils files is that they should only exist if a specific utility function is shared across multiple files and they are non-trivial (meaning it isn't a one liner).

Also, I don't see why this is even needed... since it is essentially performing a no-op.

Maybe you grabbed this because it was needed for our html report in the results world (where we need to actually do sanitizing of a json string)? But we don't have that need here.

packages/code-analyzer-core/src/output-formats/rules/json-rules-format.ts
```diff
@@ -0,0 +1,59 @@
+import { RuleSelectionFormatter } from "../../output-format";
+import { Rule, RuleSelection } from "../../rules";
+import { sanitize } from "../format-utils";
+
+export type JsonRulesOutput = {
+    rules: JsonRuleOutput[];
+}
```
stephen-carter-at-sf: Can we instead call this JsonRuleSelectionOutput since it is paralleling the RuleSelection class and not the Rule class?

packages/code-analyzer-core/src/output-formats/rules/json-rules-format.ts
```diff
@@ -0,0 +1,59 @@
...
+ * Formatter for Rules JSON Output Format
+ */
+export class JsonRulesFormatter implements RuleSelectionFormatter {
+    format(ruleSelection: RuleSelection): string {
+        const rulesOutput: JsonRulesOutput = toJsonRulesOutput(ruleSelection);
+        return JSON.stringify(rulesOutput, undefined, 2);
+    }
+}
+
+function toJsonRulesOutput(ruleSelection: RuleSelection): JsonRulesOutput {
```
stephen-carter-at-sf: Maybe call this toJsonRuleSelectionOutput

packages/code-analyzer-core/src/output-formats/rules/json-rules-format.ts
```diff
@@ -0,0 +1,59 @@
...
+    return {
+        rules: toJsonRulesOutputArray(ruleSelection)
+    }
+}
+
+function toJsonRulesOutputArray(ruleSelection: RuleSelection): JsonRuleOutput[] {
```
stephen-carter-at-sf: Please call this toJsonRuleOutputArray (drop the s)

packages/code-analyzer-core/src/output-formats/rules/json-rules-format.ts
```diff
@@ -0,0 +1,59 @@
...
+    const selectedRules: Rule[] = ruleSelection.getEngineNames().flatMap(name => ruleSelection.getRulesFor(name));
+    return selectedRules.map((rule) => toJsonRuleOutput(rule));
```
stephen-carter-at-sf, line 46: I don't think we discussed what order we want... but yeah this matches what the CLI rules command current gives... so this is fine.

packages/code-analyzer-core/src/output-formats/rules/json-rules-format.ts
```diff
@@ -0,0 +1,59 @@
...
+}
+
+function toJsonRuleOutput(rule: Rule): JsonRuleOutput {
+    return {
+        name: sanitize(rule.getName()),
+        description: sanitize(rule.getDescription()),
+        engine: rule.getEngineName(),
+        severity: rule.getSeverityLevel(),
+        tags: rule.getTags().map(sanitize),
+        resources: rule.getResourceUrls()
```
stephen-carter-at-sf: No need for sanitize function here with each of these. We aren't feeding this into an html report or anything which would require it.

packages/code-analyzer-core/src/rules.ts
```diff
@@ -207,4 +214,8 @@ export class RuleSelectionImpl implements RuleSelection {
         }
         throw new Error(getMessage('RuleDoesNotExistInSelection', ruleName, engineName));
     }
+
+    toFormattedOutput(format: OutputFormat): string {
```
stephen-carter-at-sf, line 218: Please add in JSDoc

packages/code-analyzer-core/test/output-format.test.ts
```diff
@@ -28,141 +30,171 @@ beforeAll(async () => {
...
+            expect(formattedText).toEqual(expectedText);
+        });
+
+        it("When multiple rules are selected, we create json text with a populated rules array", () => {
+            const formattedText: string = ruleSelection.toFormattedOutput(OutputFormat.JSON);
+            const expectedText: string = getContentsOfExpectedOutputFile('multipleRules.goldfile.json', true, true);
+            expect(formattedText).toEqual(expectedText);
+        });
+    });
+
```
stephen-carter-at-sf, line 198: We need one more test here... please create a test where we have a rule that returns zero tags ... that we still show the tags array as empty in the JSON output instead of having the field not defined

stephen-carter-at-sf (changes requested):

packages/code-analyzer-core/src/output-formats/rules/json-rules-format.ts
randi274: So I noticed we are doing the same thing for the results formatter: https://github.com/forcedotcom/code-analyzer-core/blob/87fbc15979ed9621a3c1e42f6c9025ad8726ef51/packages/code-analyzer-core/src/output-formats/results/json-run-results-format.ts#L98

Is that not necessary there then? I was unclear why it was being used.

packages/code-analyzer-core/src/rules.ts
randi274: Sure! Do you want me to add for the result of the RuleSelectionImp methods while I'm here?

packages/code-analyzer-core/src/output-formats/rules/json-rules-format.ts
randi274: Good point of something to be aware of going forward 👍

packages/code-analyzer-core/src/output-formats/rules/json-rules-format.ts
stephen-carter-at-sf: The santizeFcn is for the results world only... it is because we need to escape quotes when putting json into html report.

packages/code-analyzer-core/src/rules.ts
stephen-carter-at-sf: Oh nevermind... I thought this was the interface.

JSDoc just needs to be on exported functionality for clients or engine api authors.
Implementation classes that are behind interfaces don't need it typically because folks should be working against interfaces.
So disregard my comment.

stephen-carter-at-sf (approved):

PR #219, opened 2025-03-11
NEW (SFGE) @W-17915698@ Updated dependencies

jfeingold35:
This is the second of multiple (probably three) PRs that add the Java layer of Salesforce Graph Engine to the SFGE engine plugin (the first PR being #216 ).
This one updates all of the dependencies to their latest minor version, switches them to the TOML version format, and makes minor changes necessary for tests to continue passing.


packages/code-analyzer-sfge-engine/sfge/src/test/java/com/salesforce/graph/build/CustomerApexVertexBuilderTest.java
```diff
@@ -390,9 +390,7 @@ public void testSharingPolicyLabels(String policy) {
                                         __.where(
                                                 __.out(Schema.PARENT)
                                                         .hasLabel(NodeType.USER_CLASS)
-                                                        .has(Schema.NAME, "SubClass")))
-                                .order()
-                                .by(Schema.NAME));
+                                                        .has(Schema.NAME, "SubClass"))));
```
jfeingold35, line 393: It turns out that the `Name` field on these vertices is actually undefined, so the sort wasn't actually doing anything. And in the latest version of Gremlin, sorting on an undefined field discards the vertices, so that broke this test.
So I just removed the sort here, and it'll be fine. This was the only test that broke, and our coverage is such that I'm confident in the rest of the codebase.

stephen-carter-at-sf (approved):

PR #222, opened 2025-03-13
FIX: @W-18030774@: Update definition of wasSuppliedByUser to treat ex…

stephen-carter-at-sf:
…plicit null and undefined values properly

### DIFF
--- a/packages/code-analyzer-core/src/utils.ts
+++ b/packages/code-analyzer-core/src/utils.ts
@@ -52,3 +52,56 @@ export class EngineProgressAggregator {
         return sumOfPercentages / Math.max(this.percentagesMap.size, 1);
     }
 }
+
+// Typescript says that {} === {} is false because it does a shallow comparison. To be more robust in our object
+// comparison we often need to do a deep equality check, thus the need for this function.
+export function deepEquals(value1: unknown, value2: unknown): boolean {
+    // Check for strict equality first
+    if (value1 === value2) {
+        return true;
+    }
+
+    // If one is null or undefined, return false
+    if (value1 == null || value2 == null) {
+        return false;
+    }
+
+    // If both are objects (arrays or plain objects), compare them deeply
+    if (typeof value1 === 'object' && typeof value2 === 'object') {
+        // If both are arrays
+        if (Array.isArray(value1) && Array.isArray(value2)) {
+            if (value1.length !== value2.length) {
+                return false;
+            }
+            // Compare each element in arrays
+            for (let i = 0; i < value1.length; i++) {
+                if (!deepEquals(value1[i], value2[i])) {
+                    return false;
+                }
+            }
+            return true;
+        }
+
+        // If both are objects (not arrays)
+        if (!Array.isArray(value1) && !Array.isArray(value2)) {
+            const keys1: string[] = Object.keys(value1);
+            const keys2: string[] = Object.keys(value2);
+
+            // If they have different numbers of keys, they are not equal
+            if (keys1.length !== keys2.length) {
+                return false;
+            }
+
+            // Compare each key and its corresponding value
+            for (const key of keys1) {
+                if (!keys2.includes(key) || !deepEquals(value1[key as keyof object], value2[key as keyof object])) {
+                    return false;
+                }
+            }
+            return true;
+        }
+    }
+
+    // For all other types (number, string, boolean, etc.), use strict equality
+    return false;
+}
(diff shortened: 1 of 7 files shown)


stephen-carter-at-sf:
Fixes https://github.com/forcedotcom/sfdx-scanner/issues/1754

packages/code-analyzer-core/test/utils.test.ts
```diff
@@ -0,0 +1,146 @@
+import {deepEquals} from "../src/utils";
+
+describe('deepEquals function tests', () => {
+
+    // Primitive types
```
jfeingold35: Personal preference: I'm always a fan of grouping related tests in nested describes, e.g., putting all of these in something like `describe('Primitive types', () => {` and others in `describe('Object comparisons', () => {`

packages/code-analyzer-core/test/utils.test.ts
```diff
@@ -0,0 +1,146 @@
...
+
+    test('should return false for null vs undefined', () => {
+        expect(deepEquals(null, undefined)).toBe(false);
+    });
+
+    test('should return true for null vs null', () => {
+        expect(deepEquals(null, null)).toBe(true);
+    });
+
+    test('should return true for undefined vs undefined', () => {
```
jfeingold35: Might it be worth adding cases that compare various things to `NaN`? `NaN` is the only value in JS that is not equal to itself (i.e., `(NaN !== NaN) === true`), so it's a case where weirdness could emerge.
Alternatively, if you say that sounds like overkill, I'm prepared to go along with that.

packages/code-analyzer-core/test/add-engines.test.ts
```diff
@@ -208,6 +208,21 @@ describe("Tests for adding engines to Code Analyzer", () => {
...
+                stubEngine1: {
+                    misc_value: null
+                }
+            }
+        }));
+        await codeAnalyzer.addEnginePlugin(new stubs.StubEnginePlugin());
+
+        const engineConfigDescription: ConfigDescription = codeAnalyzer.getEngineConfigDescription('stubEngine1');
+        const miscValueFieldDesc: ConfigFieldDescription = engineConfigDescription.fieldDescriptions['misc_value'];
+        expect(miscValueFieldDesc.wasSuppliedByUser).toEqual(false);
```
jfeingold35, line 223: This rubs me wrong. Like, it _was_ supplied by a user, it was just supplied as `null`. Are we naming things poorly, or is there a conceptual problem somewhere? Or alternatively, is there just a gap in my understanding?

packages/code-analyzer-core/src/utils.ts
```diff
@@ -51,4 +51,57 @@ export class EngineProgressAggregator {
         }
         return sumOfPercentages / Math.max(this.percentagesMap.size, 1);
     }
+}
+
+// Typescript says that {} === {} is false because it does a shallow comparison. To be more robust in our object
+// comparison we often need to do a deep equality check, thus the need for this function.
+export function deepEquals(value1: unknown, value2: unknown): boolean {
```
jfeingold35, line 58: Question: Is this implementation more correct than `JSON.parse(value1) === JSON.parse(value2)`?

packages/code-analyzer-core/test/utils.test.ts
stephen-carter-at-sf: Sure I can do that.

packages/code-analyzer-core/test/utils.test.ts
stephen-carter-at-sf: Sure... i can do that.

packages/code-analyzer-core/test/add-engines.test.ts
stephen-carter-at-sf: So when we generate a config for the user with sf code-analyzer config -f code-analyzer.yml in that file we are putting in "null" values whereever the default value is null.

But "null" is already our marker to tell our config world that the user didn't explicitly set it. In yaml, there is no "undefined"... null is the way to do this.

This test is exactly what this PR is supposed to do to fix our code in our CLI's config command. 

See over in https://github.com/forcedotcom/sfdx-scanner/blob/dev/src/lib/models/ConfigModel.ts#L92 we have
```
		if (!fieldDescription.wasSuppliedByUser && resolvedValueJson !== defaultValueJson) {
```
the wasSuppliedByUser is what need to be false when null is in the config to make things work.

packages/code-analyzer-core/src/utils.ts
stephen-carter-at-sf: That is another way of doing things... I think that is more expensive though.

packages/code-analyzer-core/test/utils.test.ts
stephen-carter-at-sf: Done.

jfeingold35 (approved):

PR #224, opened 2025-03-13
NEW (SFGE) @W-17915791@ Implemented describeRules behavior for SFGE

jfeingold35:


packages/code-analyzer-engine-api/package.json
```diff
@@ -11,11 +11,14 @@
     "directory": "packages/code-analyzer-core"
   },
   "exports": {
-    ".": "./dist/index.js"
+    ".": "./dist/index.js",
+    "./utils": "./dist/utils/index.js"
```
jfeingold35, line 15: Exporting the utils as a separate import for cleanness, the way that `fs` has `fs.promises`.

packages/code-analyzer-engine-api/src/messages.ts
```diff
@@ -53,7 +53,16 @@ export const SHARED_MESSAGE_CATALOG: MessageCatalog = {
         `The '%s' configuration value is invalid. The value '%s' must be a file instead of a folder`,
 
     ConfigFolderValueMustNotBeFile:
-        `The '%s' configuration value is invalid. The value '%s' must be a folder instead of a file.`
+        `The '%s' configuration value is invalid. The value '%s' must be a folder instead of a file.`,
+
+    ErrorParsingOutputFile:
```
jfeingold35, line 58: Moved these messages up from the PMD level.

packages/code-analyzer-engine-api/src/utils/utils.ts
```diff
@@ -0,0 +1,75 @@
+import * as tmp from 'tmp';
+import {promisify} from "node:util";
+import {getMessage} from "../messages";
+import path from "node:path";
+import {LogLevel} from "../events";
+import {ChildProcessWithoutNullStreams, spawn} from "node:child_process";
+
+tmp.setGracefulCleanup();
+// istanbul ignore next
```
jfeingold35: Attention.

packages/code-analyzer-engine-api/src/utils/utils.ts
```diff
@@ -0,0 +1,75 @@
...
+const tmpDirAsync = promisify((options: tmp.DirOptions, cb: tmp.DirCallback) => tmp.dir(options, cb));
+
+/**
+ * Creates a temporary directory that eventually cleans up after itself
+ * @param parentTempDir - if supplied, then a temporary folder is placed directly underneath this parent folder.
+ */
+// istanbul ignore next
```
jfeingold35: Attention.

packages/code-analyzer-engine-api/src/utils/utils.ts
```diff
@@ -0,0 +1,75 @@
...
+
+    constructor(javaCommand: string = 'java', emitLogEvent: (logLevel: LogLevel, message: string) => void = () => {}) {
+        this.javaCommand = javaCommand;
+        this.emitLogEvent = emitLogEvent;
+    }
+
+    async exec(javaCmdArgs: string[], javaClassPaths: string[] = [], processStdOut: ProcessStdOutFcn = NO_OP): Promise<void> {
+        return new Promise<void>((resolve, reject) => {
+            const stderrMessages: string[] = [];
+            // istanbul ignore next
```
jfeingold35, line 34: Attention.

packages/code-analyzer-engine-api/src/utils/utils.ts
```diff
@@ -0,0 +1,75 @@
...
+
+            const javaProcess: ChildProcessWithoutNullStreams = spawn(this.javaCommand, allJavaArgs);
+
+            javaProcess.stdout.on('data', (data: Buffer) => {
+                const msg: string = data.toString().trim();
+                if(msg.length > 0) { // Not sure why stdout spits out empty lines, but we ignore them nonetheless
+                    try {
+                        msg.split("\n").map(line => processStdOut(line));
+                    } catch (err) {
+                        // istanbul ignore next
```
jfeingold35, line 49: Attention.

packages/code-analyzer-engine-api/test/utils/utils.test.ts
```diff
@@ -0,0 +1,31 @@
+import {indent, JavaCommandExecutor} from "../../src/utils";
+
+describe('Tests for JavaCommandExecutor', () => {
```
jfeingold35, line 21: Tests copy-pasted from the PMD layer.

packages/code-analyzer-engine-api/src/utils/utils.ts
```diff
@@ -0,0 +1,75 @@
+import * as tmp from 'tmp';
```
jfeingold35, line 1: All of this code was copy-pasted from the PMD layer, with the sole exception being the addition of `istanbul ignore next` added in a few places where the `utils.test.ts` coverage was not by itself enough to cover that line.

packages/code-analyzer-sfge-engine/sfge/sfge-junit-03-11-2025-1.log.gz
jfeingold35, line 1: Don't know how this got here, but it shouldn't have been. So it's gone.

packages/code-analyzer-sfge-engine/sfge/src/main/java/com/salesforce/cli/OutputFormatter.java
```diff
@@ -20,6 +20,6 @@ public String formatRuleJsons(List<AbstractRule> rules) {
 
         Gson gson = new Gson();
 
-        return "CATALOG_START" + gson.toJson(descriptors) + "CATALOG_END";
+        return gson.toJson(descriptors);
```
jfeingold35, line 23: Since we're writing to a file now, no need for the string wrappers.

packages/code-analyzer-sfge-engine/sfge/src/main/java/com/salesforce/rules/AbstractRule.java
```diff
@@ -69,13 +69,15 @@ protected String getUrl() {
     public static class Descriptor {
         private final String name;
         private final String description;
+        private final int severity;
```
jfeingold35, line 74: The java rule descriptors weren't including severity for some reason. Now they do.

packages/code-analyzer-sfge-engine/src/engine.ts
jfeingold35, line 1: Used the PMD engine implementation as a broad scaffold here.

packages/code-analyzer-sfge-engine/src/sfge-wrapper.ts
jfeingold35, line 1: Used the PMD wrapper as a broad scaffold here.

packages/code-analyzer-sfge-engine/src/sfge-wrapper.ts
```diff
@@ -0,0 +1,64 @@
...
+
+    public async invokeDescribeCommand(emitProgress: (percComplete: number) => void): Promise<SfgeRuleInfo[]> {
+        const tmpDir: string = await this.getTemporaryWorkingDir();
+        const sfgeRulesOutputFile: string = path.join(tmpDir, 'ruleInfo.json');
+        emitProgress(10);
+
+        const javaCmdArgs: string[] = [SFGE_MAIN_JAVA_CLASS, 'catalog', 'all', sfgeRulesOutputFile];
+        const javaClassPaths: string[] = [path.join(SFGE_WRAPPER_LIB_FOLDER, '*')];
+
+        await this.javaCommandExecutor.exec(javaCmdArgs, javaClassPaths);
```
jfeingold35, line 41: Note: SFGE's `catalog` flow doesn't produce any meaningful logs, so I'm not passing in a `stdoutProcess` callback here. We'll need one for the `execute` flow, but that's a problem for Future Josh.

packages/code-analyzer-sfge-engine/test/engine.test.ts
jfeingold35, line 1: New tests broadly derived from the PMD engine tests.

packages/code-analyzer-sfge-engine/test/test-data/goldfiles/all_rules.goldfile.json
jfeingold35, line 1: This goldfile is derived from a totally unchanged rip of what we get from SFGE itself. I expect there to be discourse about whether the default severities and categories make sense, or if we need to do post-processing like adding/removing tags, or ignoring certain rules.
That discussion can commence now.

packages/code-analyzer-sfge-engine/package.json
```diff
@@ -14,7 +14,9 @@
   "types": "dist/index.d.ts",
   "dependencies": {
     "@salesforce/code-analyzer-engine-api": "0.18.0",
```
stephen-carter-at-sf: Needs to be updated to 0.19.0-SNAPSHOT

packages/code-analyzer-engine-api/src/utils/utils.ts
stephen-carter-at-sf: Why are we needing to ignore all this?

Maybe just add in a unit test for the utils to cover the createTempDir actually creates a directory. That should cover this right?

packages/code-analyzer-engine-api/src/utils/utils.ts
stephen-carter-at-sf: Why are we ignoring this?

Maybe just add in a unit test for the utils to cover the createTempDir actually creates a directory. That should cover this right?

packages/code-analyzer-sfge-engine/package.json
jfeingold35: Fixed.

packages/code-analyzer-engine-api/src/utils/utils.ts
stephen-carter-at-sf: How did we lose coverage here? Clearly this line gets hit with PMD use case. Is it because we don't have an example jar to test over here with a processStdOut function that errors?

packages/code-analyzer-sfge-engine/src/engine.ts
```diff
@@ -5,16 +5,30 @@ import {
...
+    Workspace
 } from '@salesforce/code-analyzer-engine-api';
+import {JavaCommandExecutor} from '@salesforce/code-analyzer-engine-api/utils';
+import {
+    RuntimeSfgeWrapper,
+    SfgeRuleInfo
+} from "./sfge-wrapper";
 import {SfgeEngineConfig} from "./config";
 
+const SFGE_RELEVANT_FILE_EXTENSIONS = ['.cls'];
```
stephen-carter-at-sf, line 20: Are we not able to run SFGE on trigger files?

packages/code-analyzer-sfge-engine/src/engine.ts
```diff
@@ -28,13 +42,55 @@ export class SfgeEngine extends Engine {
...
+    const expandedFiles: string[] = await workspace.getExpandedFiles();
+    return SFGE_RELEVANT_FILE_EXTENSIONS.some(extension => expandedFiles.some(file => file.endsWith(extension)));
+}
+
+function toRuleDescription(sfgeRuleInfo: SfgeRuleInfo): RuleDescription {
+    return {
+        name: sfgeRuleInfo.name,
+        severityLevel: sfgeRuleInfo.severity,
+        tags: [sfgeRuleInfo.category.replaceAll(' ', '')],
+        description: sfgeRuleInfo.description,
```
stephen-carter-at-sf: I think we need to add in a prefix or suffix that says something like "Note that this rule is still in Developer Preview and is subject to change." or something. You'll need to ask Juliet & John what the wording should be.

packages/code-analyzer-sfge-engine/src/engine.ts
```diff
@@ -28,13 +42,55 @@ export class SfgeEngine extends Engine {
...
+async function workspaceContainsSfgeRelevantFiles(workspace: Workspace): Promise<boolean> {
+    const expandedFiles: string[] = await workspace.getExpandedFiles();
+    return SFGE_RELEVANT_FILE_EXTENSIONS.some(extension => expandedFiles.some(file => file.endsWith(extension)));
+}
+
+function toRuleDescription(sfgeRuleInfo: SfgeRuleInfo): RuleDescription {
+    return {
+        name: sfgeRuleInfo.name,
+        severityLevel: sfgeRuleInfo.severity,
+        tags: [sfgeRuleInfo.category.replaceAll(' ', '')],
```
stephen-carter-at-sf: I don't see a rule mapping file to correct any tag issues. Probably don't need it since we own all the tooling. But I think we need at least one of the catagory and language tags from https://github.com/forcedotcom/code-analyzer-core/blob/dev/packages/code-analyzer-engine-api/src/rules.ts#L38

packages/code-analyzer-sfge-engine/src/engine.ts
```diff
@@ -28,13 +42,55 @@ export class SfgeEngine extends Engine {
...
+        description: sfgeRuleInfo.description,
+        resourceUrls: [sfgeRuleInfo.url]
```
stephen-carter-at-sf: I think this old url points to the v4 documentation... we need to remove this url until we have the new v5 rule url's set to a more permanent place.

packages/code-analyzer-sfge-engine/src/engine.ts
```diff
@@ -28,13 +42,55 @@ export class SfgeEngine extends Engine {
...
     }
+}
+
+function getCacheKey(workspace?: Workspace): string {
+    return workspace ? workspace.getWorkspaceId() : process.cwd();
+}
+
+async function workspaceContainsSfgeRelevantFiles(workspace: Workspace): Promise<boolean> {
+    const expandedFiles: string[] = await workspace.getExpandedFiles();
+    return SFGE_RELEVANT_FILE_EXTENSIONS.some(extension => expandedFiles.some(file => file.endsWith(extension)));
```
stephen-carter-at-sf: file.toLowerCase() here to allow for case insensitive extension matching?

packages/code-analyzer-sfge-engine/test/test-data/goldfiles/all_rules.goldfile.json
```diff
@@ -0,0 +1,90 @@
...
+    "name": "ApexNullPointerExceptionRule",
+    "severityLevel": 2,
+    "tags": [
+      "ErrorProne"
+    ],
+    "description": "Identifies Apex operations that dereference null objects and throw NullPointerExceptions.",
+    "resourceUrls": [
+      "https://developer.salesforce.com/docs/platform/salesforce-code-analyzer/guide/apexnullpointerexception-rule.html"
+    ]
+  },
```
stephen-carter-at-sf: * Are these severities all correct? I thought we only had 3 severity levels before and now we have 5. So I'm just asking to make sure this is what we want.

* Missing language tags. I'm surprised to see the tags here are already using the ones we wanted. I see https://developer.salesforce.com/docs/platform/salesforce-code-analyzer/guide/rules.html showed "Error-Prone" category instead of this normalized "ErrorProne"... so it looks like we are good here.

* Need to remove all these v4 specific urls.

* Need to add in to each rule description that these are "Dev Preview" based rules.

packages/code-analyzer-sfge-engine/src/engine.ts
stephen-carter-at-sf, line 1: If we ship this month, then we'll want to at least add in a config description that says something like: "Note this engine is in Developer Preview and is subject to change".

stephen-carter-at-sf (changes requested):

packages/code-analyzer-sfge-engine/src/engine.ts
jfeingold35: Nope. We've explored this, but the AST for triggers is _super_ idiosyncratic, so we've never had the resources to devote to it.

packages/code-analyzer-sfge-engine/src/engine.ts
jfeingold35: Can do.

packages/code-analyzer-sfge-engine/src/engine.ts
jfeingold35: The categories defined in SFGE actually map properly to the ones we decided for the Common Tags (notwithstanding spaces), so I think we don't necessarily need a map. We could also hardcode add all of the Language tags.

packages/code-analyzer-sfge-engine/src/engine.ts
jfeingold35: Fair enough. Just empty array for now?

packages/code-analyzer-sfge-engine/src/engine.ts
jfeingold35: Yeah, good call.

packages/code-analyzer-sfge-engine/test/test-data/goldfiles/all_rules.goldfile.json
jfeingold35: Yeah, good catch. I thought the severities were out of 5, turns out they're out of 3. So yeah, we'll want to map them in some way.

packages/code-analyzer-sfge-engine/src/engine.ts
stephen-carter-at-sf: Yeah, for now until Juliet creates pages for these.

packages/code-analyzer-sfge-engine/test/test-data/goldfiles/all_rules.goldfile.json
stephen-carter-at-sf: Or just update the severities on the JAVA side if that is easiest.

stephen-carter-at-sf (approved):

PR #230, opened 2025-03-18
NEW (SFGE) @W-17915822@ Added runRules functionality for SFGE

jfeingold35:
This PR does the following:
- Implements the `runRules` functionality for the new SFGE engine plugin
- Allows `EngineRunProgressEvents` to include a `message` in addition to their `percentComplete` property, to facilitate more informative spinners in SFGE or other long-running engines

### DIFF
--- a/packages/code-analyzer-core/src/code-analyzer.ts
+++ b/packages/code-analyzer-core/src/code-analyzer.ts
@@ -424,6 +424,7 @@ export class CodeAnalyzer {
                 type: EventType.EngineRunProgressEvent,
                 timestamp: this.clock.now(),
                 engineName: engine.getName(),
+                message: event.message,
                 percentComplete: event.percentComplete
             });
         });
--- a/packages/code-analyzer-engine-api/src/engines.ts
+++ b/packages/code-analyzer-engine-api/src/engines.ts
@@ -142,12 +142,14 @@ export abstract class Engine {
     /**
      * Convenience method that subclasses can use to easily emit an event of type {@link RunRulesProgressEvent}
      * @param percentComplete the percent of completion between 0 and 100
+     * @param message optional message
      * @protected
      */
-    protected emitRunRulesProgressEvent(percentComplete: number): void {
+    protected emitRunRulesProgressEvent(percentComplete: number, message?: string): void {
         this.emitEvent({
             type: EventType.RunRulesProgressEvent,
-            percentComplete: roundToHundredths(percentComplete)
+            percentComplete: roundToHundredths(percentComplete),
+            message
         });
     }
 }
--- a/packages/code-analyzer-engine-api/src/events.ts
+++ b/packages/code-analyzer-engine-api/src/events.ts
@@ -34,6 +34,7 @@ export type LogEvent = {
  */
 export type RunRulesProgressEvent = {
     type: EventType.RunRulesProgressEvent,
+    message?: string,
     percentComplete: number
 }
 
--- a/packages/code-analyzer-sfge-engine/.gitignore
+++ b/packages/code-analyzer-sfge-engine/.gitignore
@@ -8,4 +8,5 @@ sfge/build
 sfge/bin
 
 # Ignore log files
-*.log
+**/*.log
+**/*.gz
--- a/packages/code-analyzer-sfge-engine/.gitignore
+++ b/packages/code-analyzer-sfge-engine/.gitignore
@@ -6,7 +6,3 @@ sfge/build
 
 # Ignore Gradle bin output directory (for VS Code's gradle plugin)
 sfge/bin
-
-# Ignore log files
-**/*.log
-**/*.gz
--- a/packages/code-analyzer-sfge-engine/sfge/src/main/java/com/salesforce/Sfge.java
+++ b/packages/code-analyzer-sfge-engine/sfge/src/main/java/com/salesforce/Sfge.java
@@ -57,8 +57,7 @@ import org.apache.tinkerpop.gremlin.process.traversal.dsl.graph.GraphTraversalSo
 @SuppressWarnings(
         "PMD.SystemPrintln") // Since println is currently used to communicate to outer layer
 public class Sfge {
-    @VisibleForTesting static final int EXIT_GOOD_RUN_NO_VIOLATIONS = 0;
-    @VisibleForTesting static final int EXIT_GOOD_RUN_WITH_VIOLATIONS = 4;
+    @VisibleForTesting static final int EXIT_GOOD_RUN = 0;
     @VisibleForTesting static final int EXIT_WITH_INTERNAL_ERROR_NO_VIOLATIONS = 1;
     @VisibleForTesting static final int EXIT_WITH_INTERNAL_ERROR_AND_VIOLATIONS = 5;
 
@@ -121,7 +120,7 @@ public class Sfge {
         }
         OutputFormatter formatter = new OutputFormatter();
         dependencies.writeOutput(formatter.formatRuleJsons(rules), cap.getOutfile());
-        return EXIT_GOOD_RUN_NO_VIOLATIONS;
+        return EXIT_GOOD_RUN;
     }
 
     /** Expectations for args documented in class header above. */
@@ -204,9 +203,7 @@ public class Sfge {
                         ? EXIT_WITH_INTERNAL_ERROR_NO_VIOLATIONS
                         : EXIT_WITH_INTERNAL_ERROR_AND_VIOLATIONS;
             }
-            return violations.isEmpty()
-                    ? EXIT_GOOD_RUN_NO_VIOLATIONS
-                    : EXIT_GOOD_RUN_WITH_VIOLATIONS;
+            return EXIT_GOOD_RUN;
         }
     }
 
--- a/packages/code-analyzer-sfge-engine/sfge/src/main/java/com/salesforce/cli/OutputFormatter.java
+++ b/packages/code-analyzer-sfge-engine/sfge/src/main/java/com/salesforce/cli/OutputFormatter.java
@@ -11,7 +11,7 @@ public class OutputFormatter {
     public String formatViolationJsons(List<Violation> violations) {
         Gson gson = new Gson();
 
-        return "VIOLATIONS_START" + gson.toJson(violations) + "VIOLATIONS_END";
+        return gson.toJson(violations);
     }
 
     public String formatRuleJsons(List<AbstractRule> rules) {
--- a/packages/code-analyzer-sfge-engine/sfge/src/test/resources/log4j2.xml
+++ b/packages/code-analyzer-sfge-engine/sfge/src/test/resources/log4j2.xml
@@ -4,21 +4,12 @@
         <Console name="LogToConsole" target="SYSTEM_OUT">
             <PatternLayout pattern="%d{yyyy-MM-dd HH:mm:ss} %X{ruleRunId} %-5p %c{1}:%L - %m%n"/>
         </Console>
-        <RollingFile name="LogToFile" fileName="sfge-junit.log" filePattern="sfge-junit-%d{MM-dd-yyyy}-%i.log.gz">
-            <PatternLayout pattern="%d{yyyy-MM-dd HH:mm:ss} %X{ruleRunId} %-5p %c{1}:%L - %m%n"/>
-            <Policies>
-                <TimeBasedTriggeringPolicy />
-                <SizeBasedTriggeringPolicy size="1 GB"/>
-            </Policies>
-        </RollingFile>
     </Appenders>
     <Loggers>
         <Logger name="com.salesforce" level="info" additivity="false">
-            <AppenderRef ref="LogToFile"/>
             <AppenderRef ref="LogToConsole"/>
         </Logger>
         <Root level="error">
-            <AppenderRef ref="LogToFile"/>
             <AppenderRef ref="LogToConsole"/>
         </Root>
     </Loggers>
--- a/packages/code-analyzer-sfge-engine/src/engine.ts
+++ b/packages/code-analyzer-sfge-engine/src/engine.ts
@@ -20,7 +20,6 @@ import {
     SfgeRunResult
 } from "./sfge-wrapper";
 import {SfgeEngineConfig} from "./config";
-import {dir} from "tmp";
 
 const SFGE_RELEVANT_FILE_EXTENSIONS = ['.cls'];
 
--- a/packages/code-analyzer-sfge-engine/src/engine.ts
+++ b/packages/code-analyzer-sfge-engine/src/engine.ts
@@ -86,9 +86,12 @@ export class SfgeEngine extends Engine {
             return { violations: [] };
         }
 
+        const relevantFiles: string[] = await this.getRelevantFilesInWorkspace(runOptions.workspace);
+
         const sfgeResults: SfgeRunResult[] = await this.sfgeWrapper.invokeRunCommand(
             selectedRuleInfoList,
-            runOptions.workspace,
+            relevantFiles, // TODO: WHEN WE ADD PATH-START TARGETING, THIS NEEDS TO CHANGE.
+            relevantFiles,
             (innerPerc: number, message?: string) => this.emitRunRulesProgressEvent(5 + 93*innerPerc/100, message) // 5%-98%
         );
 
--- a/packages/code-analyzer-sfge-engine/src/engine.ts
+++ b/packages/code-analyzer-sfge-engine/src/engine.ts
@@ -66,6 +66,11 @@ export class SfgeEngine extends Engine {
     public override async runRules(selectedRuleNames: string[], runOptions: RunOptions): Promise<EngineRunResults> {
         this.emitRunRulesProgressEvent(2);
 
+        if (selectedRuleNames.length === 0) {
+            this.emitRunRulesProgressEvent(100);
+            return { violations: [] };
+        }
+
         await this.validateWorkspaceCompleteness(runOptions.workspace);
 
         const allRulesInfoList: SfgeRuleInfo[] = await this.getSfgeRuleInfoList(
--- a/packages/code-analyzer-sfge-engine/src/engine.ts
+++ b/packages/code-analyzer-sfge-engine/src/engine.ts
@@ -6,6 +6,7 @@ import {
     DescribeOptions,
     Engine,
     EngineRunResults,
+    LogLevel,
     RuleDescription,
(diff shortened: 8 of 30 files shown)


packages/code-analyzer-core/src/code-analyzer.ts
```diff
@@ -424,6 +424,7 @@ export class CodeAnalyzer {
                 type: EventType.EngineRunProgressEvent,
                 timestamp: this.clock.now(),
                 engineName: engine.getName(),
+                message: event.message,
```
jfeingold35, line 427: `code-analyzer-core` passes the message up to the CLI layer (or wherever)

packages/code-analyzer-engine-api/src/events.ts
```diff
@@ -34,6 +34,7 @@ export type LogEvent = {
  */
 export type RunRulesProgressEvent = {
     type: EventType.RunRulesProgressEvent,
+    message?: string,
```
jfeingold35, line 37: `code-analyzer-engine-api` defines the new `message?` property so the plugins can use it.

packages/code-analyzer-engine-api/src/engines.ts
```diff
@@ -142,12 +142,14 @@ export abstract class Engine {
...
+     * @param message optional message
      * @protected
      */
-    protected emitRunRulesProgressEvent(percentComplete: number): void {
+    protected emitRunRulesProgressEvent(percentComplete: number, message?: string): void {
         this.emitEvent({
             type: EventType.RunRulesProgressEvent,
-            percentComplete: roundToHundredths(percentComplete)
+            percentComplete: roundToHundredths(percentComplete),
+            message
```
jfeingold35, line 152: `code-analyzer-engine-api` passes the message property up to the Core layer.

packages/code-analyzer-sfge-engine/.gitignore
```diff
@@ -5,4 +5,5 @@
 sfge/build
 
 # Ignore log files
-*.log
+**/*.log
```
jfeingold35: This should at least prevent log files and test result files from being accidentally included in a `git add`.

packages/code-analyzer-sfge-engine/sfge/src/main/java/com/salesforce/Sfge.java
```diff
@@ -57,8 +57,7 @@
 @SuppressWarnings(
         "PMD.SystemPrintln") // Since println is currently used to communicate to outer layer
 public class Sfge {
-    @VisibleForTesting static final int EXIT_GOOD_RUN_NO_VIOLATIONS = 0;
-    @VisibleForTesting static final int EXIT_GOOD_RUN_WITH_VIOLATIONS = 4;
+    @VisibleForTesting static final int EXIT_GOOD_RUN = 0;
```
jfeingold35, line 60: Since `JavaCommandExecutor` treats any non-zero exit code as a failure, SFGE has to switch to exiting with `0` for a clean run regardless of whether it produced violations.

packages/code-analyzer-sfge-engine/sfge/src/main/java/com/salesforce/cli/OutputFormatter.java
```diff
@@ -11,7 +11,7 @@ public class OutputFormatter {
     public String formatViolationJsons(List<Violation> violations) {
         Gson gson = new Gson();
 
-        return "VIOLATIONS_START" + gson.toJson(violations) + "VIOLATIONS_END";
+        return gson.toJson(violations);
```
jfeingold35, line 14: No longer need the wrapper, since we're writing to a file instead of the CLI.

packages/code-analyzer-sfge-engine/src/engine.ts
```diff
@@ -71,20 +132,46 @@ export class SfgeEngine extends Engine {
...
-        return Promise.resolve({
-            violations: []
-        });
+    private toViolation(sfgeViolation: SfgeRunResult): Violation {
+        const codeLocations: CodeLocation[] = [{
+            file: sfgeViolation.sourceFileName,
+            startLine: sfgeViolation.sourceLineNumber,
+            startColumn: sfgeViolation.sourceColumnNumber
+        }];
+        if (sfgeViolation.sinkFileName) {
```
jfeingold35, line 149: In SFGE-Land, every violation has a Source but not every violation has a Sink. So we're treating the Source as the primary location always, and also adding the Sink as a secondary location if it exists.

packages/code-analyzer-sfge-engine/src/engine.ts
```diff
@@ -58,10 +63,66 @@ export class SfgeEngine extends Engine {
         return ruleDescriptions;
     }
 
+    public override async runRules(selectedRuleNames: string[], runOptions: RunOptions): Promise<EngineRunResults> {
```
jfeingold35, line 66: This is heavily derived from the PMD Engine's implementation of `runRules`.

packages/code-analyzer-sfge-engine/src/sfge-wrapper.ts
```diff
@@ -54,10 +96,121 @@ export class RuntimeSfgeWrapper {
...
+    if (stdOutMsg.startsWith(SFCA_REALTIME_START) && stdOutMsg.endsWith(SFCA_REALTIME_END)) {
+        const sfgeMessages: SfgeMessage[] = JSON.parse(stdOutMsg.slice(
+            SFCA_REALTIME_START.length,
+            stdOutMsg.length - SFCA_REALTIME_END.length)
+        );
+        for (const sfgeMessage of sfgeMessages) {
+            const processedMessage = getMessage(sfgeMessage.messageKey, ...sfgeMessage.args);
+            if (isSfgeLogMessage(sfgeMessage)) {
+                if (sfgeMessage.messageSeverity === 'TELEMETRY') {
+                    // TODO: TELEMETRY
```
jfeingold35, line 169: Right now, there's no layer in the system that can do anything productive with Telemetry events. So my instinct is to just `TODO` it for now, and we can add it in the next Sprint.

packages/code-analyzer-sfge-engine/src/sfge-wrapper.ts
```diff
@@ -54,10 +96,121 @@ export class RuntimeSfgeWrapper {
...
+        }
+    }
+}
+
+function processMessageFromFailedRun(errMsg: string): string {
+    const indexOfErrorDelimiter: number = errMsg.indexOf(SFGE_ERROR_START);
+    return indexOfErrorDelimiter > 0
+        // If the error message includes our error delimiter string, then we should pull out that error
+        // and wrap it in something slightly easier for the user to read.
+        ? getMessage('error_external_sfgeIncompleteAnalysis', errMsg.slice(indexOfErrorDelimiter + SFGE_ERROR_START.length).trim())
```
jfeingold35, line 188: Side-effect: Since the Start token is on a different line than the error itself, the leading `| ` is kept.
So you get something like `| Blah blah blah` where v4 would produce just `Blah blah blah`. To me, it's not the worst thing in the world, but I understand if we want to try and fix that.

packages/code-analyzer-sfge-engine/src/sfge-wrapper.ts
```diff
@@ -54,10 +96,121 @@ export class RuntimeSfgeWrapper {
...
+        // Otherwise, keep the error message styled as it came back from the Java Command Executor.
+        : /* istanbul ignore next */ errMsg;
```
jfeingold35, line 190: Couldn't make this happen in a test scenario without a psychotic amount of stubbing.

packages/code-analyzer-sfge-engine/src/sfge-wrapper.ts
```diff
@@ -54,10 +96,121 @@ export class RuntimeSfgeWrapper {
...
+            const processedMessage = getMessage(sfgeMessage.messageKey, ...sfgeMessage.args);
+            if (isSfgeLogMessage(sfgeMessage)) {
+                if (sfgeMessage.messageSeverity === 'TELEMETRY') {
+                    // TODO: TELEMETRY
+                } else {
+                    emitLog(sfgeLogLevelToSfcaLogLevel(sfgeMessage.messageSeverity), processedMessage);
+                }
+            } else if (isSfgeProgressMessage(sfgeMessage)) {
+                const completionPercent: number = sfgeMessage.progressPercent;
+                emitProgress(10 + 85 * completionPercent / 100, processedMessage); // 10%-95%
```
jfeingold35, line 177: SFGE plugin passes progress event messages up from SFGE to the Engine API layer.

packages/code-analyzer-sfge-engine/sfge/src/test/resources/log4j2.xml
```diff
@@ -4,21 +4,12 @@
         <Console name="LogToConsole" target="SYSTEM_OUT">
             <PatternLayout pattern="%d{yyyy-MM-dd HH:mm:ss} %X{ruleRunId} %-5p %c{1}:%L - %m%n"/>
         </Console>
-        <RollingFile name="LogToFile" fileName="sfge-junit.log" filePattern="sfge-junit-%d{MM-dd-yyyy}-%i.log.gz">
```
jfeingold35, line 7: This should prevent test execution from creating log files. If we need to, we can add it back.

packages/code-analyzer-sfge-engine/.gitignore
stephen-carter-at-sf: For both log and test gz files.. I'd prefer to simply not produce any local files here. Tests shouldn't be producing artifacts anywhere other than in temporary directories. And where are the log files coming from.

Is there a workaround in the code to prevent these files from being produced locally?

---
I see below you made some changes so these shouldn't be produced any more. If that's true, then please just remove the `**/*.log` and `**/*.gz` from this .gitignore file

packages/code-analyzer-sfge-engine/src/engine.ts
```diff
@@ -58,10 +63,71 @@ export class SfgeEngine extends Engine {
...
+        this.emitRunRulesProgressEvent(2);
+
+        if (selectedRuleNames.length === 0) {
+            this.emitRunRulesProgressEvent(100);
+            return { violations: [] };
+        }
+
+        await this.validateWorkspaceCompleteness(runOptions.workspace);
```
stephen-carter-at-sf, line 74: Instead of passing in the workspace here, instead go ahead and call 
const allRelevantFiles: string[] = await this.getRelevantFilesInWorkspace(workspace);
and then check if allRelevantFiles is empty, and if so then just do a quick return.

Otherwise, pass in relevant files to the validateWorkspaceCompleteness function instead of the workspace.

This way we don't call SFGE if there are no relevant files.


In fact, (as I'm thinking here)... we should be doing this in the describeRules method if the workspace is provided. So please update describeRules to not return any rules if there are no relevant files and add in a test for this. Then zero rules will be selected before even attempting to hit the runRules method.

packages/code-analyzer-sfge-engine/src/engine.ts
stephen-carter-at-sf: Eventually we'll want to enhance SFGE java side to return the entire path (possible with the middle trimmed if needed). And then we'd be giving all these code locations to the user just like we do with flowtest.

stephen-carter-at-sf: Is the primary location dependent on the type of rule that ran? If so, do we need another return from the JAVA side saying the type of rule it is so we can correctly set the primary location accurately to what we think is best?

packages/code-analyzer-sfge-engine/src/engine.ts
```diff
@@ -71,20 +137,42 @@ export class SfgeEngine extends Engine {
...
+                file: sfgeViolation.sinkFileName,
+                startLine: sfgeViolation.sinkLineNumber!,
+                startColumn: sfgeViolation.sinkColumnNumber!
+            });
+        }
+        return {
+            ruleName: sfgeViolation.ruleName,
+            message: sfgeViolation.message,
+            codeLocations,
+            primaryLocationIndex: 0
```
stephen-carter-at-sf, line 160: As I mentioned before... do we want to have this be the sink in some cases? What makes more sense?

packages/code-analyzer-sfge-engine/src/engine.ts
stephen-carter-at-sf: Note that I did in fact do this with the PMD describeRules... it's just inside of the helper method: https://github.com/forcedotcom/code-analyzer-core/blob/dev/packages/code-analyzer-pmd-engine/src/pmd-engine.ts#L125-L126

packages/code-analyzer-sfge-engine/src/messages.ts
```diff
@@ -13,12 +13,21 @@ const MESSAGE_CATALOG : { [key: string]: string } = {
     UnsupportedEngineName:
         `The SfgeEnginePlugin does not support an engine with the name '%s'.`,
 
+    WorkspaceAppearsIncomplete:
+        `Specified workspace is missing %d possibly-relevant file(s) in %s. SFGE results may be incomplete or incorrect.`,
```
stephen-carter-at-sf: I suggest we do more hand holding for the user, and explain what's going on here... eventually once we add in the path-start then we can give an example of showing to include all project files in the workspace while still targeting the scan using path-start.

Anyway, we should brainstorm here.

packages/code-analyzer-sfge-engine/src/sfge-wrapper.ts
```diff
@@ -54,10 +97,123 @@ export class RuntimeSfgeWrapper {
...
 
+    public async invokeRunCommand(selectedRuleInfos: SfgeRuleInfo[], workspace: Workspace, emitProgress: (percComplete: number) => void): Promise<SfgeRunResult[]> {
+        const tmpDir: string = await this.getTemporaryWorkingDir();
+        emitProgress(2);
+
+        const inputFileName: string = path.join(tmpDir, 'sfgeInput.json');
+        const logFilePath: string = path.join(os.tmpdir(), 'sfge.log');
+        const ruleNames: string[] = selectedRuleInfos.map(sri => sri.name);
+        const targets: string[] = await workspace.getExpandedFiles(); // TODO: When we add path-start-targeting, this needs to change.
+        const projectFolders: string[] = await workspace.getExpandedFiles();
```
stephen-carter-at-sf: I thought you are already reducing the expanded files to the relevant files in the engine.ts file. Shouldn't you be passing in that list of relevant files instead of using all the files in the workspace?

stephen-carter-at-sf (changes requested):

packages/code-analyzer-sfge-engine/src/engine.ts
jfeingold35: The primary location can honestly be whatever we say it is. In my opinion, it makes more sense for the primary location to always be the Source Vertex in a path-based violation, because that's the path start.

packages/code-analyzer-sfge-engine/.gitignore
jfeingold35: Yeah, I should be in the clear to remove this. I'll do that and see what happens.

packages/code-analyzer-sfge-engine/src/messages.ts
```diff
@@ -13,12 +13,21 @@ const MESSAGE_CATALOG : { [key: string]: string } = {
...
+        `SFGE execution logs being written to %s.`,
+
     debug_sfgeInfoLog:
         `%s`,
 
     info_sfgeMetaInfoCollected:
         `Loaded %s: [ %s ]`,
 
+    info_telemetry:
+        `%s`,
```
stephen-carter-at-sf: Ideally we wouldn't have just %s in our message catalog. The message catalog should be simple a place where in the future we can support multiple locale's and we wouldn't want a doc writer or a translator having to review these non-locale specific messages.

packages/code-analyzer-sfge-engine/src/engine.ts
jfeingold35: Per line 130, if `getRelevantFilesInWorkspace` returns 0, then an empty list of rules is returned and the wrapper's `invokeDescribeCommand` method is never invoked. This also leads to `runRules` exiting early at line 86.

packages/code-analyzer-sfge-engine/src/messages.ts
jfeingold35: Yeah, a brainstorm for this one is a good idea.

packages/code-analyzer-sfge-engine/src/sfge-wrapper.ts
jfeingold35: Yeah, that's a good idea.

packages/code-analyzer-sfge-engine/.gitignore
jfeingold35: Done.

packages/code-analyzer-sfge-engine/src/engine.ts
stephen-carter-at-sf: Ah yes... i missed that. Nice.

packages/code-analyzer-sfge-engine/src/engine.ts
stephen-carter-at-sf: OK... whatever you think is best.... since it'll be the place where in VS Code it shows the violation message.

I'll create a work item for VS Code so that if the number of code locations > 0 then we should put them within the diagnostic.

packages/code-analyzer-sfge-engine/src/messages.ts
jfeingold35: Yeah, I guess we don't actually need this key, since the `messageSeverity` is what denotes it as a Telemetry event. I'll remove it.

packages/code-analyzer-sfge-engine/src/messages.ts
stephen-carter-at-sf: For now, let's at least say
```
Specified workspace is missing %d possibly-relevant file(s) from the folder %s. Salesforce Graph Engine may be unable to create a complete graph of your project without all apex files included in your workspace. This may result in incomplete or incorrect results.
```

But maybe in the future, we add as a debug log line the list of files that we found that the user should consider including in their workspace. Feel free to pull in Juliet for her thoughts here.

stephen-carter-at-sf (changes requested):

stephen-carter-at-sf (approved):
Just update that one warning message and I think you are good to merge.

PR #241, opened 2025-03-26
NEW (SFGE) @W-17915898@ Updated logging messages for SFGE

jfeingold35:
This PR does the following:
- Adds a Debug-level log statement during the `describe` flow indicating where the SFGE log file is being written.
- Changes the equivalent Info-level log statement from the `run` flow to Debug-level.
- Causes the SFGE log file to have a dynamic name based on the current timestamp, instead of always using `sfge.log`.

### DIFF
--- a/packages/code-analyzer-sfge-engine/src/engine.ts
+++ b/packages/code-analyzer-sfge-engine/src/engine.ts
@@ -13,6 +13,7 @@ import {
     Workspace
 } from '@salesforce/code-analyzer-engine-api';
 import {JavaCommandExecutor} from '@salesforce/code-analyzer-engine-api/utils';
+import {Clock} from './utils';
 import {getMessage} from './messages';
 import {RuntimeSfgeWrapper, SfgeRuleInfo, SfgeRunResult} from "./sfge-wrapper";
 import {SfgeEngineConfig} from "./config";
@@ -29,11 +30,11 @@ export class SfgeEngine extends Engine {
     private sfgeRuleInfoListCache: Map<string, SfgeRuleInfo[]> = new Map();
     private relevantFilesByWorkspaceId: Map<string, string[]> = new Map();
 
-    public constructor(config: SfgeEngineConfig) {
+    public constructor(config: SfgeEngineConfig, clock: Clock) {
         super();
         // TODO: When we support custom Java commands, we'll need to use the config property instead of the hardcoded string here.
         const javaCommandExecutor: JavaCommandExecutor = new JavaCommandExecutor('java', this.emitLogEvent.bind(this));
-        this.sfgeWrapper = new RuntimeSfgeWrapper(javaCommandExecutor, this.emitLogEvent.bind(this));
+        this.sfgeWrapper = new RuntimeSfgeWrapper(javaCommandExecutor, clock, this.emitLogEvent.bind(this));
         this.config = config;
     }
 
@@ -51,7 +52,7 @@ export class SfgeEngine extends Engine {
         this.emitDescribeRulesProgressEvent(5);
 
         const ruleInfoList: SfgeRuleInfo[] = await this.getSfgeRuleInfoList(
-            describeOptions.workspace,
+            describeOptions,
             (innerPerc: number) => this.emitDescribeRulesProgressEvent(5 + (90*innerPerc/100)) // 5%-95%
         );
 
@@ -72,7 +73,7 @@ export class SfgeEngine extends Engine {
         await this.validateWorkspaceCompleteness(runOptions.workspace);
 
         const allRulesInfoList: SfgeRuleInfo[] = await this.getSfgeRuleInfoList(
-            runOptions.workspace,
+            runOptions,
             (innerPerc: number) => this.emitRunRulesProgressEvent(2 + 3*(innerPerc/100)) // 2%-5%
         );
 
@@ -90,6 +91,7 @@ export class SfgeEngine extends Engine {
             selectedRuleInfoList,
             relevantFiles, // TODO: WHEN WE ADD PATH-START TARGETING, THIS NEEDS TO CHANGE.
             relevantFiles,
+            runOptions.logFolder,
             (innerPerc: number, message?: string) => this.emitRunRulesProgressEvent(5 + 93*innerPerc/100, message) // 5%-98%
         );
 
@@ -125,13 +127,15 @@ export class SfgeEngine extends Engine {
         }
     }
 
-    private async getSfgeRuleInfoList(workspace: Workspace|undefined, emitProgress: (percComplete: number) => void): Promise<SfgeRuleInfo[]> {
+    private async getSfgeRuleInfoList(options: DescribeOptions|RunOptions, emitProgress: (percComplete: number) => void): Promise<SfgeRuleInfo[]> {
+        const workspace: Workspace|undefined = options.workspace;
+        const logFolder: string = options.logFolder;
         const cacheKey: string = getCacheKey(workspace);
         if (!this.sfgeRuleInfoListCache.has(cacheKey)) {
             if (workspace && (await this.getRelevantFilesInWorkspace(workspace)).length === 0) {
                 this.sfgeRuleInfoListCache.set(cacheKey, []);
             } else {
-                const ruleInfoList: SfgeRuleInfo[] = await this.sfgeWrapper.invokeDescribeCommand(emitProgress);
+                const ruleInfoList: SfgeRuleInfo[] = await this.sfgeWrapper.invokeDescribeCommand(emitProgress, logFolder);
                 this.sfgeRuleInfoListCache.set(cacheKey, ruleInfoList);
             }
         }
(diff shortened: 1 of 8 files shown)


packages/code-analyzer-sfge-engine/src/utils.ts
```diff
@@ -0,0 +1,20 @@
+export interface Clock {
```
jfeingold35, line 1: At this point, I'm extremely tempted to pull all of the `Clock`-related code up into the `code-analyzer-engine-api/utils` module, so we don't have to keep declaring it over and over. I could do that as part of this PR. Thoughts?

packages/code-analyzer-sfge-engine/src/sfge-wrapper.ts
```diff
@@ -64,18 +64,21 @@ const SFGE_ERROR_START: string = 'SfgeErrorStart';
...
 export class RuntimeSfgeWrapper {
     private readonly javaCommandExecutor: JavaCommandExecutor;
+    private readonly logFileName: string;
     private temporaryWorkingDir?: string;
     private readonly emitLogEvent: (logLeveL: LogLevel, message: string) => void;
 
-    public constructor(javaCommandExecutor: JavaCommandExecutor, emitLogEvent: (logLevel: LogLevel, message: string) => void) {
+    public constructor(javaCommandExecutor: JavaCommandExecutor, clock: Clock, emitLogEvent: (logLevel: LogLevel, message: string) => void) {
         this.javaCommandExecutor = javaCommandExecutor;
+        this.logFileName = `sfge-${formatToDateTimeString(clock.now())}.log`;
```
jfeingold35: LogFileName is derived when the wrapper is constructed, so that both the Describe and Run flows use the same log file.

packages/code-analyzer-sfge-engine/src/sfge-wrapper.ts
```diff
@@ -97,17 +100,17 @@ export class RuntimeSfgeWrapper {
...
 
         const inputFileName: string = path.join(tmpDir, 'sfgeInput.json');
-        const logFilePath: string = path.join(os.tmpdir(), 'sfge.log');
+        const logFilePath: string = path.join(logFolder, this.logFileName);
         const ruleNames: string[] = selectedRuleInfos.map(sri => sri.name);
 
         await this.createSfgeInputFile(inputFileName, ruleNames, targetPaths, projectFilePaths);
         const resultsOutputFile: string = path.join(tmpDir, 'resultsFile.json');
-        this.emitLogEvent(LogLevel.Info, getMessage('LoggingToFile', logFilePath));
+        this.emitLogEvent(LogLevel.Debug, getMessage('LoggingToFile', 'run', logFilePath));
```
jfeingold35, line 113: Changed from `info` to `debug` so that it doesn't show up in the CLI output.

packages/code-analyzer-sfge-engine/src/sfge-wrapper.ts
```diff
@@ -64,18 +64,21 @@ const SFGE_ERROR_START: string = 'SfgeErrorStart';
...
         this.emitLogEvent = emitLogEvent;
     }
 
-    public async invokeDescribeCommand(emitProgress: (percComplete: number) => void): Promise<SfgeRuleInfo[]> {
+    public async invokeDescribeCommand(emitProgress: (percComplete: number) => void, logFolder: string): Promise<SfgeRuleInfo[]> {
         const tmpDir: string = await this.getTemporaryWorkingDir();
-        const logFilePath: string = path.join(os.tmpdir(), 'sfge.log');
+        const logFilePath: string = path.join(logFolder, this.logFileName);
         const sfgeRulesOutputFile: string = path.join(tmpDir, 'ruleInfo.json');
+        this.emitLogEvent(LogLevel.Debug, getMessage('LoggingToFile', 'describe', logFilePath));
```
jfeingold35, line 81: New debug log line.

packages/code-analyzer-sfge-engine/src/utils.ts
stephen-carter-at-sf: Yeah I'm fine with this. If you do this, I recommend switching the interface Clock to be an abstract clock with a "toDateTimeString()" method that uses the clock's now() function. This way we wouldn't need to have the format function be a separate function - but make it a first class citizen of clock.

packages/code-analyzer-sfge-engine/src/engine.ts
```diff
@@ -29,11 +30,11 @@ export class SfgeEngine extends Engine {
     private sfgeRuleInfoListCache: Map<string, SfgeRuleInfo[]> = new Map();
     private relevantFilesByWorkspaceId: Map<string, string[]> = new Map();
 
-    public constructor(config: SfgeEngineConfig) {
+    public constructor(config: SfgeEngineConfig, clock: Clock) {
```
stephen-carter-at-sf: I recommend making clock have a default value of 
clock: Clock = new RealClock()

packages/code-analyzer-sfge-engine/src/messages.ts
```diff
@@ -17,7 +17,7 @@ const MESSAGE_CATALOG : { [key: string]: string } = {
         `Specified workspace is missing %d possibly-relevant file(s) from the folder %s. Salesforce Graph Engine may be unable to create a complete graph of your project without all apex files included in your workspace. This may result in incomplete or incorrect results.`,
 
     LoggingToFile:
-        `SFGE execution logs being written to %s.`,
+        `Invoking SFGE %s flow. Logs being written to %s.`,
```
stephen-carter-at-sf: The word "flow" is overloaded. Instead maybe say "Invoking SFGE's %s method." or Invoking SFGE's %s command." instead.

packages/code-analyzer-sfge-engine/src/plugin.ts
```diff
@@ -35,12 +36,12 @@ export class SfgeEnginePlugin extends EnginePluginV1 {
 
     public override async createEngine(engineName: string, resolvedConfig: ConfigObject): Promise<Engine> {
         validateEngineName(engineName);
-        return new SfgeEngine(resolvedConfig as SfgeEngineConfig);
+        return new SfgeEngine(resolvedConfig as SfgeEngineConfig, new RealClock());
```
stephen-carter-at-sf: Could avoid touching this file if new RealClock() was default value of SfgeEngine's constructor.

packages/code-analyzer-sfge-engine/src/sfge-wrapper.ts
stephen-carter-at-sf: Hmmm... although this sounds like a good idea, I wonder about when we have VSCode up and running and the engines are all instantiated. Since this wrapper seems to be tied to the constructor of the SFGE engine's constructor... I think if the engine is reused then each run will keep appending to the same file. I don't think we want this... do we? IDK, maybe we do. But right now the flow engine always creates a new log file each time its engine is ran. I think we should probably do the same here for consistency.

Having one file for run and one file for describe may actually be a good idea. Maybe have the sfge-describe- prefix for the describe command call and the sfge-run- prefix for the run command call.

Thoughts?

stephen-carter-at-sf: Can we prefix everything with 'sfca-` like we do with flowtest.

I think the pattern we'll want will be ```sfca-<engine>-<dateTime>.log``` or ```sfca-<engine>-<topic>-<dateTime>.log``` if we decide to have separate run and describe log files.

jfeingold35: So `sfca--sfge.....log`? Yeah, we can do that.

packages/code-analyzer-sfge-engine/src/plugin.ts
jfeingold35: Good point.

packages/code-analyzer-sfge-engine/src/messages.ts
jfeingold35: Fair enough.

packages/code-analyzer-sfge-engine/src/utils.ts
jfeingold35: Will do.

packages/code-analyzer-sfge-engine/src/sfge-wrapper.ts
jfeingold35: V4's SFGE already keeps appending to the same file forever, so there wouldn't be a massive behavioral change here from that perspective. I think that might need to be part of the broader discussion about the future of SFGE in VSCode.
We _could_ use separate files for `describe` and `run`, but the question becomes whether we'd want to use the same timestamp for the filename.

stephen-carter-at-sf: I'm saying that in VSCode we are probably going initialize the engines for the duration of the window being opened. The question then becomes whether we want CodeAnalyzer.runRules will produce separate log files for each run or to just have 1 log file for all runs that occur.

So this will kind of set the precedence for that.

Currently the flowtest engine uses different files per run  - because it doesn't like appending currently.

So what approach do you think we should go with for the future (since we can always ask Robert for an enhancement to allow his tool to append to an existing log file instead of having to have a fresh one each time)?

stephen-carter-at-sf: Sorry - fixed my comment to make it clear what I was trying to say.

stephen-carter-at-sf (approved):

PR #246, opened 2025-03-31
NEW (SFGE) @W-17915919@ java_max_heap_size accepts trailing b's now.

jfeingold35:


packages/code-analyzer-sfge-engine/src/config.ts
```diff
@@ -44,15 +44,19 @@ export const SFGE_ENGINE_CONFIG_DESCRIPTION: ConfigDescription = {
             valueType: "boolean",
             defaultValue: DEFAULT_SFGE_ENGINE_CONFIG.disable_limit_reached_violations
         },
+        // Indicates the specific 'java' command associated with the JRE or JDK to use for the 'sfge' engine.
```
jfeingold35, line 47: Missed this comment in the initial PR. Adding it now.

packages/code-analyzer-sfge-engine/src/sfge-wrapper.ts
```diff
@@ -128,7 +128,12 @@ export class RuntimeSfgeWrapper {
             javaCmdArgs.push('-DSFGE_PATH_EXPANSION_LIMIT=-1');
         }
         if (sfgeRunOptions.heapSizeArg) {
-            javaCmdArgs.push(`-Xmx${sfgeRunOptions.heapSizeArg}`);
+            // We technically allow a trailing `b` in the heap size arg so that users can write stuff like "2mb" instead
```
jfeingold35: We rip the trailing `b` off here instead of at config-generation time to avoid modifying the actual config value. That way, `sf code-analyzer config` won't erase a trailing `b`.

packages/code-analyzer-sfge-engine/test/plugin.test.ts
```diff
@@ -185,19 +185,27 @@ describe('SfgeEnginePlugin', () => {
...
-                {val: '2098176'}, // Un-suffixed numbers must be increments of 1024.
-                {val: '8193K'},   // Suffixed numbers can be any value, and suffix can be any case.
-                {val: '8193k'},   // Suffixed numbers can be any value, and suffix can be any case.
-                {val: '9M'},      // Suffixed numbers can be any value, and suffix can be any case.
-                {val: '9m'},      // Suffixed numbers can be any value, and suffix can be any case.
-                {val: '2G'},      // Suffixed numbers can be any value, and suffix can be any case.
-                {val: '2g'},      // Suffixed numbers can be any value, and suffix can be any case.
-                {val: '02g'}      // Leading zeroes are weird, but they're not invalid.
+                {val: '2097152'},  // Java needs >=2MB to run at all.
+                {val: '2097152b'}, // Java needs >=2MB to run at all.
```
jfeingold35: There's no reason you'd write `2097152b` instead of `2097152`, but writing a regex that accepted `2gb` while rejecting `2b` would have been an excessive amount of effort.
So I've taken the middle ground of allowing this without documenting it. Seems harmless enough.

packages/code-analyzer-sfge-engine/test/engine.test.ts
```diff
@@ -241,6 +241,7 @@ describe('SfgeEngine', () => {
         it.each([
             {prop: 'disable_limit_reached_violations', value: true, javaArg: '-DSFGE_PATH_EXPANSION_LIMIT=-1'},
             {prop: 'java_max_heap_size', value: '2g', javaArg: '-Xmx2g'},
+            {prop: 'java_max_heap_size', value: '2gb', javaArg: '-Xmx2g'},
```
jfeingold35: New test makes sure that the `b` gets properly pulled off before it's fed to Java.

packages/code-analyzer-sfge-engine/src/sfge-wrapper.ts
stephen-carter-at-sf: Oh I thought we were going to remove the "b" on line 171 of config.ts. If we do it over there we get the added benefit of normalizing the the output so that subsequent calls to the config command show the normalized output and then we don't have any processing to do here when apply the arguments.

packages/code-analyzer-sfge-engine/src/config.ts
```diff
@@ -165,7 +169,7 @@ class SfgeConfigValueExtractor {
             return undefined;
         }
 
-        if (javaMaxHeapSize.toLowerCase().endsWith('g')) {
+        if (javaMaxHeapSize.toLowerCase().endsWith('g') || javaMaxHeapSize.toLowerCase().endsWith('gb')) {
```
stephen-carter-at-sf: Consider normalizing first by just doing a:
```
javaMaxHeapSize = java.MaxHeapSize.toLowerCase().replace(/b$/, '');
```
then you doing need to worry about the b or casing later on in your checks... and then the return value will be normalized.

packages/code-analyzer-sfge-engine/src/sfge-wrapper.ts
jfeingold35: Do we want the `config` command to normalize the output, or do we want it to preserve it?

packages/code-analyzer-sfge-engine/src/config.ts
jfeingold35: Reasonable. Will do.

packages/code-analyzer-sfge-engine/src/sfge-wrapper.ts
stephen-carter-at-sf: The config command always normalizes the input in addition to validating it. You'll see I do this with ecmascript to javascript for example with the PMD config. But we don't need to go crazy and convert everything to mega bytes or anything.... but removing the 'b' is definitely reasonable.

jfeingold35: Okay, fair enough. I'll pull the trailing `b` off. Stand by for that.

stephen-carter-at-sf (approved):

PR #247, opened 2025-04-01
CHANGE (SFGE) @W-18097045@ SFGE rules use last location as primary location

jfeingold35:


packages/code-analyzer-sfge-engine/sfge/src/main/java/com/salesforce/rules/RemoveUnusedMethod.java
```diff
@@ -195,11 +195,6 @@ protected String getUrl() {
         return URL;
     }
 
-    @Override
```
jfeingold35, line 198: Disabling `RemoveUnusedMethod`, because it's unfinished and unpolished. We can re-enable it if we ever get around to polishing it.

packages/code-analyzer-sfge-engine/src/engine.ts
```diff
@@ -167,7 +167,7 @@ export class SfgeEngine extends Engine {
             ruleName: sfgeViolation.ruleName,
             message: sfgeViolation.message,
             codeLocations,
-            primaryLocationIndex: 0
+            primaryLocationIndex: codeLocations.length - 1
```
jfeingold35, line 169: After looking through the rules, honestly, it just makes sense for the sink to be the Primary Location for all of the rules. It's more likely to be (or be in proximity to) the place where code needs to change. So it's more valuable to anchor the violation there.

packages/code-analyzer-sfge-engine/test/engine.test.ts
```diff
@@ -167,9 +167,10 @@ describe('SfgeEngine', () => {
             expect(debugLogEvents.length).toBeGreaterThanOrEqual(2);
             expect(debugLogEvents[0].message).toEqual(`Invoking SFGE's describe command. Logs being written to ${path.join(os.tmpdir(), 'sfca-sfge-2025_03_21_12_30_25_020.log')}.`);
             expect(debugLogEvents[1].message).toEqual(`Invoking SFGE's run command. Logs being written to ${path.join(os.tmpdir(), 'sfca-sfge-2025_03_21_12_30_25_020.log')}.`);
-            expect(progressEvents.map(pe => pe.percentComplete)).toEqual(
-                [2, 2.3, 4.4, 4.7, 5, 6.86, 14.3, 22.21, 26.16, 34.06, 85.45, 93.35, 98, 100]
-            );
+            const progressPercents: number[] = progressEvents.map(pe => pe.percentComplete);
+            expect(progressPercents[0]).toEqual(2);
+            expect(progressPercents[progressPercents.length - 1]).toEqual(100);
+            expectProgressEventsToAscend(progressPercents);
```
jfeingold35, line 173: The progress events are too dependent on the number of entrypoints, rules, and other fiddly nonsense that testing for their exact values makes some of the tests a little flaky.
So I'm just testing that it starts at 2, ends at 100, and that every progress event is a higher number than the previous one.

packages/code-analyzer-sfge-engine/src/engine.ts
stephen-carter-at-sf: Yeah that's what I initially thought too. Thanks for thinking it through.

stephen-carter-at-sf (commented):
Did we want add a gold file test for all the SFGE rule descriptions? I anticipated when you removed the RemoveUnusedMethod rule that a gold file would have been updated to show this rule removed. If not, then can we add in one?

packages/code-analyzer-sfge-engine/test/test-data/sampleRelevantWorkspace/SomeClass.cls
```diff
@@ -1,10 +1,6 @@
 // This class is going to perform a SOQL query, so its lack of `with sharing` is a violation of an SFGE rule.
 // It is `global` so that it can count as an entrypoint without necessarily needing to use `pathStart` functionality.
 global class SomeClass {
-    // This static block is here to violate RemoveUnusedMethod and test the fudging of synthetic locations.
```
jfeingold35, line 4: `RemoveUnusedMethod` was the only rule that actually tripped the behavior for fudging bad locations. So I remove this block as a consequence of disabling the rule.

packages/code-analyzer-sfge-engine/test/test-data/goldfiles/all_sampleRelevantWorkspace_violations.goldfile.json
jfeingold35, line 1: Changes look more complicated than they are. I removed the `RemoveUnusedMethod` violation, changed the lines for the other violations, and added a new one for `UnimplementedTypeRule`.

packages/code-analyzer-sfge-engine/test/test-data/goldfiles/all_rules.goldfile.json
```diff
@@ -54,18 +54,6 @@
     "description": "This rule is in \"Developer Preview\" and is subject to change. Identifies SOQL queries with variables in WHERE clauses that lack null checks.",
     "resourceUrls": []
   },
-  {
```
jfeingold35, line 57: Removed the rule; removed the violation.

packages/code-analyzer-sfge-engine/src/engine.ts
```diff
@@ -179,38 +178,6 @@ export class SfgeEngine extends Engine {
         }
         return this.relevantFilesByWorkspaceId.get(workspace.getWorkspaceId())!;
     }
-
-    private adjustCodeLocationsIfNeeded(unadjustedViolation: Violation): Violation {
```
jfeingold35, line 183: `RemoveUnusedMethod` was the only case we're aware of that tripped this rule. So I'm removing the code for now in order to avoid having uncovered/dead code in the plugin, and if the issue resurfaces then we can re-add with that scenario.

packages/code-analyzer-sfge-engine/src/messages.ts
```diff
@@ -70,11 +70,6 @@ const MESSAGE_CATALOG : { [key: string]: string } = {
     LoggingToFile:
         `Invoking SFGE's %s command. Logs being written to %s.`,
 
-    ViolationLocationFudged:
```
jfeingold35, line 73: Removed the behavior that necessitated this message; removed the message.

stephen-carter-at-sf (approved):

PR #250, opened 2025-04-03
NEW (SFGE) @W-17915877@ PathStart functionality for SFGE

jfeingold35:
This PR does the following:
- Allows the Java layer to fully handle a `projectDirs` property consisting of a list of individual files.
- Adds support for the `pathStart` property on `runOptions` in SFGE.


packages/code-analyzer-sfge-engine/sfge/src/main/java/com/salesforce/metainfo/AbstractMetaInfoCollector.java
```diff
@@ -61,13 +63,21 @@ public TreeSet<String> getMetaInfoCollected() {
...
+            // files are in a sibling. So we'll go up a level before walking the file tree.
             path = path.getParent();
+        } else if (isApexFile(path)) {
+            // If the path itself is an apex file, we should assume that it is contained in the class folder, and that
+            // project files are in a sibling of its parent directory. So we'll go up two levels before walking the file
+            // tree.
+            path = path.getParent().getParent();
         }
         final ProjectFileVisitor projectFileVisitor = new ProjectFileVisitor();
+        if (this.pathsWalked.contains(path.toString())) {
```
jfeingold35, line 77: This matters because when the sourceFolder paths are lists of individual files in the same folder, then `path` will end up being the same for all of them and we don't want to walk the same path trees repeatedly.

packages/code-analyzer-sfge-engine/sfge/src/main/java/com/salesforce/metainfo/AbstractMetaInfoCollector.java
```diff
@@ -61,13 +63,21 @@ public TreeSet<String> getMetaInfoCollected() {
 
     private void processSourceFolder(String sourceFolder) throws MetaInfoLoadException {
```
jfeingold35, line 64: Technically, `sourceFolders` can now be a list of files.
But I didn't want to chase down and rename a million different variables, so I didn't do that. I _can_ if people want, but I think I'd prefer to do it in a follow-up PR to keep this one readable.

packages/code-analyzer-sfge-engine/src/engine.ts
```diff
@@ -31,10 +31,9 @@ export class SfgeEngine extends Engine {
 
     public constructor(config: SfgeEngineConfig, clock: Clock = new RealClock()) {
         super();
-        // TODO: When we support custom Java commands, we'll need to use the config property instead of the hardcoded string here.
-        const javaCommandExecutor: JavaCommandExecutor = new JavaCommandExecutor('java', this.emitLogEvent.bind(this));
-        this.sfgeWrapper = new RuntimeSfgeWrapper(javaCommandExecutor, clock, this.emitLogEvent.bind(this));
         this.config = config;
+        const javaCommandExecutor: JavaCommandExecutor = new JavaCommandExecutor(this.config.java_command, this.emitLogEvent.bind(this));
```
jfeingold35, line 35: Forgot to do this the first time around. Doing it now.

packages/code-analyzer-sfge-engine/sfge/src/main/java/com/salesforce/metainfo/AbstractMetaInfoCollector.java
randi274: I don't mind it as a follow-up! Sounds like a good cleanup item 👍

packages/code-analyzer-sfge-engine/sfge/src/main/java/com/salesforce/metainfo/AbstractMetaInfoCollector.java
```diff
@@ -61,13 +63,21 @@ public TreeSet<String> getMetaInfoCollected() {
...
         Path path = new File(sourceFolder).toPath();
-        // If the directory has any apex files, we should assume it's the class folder and that
-        // project files are in a sibling.
-        // So we'll go up a level before walking the file tree.
-        if (directoryContainsApex(path)) {
+        if (isDirectoryContainingApex(path)) {
+            // If the path is a directory with apex files in it, we should assume it's the class folder, and that project
```
randi274, line 67:
```suggestion
            // If the path is a directory with apex files in it, we should assume it's the classes folder, and that project
```

packages/code-analyzer-sfge-engine/sfge/src/main/java/com/salesforce/metainfo/AbstractMetaInfoCollector.java
```diff
@@ -61,13 +63,21 @@ public TreeSet<String> getMetaInfoCollected() {
...
+            // files are in a sibling. So we'll go up a level before walking the file tree.
             path = path.getParent();
+        } else if (isApexFile(path)) {
+            // If the path itself is an apex file, we should assume that it is contained in the class folder, and that
```
randi274, line 71:
```suggestion
            // If the path itself is an apex file, we should assume that it is contained in the classes folder, and that
```

packages/code-analyzer-sfge-engine/sfge/src/main/java/com/salesforce/metainfo/AbstractMetaInfoCollector.java
```diff
@@ -61,13 +63,21 @@ public TreeSet<String> getMetaInfoCollected() {
...
+            // project files are in a sibling of its parent directory. So we'll go up two levels before walking the file
```
randi274, line 72: Can you clarify for me what you mean by "project files" here?

packages/code-analyzer-sfge-engine/sfge/src/main/java/com/salesforce/metainfo/AbstractMetaInfoCollector.java
randi274: Actually in reading through the code, I think `sourceFolder` for the parameter is still right, isn't it? Later I see this, and the invocation reads right to me:

```
for (String sourceFolder : sourceFolders) {
       processSourceFolder(sourceFolder);
}
```

randi274 (commented):
Just reading through for understanding. Looks like a simple enough lift though!

PR #251, opened 2025-04-04
CHANGE: @W-18201387@ Template changes to support scaffolding

randi274:
Creates an ENGINE-TEMPLATE to make the scaffolding work easy to support the addition of the stylelint engine. This will also help to accelerate other teams (and ourselves) in adding engines to the Code Analyzer core library.

Ensures test coverage of the template also begins at 100%:

Screenshot 2025-04-07 at 1 27 18 PM


packages/ENGINE-TEMPLATE/README.md
```diff
@@ -0,0 +1,48 @@
...
+1. Copy This Template
+
+Copy this entire template folder, and rename the directory to `code-analyzer-x-engine`, where `x` will be the name of your engine. For example, the `code-analyzer-sfge-engine` has `sfge` as the engine name for the capabilities referred to as the "Salesforce Graph Engine". Engine names should be lowercased and succinct. Engine names are subject to Salesforce Code Analyzer team approval.
+
+2. Add Dependencies
+
+Add any dependencies you might need for external rule retrieval or other functionality. Important to note in the included dependencies is the code analyzer engine api, which is also published from this monorepo:
+
+```
+"@salesforce/code-analyzer-engine-api": "0.21.0-SNAPSHOT"
```
stephen-carter-at-sf: To avoid confusion, can we just say something like:
```
"@salesforce/code-analyzer-engine-api": "<MUST_MATCH_ENGINE_API_VERSION>"
```
or something like that. Basically they must always have the version equal to the engine api's package.json version number.

packages/ENGINE-TEMPLATE/README.md
```diff
@@ -0,0 +1,48 @@
...
+```
+
+Use the value provided in the template today, and do not try to modify it. It may say "SNAPSHOT" which is what indicates a new version will be available with our next release, but the value should match whatever the template uses, and will keep this up to date.
+
+3. Create Your Engine
+
+Open the [engine](src/engine.ts) file. Set your engine name, which should match the name of `x` above. Follow the rest of the instructions in the file.
+
+4. Create Your Plugin Provider
```
stephen-carter-at-sf: Maybe say:
```
4. Create Your Plugin (which acts as an Factory or Provider of one or more engines)
```

packages/ENGINE-TEMPLATE/README.md
```diff
@@ -0,0 +1,48 @@
...
+
+Open the [plugin.ts](src/plugin.ts) file. This currently extends EnginePluginV1, managed from the Engine API. Set your available engine name, which should match the name of `x` above (thought there is the option of adding multiple engines within your plugin as well).
+
+5. Export Your Plugin
+
+Update the [index.ts](src/index.ts) to update your imports. This will allow for dynamic JIT loading when users decide which engines they want to use to scan their projects.
```
stephen-carter-at-sf: Maybe be more explicit:
```
Update the [index.ts](src/index.ts) to export your plugin and a `createEnginePlugin` function which returns an instance of your plugin.
```

packages/ENGINE-TEMPLATE/README.md
```diff
@@ -0,0 +1,48 @@
...
+
+6. Add Configuration (Optional)
+
+With v5 of Code Analyzer, users are able to add engine-specific [configuration](https://developer.salesforce.com/docs/platform/salesforce-code-analyzer/guide/config-custom.html). If your engine would like to have specific configuration details, add this via a `config.ts` file, and consume it in the Engine constructor.
```
stephen-carter-at-sf, line 36: This is good enough for now. We might expand on this later.

packages/ENGINE-TEMPLATE/package.json
```diff
@@ -0,0 +1,63 @@
...
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/forcedotcom/code-analyzer-core.git",
+    "directory": "packages/ENGINE-TEMPLATE"
+  },
+  "main": "dist/index.js",
+  "types": "dist/index.d.ts",
+  "dependencies": {
+    "@types/node": "^20.0.0",
+    "@salesforce/code-analyzer-engine-api": "0.21.0-SNAPSHOT"
```
stephen-carter-at-sf, line 17: I believe that the build script will need to be updated that ensures that this is always up to date. Josh I believe updates the T-E-M-P-L-A-T-E's package.json file with our github workflows... so go take a look to see if it'll automatically pick this up or not.

packages/ENGINE-TEMPLATE/src/engine.ts
```diff
@@ -0,0 +1,61 @@
...
+
+    // *** Update if you want to get your engine version any other way
+    public async getEngineVersion(): Promise<string> {
+        const pathToPackageJson: string = path.join(__dirname, '..', 'package.json');
+        const packageJson: {version: string} = JSON.parse(await fsp.readFile(pathToPackageJson, 'utf-8'));
+        return packageJson.version;
+    }
+
+    // *** Remove underscore for private naming convention if you need any of the DescribeOptions
+    async describeRules(_describeOptions: DescribeOptions): Promise<RuleDescription[]> {
```
stephen-carter-at-sf, line 28: Let's add emit at least one emitDescribeRulesProgressEvent progress event here as well as an example.

packages/ENGINE-TEMPLATE/src/engine.ts
```diff
@@ -0,0 +1,61 @@
...
+        // *** If Violations do not have CodeLocations they will be shown in modal form
+
+        // *** Best Practice - Use messages sparingly to keep users informed
+        // *** Ideal for failures, or an announcement
+        const one = '1'
+        this.emitLogEvent(LogLevel.Debug, getMessage('TemplateMessage2', one, '2'));
+
+        this.emitRunRulesProgressEvent(100);
+        return {
+            violations
```
stephen-carter-at-sf: I know some users get confused by this syntax, so since this is in a template, let's use:
```
return {
    violations: violations
}
```

packages/ENGINE-TEMPLATE/src/engine.ts
```diff
@@ -0,0 +1,61 @@
...
+import * as fsp from 'node:fs/promises';
+import path from "path";
+import { getMessage } from "./messages";
+
+// *** Change the Class Name to match your engine's name
+export class TemplateEngine extends Engine {
+    // *** Change the NAME to match your engine's name
+    static readonly NAME = "template";
+
+    // *** Update by setting config settings, if user-configuration is desired
```
stephen-carter-at-sf: This could be made more clear... maybe say:
```
// *** Consider passing in a configuration object from your engine's plugin if you want to provide user-configuration
```

packages/ENGINE-TEMPLATE/src/plugin.ts
```diff
@@ -0,0 +1,32 @@
+import {
+    ConfigObject,
+    Engine,
+    EnginePluginV1,
+} from "@salesforce/code-analyzer-engine-api";
+import {getMessage} from "./messages";
+import { TemplateEngine } from "./engine";
+
+// *** Change the Class Name to match your engine's name
+export class TemplateEnginePlugin extends EnginePluginV1 {
```
stephen-carter-at-sf, line 10: Let's add in a comment that says that if they want to provide config, then they could also override the other 2 methods.

packages/ENGINE-TEMPLATE/test/temp.test.ts
```diff
@@ -0,0 +1,12 @@
...
+
+changeWorkingDirectoryToPackageRoot();
+
+describe('Temporary test', () => {
+    it('Temporary test', () => {
+        // This is just a temporary assertion to get code coverage
+        expect(getMessage('TemplateMessage2', 'hello', 3)).toEqual(
+            `This message has two variables, 'hello' and 3, in its message.`);
+    });
+});
```
stephen-carter-at-sf: This is ok for now, but we might want to start with a engine.test.ts and plugin.test.ts file with some boiler plate tests.

README.md
```diff
@@ -1,7 +1,11 @@
 # Salesforce Code Analyzer - Core Module
 
-The code in repository is currently for internal use only while we build the next generate Salesforce Code Analyzer.
+The code in repository is currently for internal use only while we build the next Salesforce Code Analyzer.
 
-See instead:
+See our external products:
 * [sfdx-scanner](https://github.com/forcedotcom/sfdx-scanner)
```
kenamverma, line 6: We also have cli for CA, is it sfdx-scanner?

randi274: We currently support 2 CLI plugins, scanner (v4) and code-analyzer (v5). At the end of the month we'll be deprecating v4 and moving exclusively to the v5 commands, so it would be good to be familiar with just those: https://developer.salesforce.com/docs/platform/salesforce-code-analyzer/guide/analyze.html

packages/ENGINE-TEMPLATE/package.json
```diff
@@ -0,0 +1,63 @@
...
+  },
+  "devDependencies": {
+    "@eslint/js": "^8.57.1",
+    "@types/jest": "^29.5.14",
+    "eslint": "^8.57.1",
+    "jest": "^29.7.0",
+    "rimraf": "*",
+    "ts-jest": "29.2.3",
+    "typescript": "^5.7.3",
+    "typescript-eslint": "^8.22.0"
```
kenamverma, line 27: So, these dependencies are default and will always be present. While creating a new engine, we will just add dependencies on top of this?

randi274: Yes, if you have any dependencies you need for development, they would go here. If you need dependencies to ship with the package, they will go in `dependencies`. Each of these plugins is published as a separate npm package so you can't rely on any top-level dependencies in this monorepo.

packages/ENGINE-TEMPLATE/test/temp.test.ts
randi274: Good call! I added those as well.

packages/ENGINE-TEMPLATE/package.json
randi274: @jfeingold35 after reviewing the workflows, I *think* this gets picked up automatically thanks to its inclusion in the `package-lock.json` as a linked package, and with the `--workspaces` flag on the build task. Is that correct? Or do I need to do anything else to ensure it gets picked up? This is not something we want to publish, same as the original template. (The actual engine that gets created from the template I know will need github action updates.)

packages/ENGINE-TEMPLATE/test/temp.test.ts
stephen-carter-at-sf: OK this file is no longer needed then. Let's delete it.

stephen-carter-at-sf (approved):
Just delete the dummy temp test file and I think we are good. Thanks for doing this.

Also, double check to see if we have 100% coverage on this package.

packages/ENGINE-TEMPLATE/package.json
jfeingold35: I'm slightly confused about the question. Maybe we can clear this up during scrum?

PR #252, opened 2025-04-06
NEW: @W-18210740@: Rework Workspace object to include target files in addition to still having all workspace files

stephen-carter-at-sf:
* Updated Workspace objects (for both core and engine-api) to have new contract that supplies all workspace files and the targeted files within the workspace.
** Optimized the methods to remove redundant information. ~For example if a user targets both the file and a method within the file, then the file wins and we can safely remove the method level target.~
* Updated all engines to appropriate targeted files, with the exception of the SFGE. Note that Flow Scanner will need updating too.
* ~Added info level log message from 'pmd', 'cpd', and 'regex' engines if at least one rule runs and a method level target was provided to inform users that the method level target will be ignored.~


packages/code-analyzer-core/src/code-analyzer.ts
```diff
@@ -27,31 +27,62 @@ import path from 'node:path';
...
+    /**
+     * Returns a list of targeted methods, separate from the targeted files and folders, that an engine may analyse.
+     *
+     * Not all engines will be able to target individual methods. If an engine does support the file associated with the
+     * targeted method but does not support method level targeting, then the engine may emit an info or warning log
+     * event when it is running to say that the method level target is being ignored.
+     *
+     * The format of each targeted method is `<filePath>#<methodName>` (ex: '/path/to/SomeApexFile.cls#SomeMethod').
+     */
+    getTargetedMethods(): Promise<string[]>
```
stephen-carter-at-sf: A big part of me would rather not support method level targeting. Waiting for John to see if we can either just hide this or remove support for it altogether. I'd hate to ship something that is only used be 1 engine that is in Dev Preview and that most other engines just ignore.

stephen-carter-at-sf: OK, after looking at some data, John agreed that we should probably just wait for a user to ask about method level targeting before we add this in to code analyzer v5. I added a commit to remove the support for all this.

packages/code-analyzer-engine-api/src/workspace.ts
```diff
@@ -55,25 +62,85 @@ export class Workspace {
...
+     * resulting list only contains file paths.
+     *
+     * Any files underneath the workspace root that Code Analyzer chooses to ignore (like .gitignore files, files in
+     * node_modules folders, etc.) are automatically excluded unless they were explicitly provided when constructing
+     * the workspace.
      */
-    async getExpandedFiles(): Promise<string[]> {
-        if (!this.expandedFiles) {
-            this.expandedFiles = (await expandToListAllFiles(this.getFilesAndFolders())).filter(f => !this.shouldExclude(f));
+    async getWorkspaceFiles(): Promise<string[]> {
```
jfeingold35, line 113: What was the motivation for removing the word "expanded" from the method name?

packages/code-analyzer-regex-engine/test/engine.test.ts
```diff
@@ -155,14 +156,14 @@ describe("Tests for RegexEngine's getName and describeRules methods", () => {
...
     it('if runRules() is called on a directory with no Apex files, it should correctly return no violations', async () => {
         const runOptions: RunOptions = createRunOptions(
-            new Workspace([path.resolve(__dirname, "test-data", "apexClassWhitespace", "1_notApexClassWithWhitespace")]));
+            new Workspace('id', [path.resolve(__dirname, "test-data", "apexClassWhitespace", "1_notApexClassWithWhitespace")]));
         const runResults: EngineRunResults = await engine.runRules( ["NoTrailingWhitespace"], runOptions);
         expect(runResults.violations).toHaveLength(0);
     });
 
-    it("Ensure runRules when called on a directory of Apex classes, it properly emits violations", async () => {
+    it("Ensure runRules when called on a workspapce that targets Apex classes, it properly emits violations", async () => {
```
jfeingold35: typo.

packages/code-analyzer-engine-api/src/workspace.ts
stephen-carter-at-sf: Well now we have 2 sets of expanded files... the expanded workspace files and expanded target files. I found this a little verbose. Since these methods will be used primarily over the raw inputs... I thought it better to make these names getWorkspaceFiles and getTargetedFiles short and instead add "Raw" to the other names.

packages/code-analyzer-regex-engine/test/engine.test.ts
stephen-carter-at-sf: Thanks. Fixed.

packages/code-analyzer-engine-api/src/workspace.ts
jfeingold35: Fair enough.

jfeingold35 (approved):

PR #254, opened 2025-04-07
NEW (SFGE) @W-17915877@ SFGE now supports targets

jfeingold35:
This PR does the following:
- Allows the Java layer to fully handle a `projectDirs` property consisting of a list of individual files.
- Adds support for the file targeting mechanism within the workspace.


packages/code-analyzer-sfge-engine/sfge/src/main/java/com/salesforce/metainfo/AbstractMetaInfoCollector.java
```diff
@@ -61,13 +63,21 @@ public TreeSet<String> getMetaInfoCollected() {
 
     private void processSourceFolder(String sourceFolder) throws MetaInfoLoadException {
```
jfeingold35, line 67: Technically, `sourceFolders` can be a list of files. But I didn't want to chase down and rename a million variables, so I didn't do that. If we want to do that later, we can do so.

packages/code-analyzer-sfge-engine/sfge/src/main/java/com/salesforce/metainfo/AbstractMetaInfoCollector.java
```diff
@@ -61,13 +63,21 @@ public TreeSet<String> getMetaInfoCollected() {
...
+            // files are in a sibling. So we'll go up a level before walking the file tree.
             path = path.getParent();
+        } else if (isApexFile(path)) {
+            // If the path itself is an apex file, we should assume that it is contained in the class folder, and that
+            // project files are in a sibling of its parent directory. So we'll go up two levels before walking the file
+            // tree.
+            path = path.getParent().getParent();
         }
         final ProjectFileVisitor projectFileVisitor = new ProjectFileVisitor();
+        if (this.pathsWalked.contains(path.toString())) {
```
jfeingold35, line 80: This matters because when the `sourceFolder` paths are lists of individual files in the same folder, then `path` will end up being the same for all of them, and we don't want to repeatedly walk the same tree.

packages/code-analyzer-sfge-engine/sfge/src/main/java/com/salesforce/metainfo/AbstractMetaInfoCollector.java
stephen-carter-at-sf: Can you at least put a comment in the code then with a "TODO" or something.

packages/code-analyzer-sfge-engine/sfge/src/main/java/com/salesforce/metainfo/AbstractMetaInfoCollector.java
```diff
@@ -61,13 +63,21 @@ public TreeSet<String> getMetaInfoCollected() {
...
-        if (directoryContainsApex(path)) {
+        if (isDirectoryContainingApex(path)) {
+            // If the path is a directory with apex files in it, we should assume it's the `classes` folder, and that project
+            // files like VF components or object metadata are in a sibling. So we'll go up a level before walking the file tree.
             path = path.getParent();
+        } else if (isApexFile(path)) {
+            // If the path itself is an apex file, we should assume that it is contained in the `classes` folder, and that
+            // project files like VF components or object metadata are in a sibling of its parent directory. So we'll go
+            // up two levels before walking the file tree.
+            path = path.getParent().getParent();
```
stephen-carter-at-sf, line 77: This seems odd... why are we getting the parent folder? What is the responsibility of processSourceFolder?Does SFGE really assume that the apex files must be in a project?

Can we not just find the files we ware looking for from the workspace files without imposing requirements on the salesforce project structure? That is, can't we just get all the apex files from the workspace, get all the visualforce files (if needed), etc purely by file extensions?

stephen-carter-at-sf: For now, if you want to put in a massive "TODO" with a big note that says that we are going beyond the bounds of the workspace files provided by the user and making a massive assumption that the users files adhere to a salesforce project structure.

Then maybe see if you timebox a follow up item to see if we can just process the workspace files directly, that would be ideal.

stephen-carter-at-sf (approved):
If you can at least put in those TODOs... I gotta run for the day - so Approving for now so that you can merge and follow up with subsequent PRs.

PR #257, opened 2025-04-09
NEW: @W-18095617@ File Targeting for Flow Engine

randi274:
Respects the option of targeted files by adding a check to `getTargetedFiles` on the workspace. If files are targeted, the engine filters out any irrelevant file execution data in post-processing.

This change ensures that the Flow Engine is aware of all of the workspace files during processing, so that rules can be aware of the relationships between all flows. This also means the engine scans all flows during each flow execution.

In a future version, this work should ideally be done by the python module and we can make an update to pass in both the workspace and the targeted files to it. For now the experience for the user should be the same since the Flow Engine runs quickly enough that latency should not be an issue.

### DIFF
--- a/packages/code-analyzer-flow-engine/src/engine.ts
+++ b/packages/code-analyzer-flow-engine/src/engine.ts
@@ -34,7 +34,7 @@ export class FlowScannerEngine extends Engine {
 
     public constructor(commandWrapper: FlowScannerCommandWrapper, clock: Clock = new RealClock()) {
         super();
-        this.commandWrapper =  commandWrapper;
+        this.commandWrapper = commandWrapper;
         this.clock = clock;
     }
 
@@ -64,8 +64,8 @@ export class FlowScannerEngine extends Engine {
 
     public async runRules(ruleNames: string[], runOptions: RunOptions): Promise<EngineRunResults> {
         this.emitRunRulesProgressEvent(0);
-        const relevantFiles: string[] = await this.getRelevantFiles(runOptions.workspace);
-        if (relevantFiles.length == 0) {
+        const allFlowsInWorkspace: string[] = await this.getRelevantFiles(runOptions.workspace);
+        if (allFlowsInWorkspace.length == 0) {
             return { violations: [] };
         }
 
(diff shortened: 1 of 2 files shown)


packages/code-analyzer-flow-engine/src/engine.ts
```diff
@@ -34,7 +34,7 @@ export class FlowScannerEngine extends Engine {
 
     public constructor(commandWrapper: FlowScannerCommandWrapper, clock: Clock = new RealClock()) {
         super();
-        this.commandWrapper =  commandWrapper;
+        this.commandWrapper = commandWrapper;
```
randi274, line 37: Space change

packages/code-analyzer-flow-engine/src/engine.ts
```diff
@@ -64,8 +64,8 @@ export class FlowScannerEngine extends Engine {
 
     public async runRules(ruleNames: string[], runOptions: RunOptions): Promise<EngineRunResults> {
         this.emitRunRulesProgressEvent(0);
-        const relevantFiles: string[] = await this.getRelevantFiles(runOptions.workspace);
-        if (relevantFiles.length == 0) {
+        const allFlowsInWorkspace: string[] = await this.getRelevantFiles(runOptions.workspace);
```
randi274, line 67: Updated this string to be clearer for what it means today

packages/code-analyzer-flow-engine/src/engine.ts
```diff
@@ -76,22 +76,32 @@ export class FlowScannerEngine extends Engine {
...
+        // execution results for targeted files in post-processing
         const executionResults: FlowScannerExecutionResult = await this.commandWrapper.runFlowScannerRules(
-            relevantFiles, logFile, percentageUpdateHandler);
-        const convertedResults: EngineRunResults = toEngineRunResults(executionResults, ruleNames);
+            allFlowsInWorkspace,
+            logFile,
+            percentageUpdateHandler
+        );
+
+        const targetedFiles = await runOptions.workspace.getTargetedFiles();
```
stephen-carter-at-sf, line 90: Not sure if it will change performance in anyway... but my instincts tell me that this should be 
```
const targetedFiles = await runOptions.workspace.getTargetedFiles();
const relevantTargetedFiles = targetedFiles.filter(fileIsFlowFile);
```
But if it isn't needed then no big deal.

packages/code-analyzer-flow-engine/src/engine.ts
```diff
@@ -76,22 +76,32 @@ export class FlowScannerEngine extends Engine {
...
         this.emitRunRulesProgressEvent(100);
         return convertedResults;
     }
 
+    /**
+     * Gets all of the flow files in the workspace or the cached files if available
+     * @param workspace
+     * @returns a list of file names
+     */
     private async getRelevantFiles(workspace: Workspace): Promise<string[]> {
```
stephen-carter-at-sf, line 101: Maybe to be more clear rename this to be getRelevantWorkspaceFiles

packages/code-analyzer-flow-engine/src/engine.ts
```diff
@@ -160,3 +179,14 @@ function toCodeLocationList(flowNodes: FlowNodeDescriptor[]): CodeLocation[] {
...
+/**
+ * Workaround function to only return targeted flow results if there were targeted files
+ */
+function getNodesForTargetedFlows(allFlowNodes: FlowNodeDescriptor[], requestedFiles: string[]): FlowNodeDescriptor[] {
+    if (requestedFiles.length === 0) {
+        return allFlowNodes;
+    }
+    const requestedFilesSet: Set<string> = new Set(requestedFiles);
+    return allFlowNodes.filter((node) => requestedFilesSet.has(node.flow_path));
+}
```
stephen-carter-at-sf, line 192: This seems wrong.

Basically if flow A depends on a child flow B. and a result has A-->B in it then:
~ If A is targeted and B is targeted, then keep the result
~ If A is targeted but B is not targeted, then keep the result only if B is in the workspace
    Technically... this shouldn't ever happen if the workspace files are being passed in because if B is not in the workspace then he shouldn't be picking up B. If he does, then that's a bug on his end. Instead he should be returning a result with A only and not B. At which point you should let Robert know.
~ If B is targeted but A is not, then do not keep the result
~ If neither are targeted, then do not keep the result.

stephen-carter-at-sf (changes requested):

packages/code-analyzer-flow-engine/src/engine.ts
randi274: This is good to think about! On review, I think we want it the way it's set now, but here's my thought process:

Kept as it is without the filter, if a user only wants to target "MyClass.cls" but the workspace had other flow violations, then no violations would be returned, because we should *only* care about MyClass.cls. The flow engine in this case isn't relevant, but if we're running all engines it could get called, right?

If I check the relevancy of target files and none are flow files, then all violations in the workspace will be returned, because requestedFiles would evaluate to empty in `getNodesForTargetedFlows`

PR #260, opened 2025-04-16
NEW (SFGE) @W-18162162@ Internal execution errors now treated as errors

jfeingold35:


packages/code-analyzer-sfge-engine/src/engine.ts
```diff
@@ -148,7 +151,18 @@ export class SfgeEngine extends Engine {
...
+            const msg = getMessage(
+                'InternalExecutionErrorMessageTemplate',
+                sfgeViolation.sourceFileName,
+                sfgeViolation.sourceLineNumber,
+                sfgeViolation.sourceColumnNumber,
+                sfgeViolation.message
+            );
+            this.emitLogEvent(LogLevel.Error, msg);
+            return undefined;
+        }
```
stephen-carter-at-sf: This in a way violates the single responsibility of this function. So may I recommend instead to do the following up at line 104:
```
sfgeResults.filter(r => r.ruleName === 'InternalExecutionError').map(r => 
    this.emitLogEvent(LogLevel.Error, getMessage(...));
const violations: Violation[] = sfgeResults.filter(r => r.ruleName !== 'InternalExecutionError').map(toViolation);
```
where toViolation is a function instead of a private method.

jfeingold35: Yeah, fair enough. Will do.

packages/code-analyzer-sfge-engine/src/engine.ts
```diff
@@ -148,7 +151,18 @@ export class SfgeEngine extends Engine {
         return this.sfgeRuleInfoListCache.get(cacheKey)!;
     }
 
-    private toViolation(sfgeViolation: SfgeRunResult): Violation {
+    private toViolation(sfgeViolation: SfgeRunResult): Violation|undefined {
+        if (sfgeViolation.ruleName === 'InternalExecutionError') {
```
jfeingold35: NOTE: I'm only checking for `InternalExecutionError`, because `LimitReached` is actually extremely difficult to cause thanks to the memory enhancements I did a while back. So I'm thinking we can just leave it as-is, and if anyone actually reports a `LimitReached`, we can figure out how they did it and add support for it.

stephen-carter-at-sf (approved):

PR #240, opened 2025-03-20
NEW: @W-17915416@: Visualforce rules for PMD

randi274:
Add PMD Visualforce rules from the AppExchange team:
- AvoidCreateElementScriptLinkTag
- AvoidHardCodedSecretsInVFAttrs
- AvoidUnauthorizedApiSessionIdVisualforce
- AvoidUnauthorizedGetSessionIdInVisualforce
- LoadCSSApexStylesheet
- LoadCSSLinkHref
- LoadJavaScriptHtmlScript
- LoadJavaScriptIncludeScript

PR #261, opened 2025-04-16
CHANGE: @W-18064974@ Update PMD and dependencies

randi274:
Notes:
- The range for ExcessiveParameterList has increased, as it now covers the entire method instead of just the signature.
- New PMD rules are also available for Visualforce (see #240)
- `no-with` is no longer a recommended Typescript rule

### DIFF
--- a/packages/ENGINE-TEMPLATE/package.json
+++ b/packages/ENGINE-TEMPLATE/package.json
@@ -22,7 +22,7 @@
     "eslint": "^8.57.1",
     "jest": "^29.7.0",
     "rimraf": "*",
-    "ts-jest": "29.2.3",
+    "ts-jest": "29.3.2",
     "typescript": "^5.7.3",
     "typescript-eslint": "^8.22.0"
   },
--- a/packages/ENGINE-TEMPLATE/package.json
+++ b/packages/ENGINE-TEMPLATE/package.json
@@ -14,7 +14,7 @@
   "types": "dist/index.d.ts",
   "dependencies": {
     "@types/node": "^20.0.0",
-    "@salesforce/code-analyzer-engine-api": "0.21.0"
+    "@salesforce/code-analyzer-engine-api": "0.21.0-SNAPSHOT"
   },
   "devDependencies": {
     "@eslint/js": "^8.57.1",
--- a/packages/ENGINE-TEMPLATE/package.json
+++ b/packages/ENGINE-TEMPLATE/package.json
@@ -23,8 +23,8 @@
     "jest": "^29.7.0",
     "rimraf": "*",
     "ts-jest": "29.3.2",
-    "typescript": "^5.7.3",
-    "typescript-eslint": "^8.22.0"
+    "typescript": "^5.8.3",
+    "typescript-eslint": "^8.30.1"
   },
   "engines": {
     "node": ">=20.0.0"
--- a/packages/code-analyzer-engine-api/test/utils/utils.test.ts
+++ b/packages/code-analyzer-engine-api/test/utils/utils.test.ts
@@ -34,9 +34,9 @@ describe('Tests for createTempDir', () => {
         // Create the directory.
         const tempDir: string = await createTempDir();
 
-        // Verify that the temp folder has one additional entry, and that an entry with the temporary name now exists.
+        // Verify that the temp folder has at least one additional entry, and that an entry with the temporary name now exists.
         const postTestTempContentsCount: number = (await fs.promises.readdir(os.tmpdir())).length;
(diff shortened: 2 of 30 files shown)


packages/ENGINE-TEMPLATE/package.json
```diff
@@ -22,7 +22,7 @@
     "eslint": "^8.57.1",
     "jest": "^29.7.0",
     "rimraf": "*",
-    "ts-jest": "29.2.3",
+    "ts-jest": "29.3.2",
```
stephen-carter-at-sf, line 25: Note in the work item we have the note:

** Note: We may not be able to upgrade ts-jest past 29.2.3 due to:  https://github.com/kulshekhar/ts-jest/issues/4641

Has this been resolved?

stephen-carter-at-sf: I see a note on that issue that says:

> This is fixed in 29.3.0 You will need to enable isolatedModules: true in packages/code-analyzer-eslint-engine/tsconfig.json I hope this helps :)

So let's give that a try. But before committing to it... let's try to understand what this actually does and understand why it fixes the issue: https://www.typescriptlang.org/tsconfig/#isolatedModules

I would hate to have unintended consequences.

And do we need it only with the eslint engine or for all engines?

randi274: So it did turn out to be necessary for all engines, or we were given this warning: 
Screenshot 2025-04-18 at 10 51 37 AM

packages/code-analyzer-engine-api/test/utils/utils.test.ts
```diff
@@ -34,9 +34,9 @@ describe('Tests for createTempDir', () => {
         // Create the directory.
         const tempDir: string = await createTempDir();
 
-        // Verify that the temp folder has one additional entry, and that an entry with the temporary name now exists.
+        // Verify that the temp folder has at least one additional entry, and that an entry with the temporary name now exists.
```
randi274, line 37: I think this may be a side effect of the `isolatedModules` config change, as we were seeing that additional files may now be added into that tempDir at the same time, so I updated the assertion that *at least* one more is added, and kept that the specific `tempDir` can be found.

Screenshot 2025-04-18 at 11 00 49 AM

packages/code-analyzer-eslint-engine/src/rule-mappings.ts
```diff
@@ -686,7 +686,7 @@ export const RULE_MAPPINGS: Record<string, {severity: SeverityLevel, tags: strin
     },
     "no-with": {
         severity: SeverityLevel.Moderate,
-        tags: [COMMON_TAGS.RECOMMENDED, COMMON_TAGS.CATEGORIES.BEST_PRACTICES, COMMON_TAGS.LANGUAGES.JAVASCRIPT, COMMON_TAGS.LANGUAGES.TYPESCRIPT]
+        tags: [COMMON_TAGS.RECOMMENDED, COMMON_TAGS.CATEGORIES.BEST_PRACTICES, COMMON_TAGS.LANGUAGES.JAVASCRIPT]
```
stephen-carter-at-sf: Can you please add in the comment as is on other lines: // Not available with TypeScript

packages/code-analyzer-engine-api/tsconfig.json
```diff
@@ -3,7 +3,8 @@
   "compilerOptions": {
     "composite": true,
     "outDir": "./dist",
-    "rootDir": "."
+    "rootDir": ".",
+    "isolatedModules": true
```
stephen-carter-at-sf, line 7: Why is this needed for this module. I thought only the code-analyzer-eslint-engine package needed to be updated.

packages/code-analyzer-engine-api/test/utils/utils.test.ts
stephen-carter-at-sf: I don't understand the connection. What other file is showing up? And I can't reproduce. Maybe you got unlucky and some other background process happen to create another file at that exact moment you ran the test (which is super unlikely and probably couldn't happen in github runner).

packages/ENGINE-TEMPLATE/package.json
stephen-carter-at-sf: OK yeah the more I looked into this... the more I do believe that yes we do want this on all our packages..

packages/code-analyzer-engine-api/test/utils/utils.test.ts
stephen-carter-at-sf: Anyway - probably just sporadic - but this test change is fine.

stephen-carter-at-sf (approved):

PR #262, opened 2025-04-16
CHANGE @W-18292873@ Modularized release branch creation script

jfeingold35:
This PR does the following:
- Replaces a script in `publish-package-to-npm.yml` that previously needed to be manually updated with every new engine with a JS file that dynamically generates an equivalent script encompassing all packages.

For an example and proof of concept, see #256 , which was able to successfully create [this branch](https://github.com/forcedotcom/code-analyzer-core/tree/experiment/1744833961) and add a commit updating all of the `package.json` files as appropriate.

### DIFF
--- a/.github/workflows/publish-package-to-npm.yml
+++ b/.github/workflows/publish-package-to-npm.yml
@@ -138,101 +138,14 @@ jobs:
           npm run build
           # No need to test; that comes later.
       - name: Commit changes to release branch
+        shell: zsh {0}
         run: |
-          # GraphQL needs to know what branch to push to.
-          BRANCH_NAME=$(git rev-parse --abbrev-ref HEAD)
-          # GraphQL needs a message for the commit.
-          MESSAGE="Preparing Core Ecosystem for release"
-          # GraphQL needs the latest versions of all the package.json files, as Base64 encoded strings.
-          CORE_PACKAGE_JSON="$(cat packages/code-analyzer-core/package.json | base64)"
-          API_PACKAGE_JSON="$(cat packages/code-analyzer-engine-api/package.json | base64)"
-          ESLINT_PACKAGE_JSON="$(cat packages/code-analyzer-eslint-engine/package.json | base64)"
-          FLOW_PACKAGE_JSON="$(cat packages/code-analyzer-flow-engine/package.json | base64)"
-          PMD_PACKAGE_JSON="$(cat packages/code-analyzer-pmd-engine/package.json | base64)"
-          REGEX_PACKAGE_JSON="$(cat packages/code-analyzer-regex-engine/package.json | base64)"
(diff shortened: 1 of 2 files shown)


.github/workflows/publish-package-to-npm.yml
```diff
@@ -138,101 +138,14 @@ jobs:
           npm run build
           # No need to test; that comes later.
       - name: Commit changes to release branch
+        shell: zsh {0}
```
jfeingold35, line 141: Using `zsh` was necessary to get a consistency of behavior between my local CLI and the GHA. Without it, none of this was working.

.github/workflows/publish-package-to-npm/create-branch-commit-script.js
jfeingold35, line 1: JS Script that outputs a dynamically generated version of the previously hardcoded `publish-package-to-npm.yml` code.

.github/workflows/publish-package-to-npm.yml
```diff
@@ -138,101 +138,14 @@ jobs:
...
-              }) {
-                commit {
-                  id
-                }
-              }
-            }'
+          cd packages
+          ALL_PACKAGES=`ls`
+          cd ..
+          node ./.github/workflows/publish-package-to-npm/create-branch-commit-script.js "$ALL_PACKAGES"
```
jfeingold35, line 146: This way, the `package.json` for every folder within `packages` will be automatically added via the script, removing the need for us to manually update it every time a new engine gets added.

.github/workflows/publish-package-to-npm/create-branch-commit-script.js
```diff
@@ -0,0 +1,100 @@
...
+
+function main() {
+    // Split on `\n` instead of ` ` because this argument comes from $(ls).
+    const packageNames = process.argv[2].split("\n");
+
+    const scriptLines = [];
+    scriptLines.push(`BRANCH_NAME=$(git rev-parse --abbrev-ref HEAD)`);
+    scriptLines.push(`MESSAGE="Preparing Core Ecosystem for release"`);
+    scriptLines.push(...generatePackageVarDeclarations(packageNames));
+    scriptLines.push(`PACKAGE_LOCK="$(cat package-lock.json | base64)"`);
```
jfeingold35, line 11: The `package-lock.json` file is hardcoded in here in a few places.

.github/workflows/publish-package-to-npm.yml
stephen-carter-at-sf, line 146: It looks like you have created a node script that then creates a graph-ql script. If we are switching over to node, then why not just perform a `npm install @octokit/graphql` and use that module to directly do the same thing so that you don't need to do an eval back over here?

PR #263, opened 2025-04-17
CHANGE @W-18292873@ Modularized release branch creation script

jfeingold35:
This PR does the following:
- Replaces a script in `publish-package-to-npm.yml` that previously needed to be manually updated with every new engine with a JS file that does the commit via `@octokit/graphql` in a modular fashion.

For an example and proof of concept, see #256  , which was able to successfully create [this branch](https://github.com/forcedotcom/code-analyzer-core/tree/experiment_2/1744911806) and add a commit updating all of the `package.json` files as appropriate.


.github/workflows/publish-package-to-npm/package.json
```diff
@@ -0,0 +1,11 @@
+{
+    "name": "publish-package-to-npm",
+    "version": "1.0.0",
+    "scripts": {},
+    "author": "",
+    "license": "ISC",
+    "description": "",
```
stephen-carter-at-sf: I don't think these are needed. Instead can you add in 
```
"private": true
```
to replace lines 4 to 7 so that it indicates this package is private and not for publishing. This prevents npm publish from working on this command.

.github/workflows/publish-package-to-npm/commit-via-graphql.js
```diff
@@ -0,0 +1,86 @@
+const fs = require('fs/promises');
+const child_process = require('child_process');
+const graphql = require('@octokit/graphql');
+const path = require('path');
```
stephen-carter-at-sf: Can we instead make the file a .mjs file so that we don't use common-js require syntax? imports are prefered.

stephen-carter-at-sf (approved):

packages/code-analyzer-sfge-engine/test/engine.test.ts
```diff
@@ -348,7 +348,7 @@ describe('SfgeEngine', () => {
         it('InternalErrorViolations are thrown as non-fatal errors', async () => {
             // ====== SETUP ======
             const configWithLowTimeout: SfgeEngineConfig = JSON.parse(JSON.stringify(DEFAULT_SFGE_ENGINE_CONFIG)) as SfgeEngineConfig;
-            configWithLowTimeout.java_thread_timeout = 10; // Set the timeout to an extremely low value, to increase the likelihood of getting a timeout.
+            configWithLowTimeout.java_thread_timeout = 3; // Set the timeout to an extremely low value, to increase the likelihood of getting a timeout.
```
stephen-carter-at-sf, line 351: Does line 365 need to be updated to or no?

jfeingold35: Oops, yeah.

PR #265, opened 2025-04-21
NEW @W-17915999@ Implemented engine-level telemetry framework.

jfeingold35:
This PR does the following:
- Adds, at the engine-level, a framework for emitting Telemetry Events.
- Adds, at the core-level, an ability to receive engine-level Telemetry Events and re-emit them for reception by CLI, VSCode, or other wrapper layers.

### DIFF
--- a/packages/code-analyzer-core/src/code-analyzer.ts
+++ b/packages/code-analyzer-core/src/code-analyzer.ts
@@ -16,8 +16,7 @@ import {
     Event,
     EventType,
     LogLevel,
-    TelemetryData,
-    TelemetryEvent
+    TelemetryData
 } from "./events"
 import {getMessage} from "./messages";
 import * as engApi from "@salesforce/code-analyzer-engine-api"
@@ -400,12 +399,14 @@ export class CodeAnalyzer {
         })
     }
 
+    // This method is currently unused, so no coverage is possible. However, it's going to be used very shortly, so we're
+    // adding it now and just disabling the coverage check for it.
     // istanbul ignore next
     private emitTelemetryEvent(eventName: string, data: TelemetryData): void {
         this.emitEvent({
             type: EventType.TelemetryEvent,
             timestamp: this.clock.now(),
-            eventName,
+            eventName,k
             uuid: this.uniqueIdGenerator.getUniversallyUniqueId(),
             data
         });
--- a/packages/code-analyzer-core/src/code-analyzer.ts
+++ b/packages/code-analyzer-core/src/code-analyzer.ts
@@ -406,7 +406,7 @@ export class CodeAnalyzer {
         this.emitEvent({
             type: EventType.TelemetryEvent,
             timestamp: this.clock.now(),
-            eventName,k
+            eventName,
             uuid: this.uniqueIdGenerator.getUniversallyUniqueId(),
             data
         });
--- a/packages/code-analyzer-core/src/code-analyzer.ts
+++ b/packages/code-analyzer-core/src/code-analyzer.ts
@@ -469,7 +469,7 @@ export class CodeAnalyzer {
                 timestamp: this.clock.now(),
                 engineName: engine.getName(),
                 type: EventType.EngineTelemetryEvent,
-                key: event.key,
+                eventName: event.eventName,
                 data: event.data
             });
         });
--- a/packages/code-analyzer-core/src/code-analyzer.ts
+++ b/packages/code-analyzer-core/src/code-analyzer.ts
@@ -22,7 +22,7 @@ import * as engApi from "@salesforce/code-analyzer-engine-api"
 import {Clock, RealClock} from '@salesforce/code-analyzer-engine-api/utils';
 import {EventEmitter} from "node:events";
 import {CodeAnalyzerConfig, ConfigDescription, EngineOverrides, FIELDS, RuleOverride} from "./config";
-import {EngineProgressAggregator, SimpleUniqueIdGenerator, toAbsolutePath, UniqueIdGenerator} from "./utils";
+import {EngineProgressAggregator, RuntimeUniqueIdGenerator, toAbsolutePath, UniqueIdGenerator} from "./utils";
 import fs from "node:fs";
 import path from 'node:path';
 
@@ -92,7 +92,7 @@ const MINIMUM_SUPPORTED_NODE = 20;
 export class CodeAnalyzer {
     private readonly config: CodeAnalyzerConfig;
     private clock: Clock = new RealClock();
-    private uniqueIdGenerator: UniqueIdGenerator = new SimpleUniqueIdGenerator();
+    private uniqueIdGenerator: UniqueIdGenerator = new RuntimeUniqueIdGenerator();
     private readonly eventEmitter: EventEmitter = new EventEmitter();
     private readonly engines: Map<string, engApi.Engine> = new Map();
     private readonly uninstantiableEnginesMap: Map<string, Error> = new Map();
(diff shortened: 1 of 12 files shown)


packages/code-analyzer-engine-api/src/events.ts
```diff
@@ -28,6 +29,17 @@ export type LogEvent = {
...
 }
 
+/**
+ * Event emitted when an engine wants to send telemetry data back up through Core.
+ * These events are received by callbacks provided to the {@link Engine.onEvent} method for {@link EventType.TelemetryEvent}.
+ */
+export type TelemetryEvent = {
+    type: EventType.TelemetryEvent,
+    key: string,
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
jfeingold35: For some reason, this rule isn't actually being run against our code (which it feels like it should be), but I added the `ignore` anyway just to be thorough.

packages/code-analyzer-engine-api/src/events.ts
```diff
@@ -28,6 +29,17 @@ export type LogEvent = {
     message: string
 }
 
+/**
+ * Event emitted when an engine wants to send telemetry data back up through Core.
+ * These events are received by callbacks provided to the {@link Engine.onEvent} method for {@link EventType.TelemetryEvent}.
+ */
+export type TelemetryEvent = {
+    type: EventType.TelemetryEvent,
+    key: string,
```
jfeingold35: A string-type `key` and a completely loose `data` object seemed to me to strike the balance between structure and flexibility that we want for telemetry.

packages/code-analyzer-core/src/events.ts
```diff
@@ -55,6 +56,19 @@ export type EngineLogEvent = {
...
+ * Event emitted when an engine wants to send a telemetry event.
+ * These events are received by callbacks provided to the {@link CodeAnalyzer.onEvent} method for {@link EventType.EngineTelemetryEvent}.
+ */
+export type EngineTelemetryEvent = {
+    type: EventType.EngineTelemetryEvent,
+    timestamp: Date,
+    engineName: string,
+    key: string,
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    data: {[key: string]: any}
```
jfeingold35: Preserving the structure of the engine-level event.

packages/code-analyzer-core/src/code-analyzer.ts
```diff
@@ -8,18 +8,21 @@ import {
...
 import {Clock, RealClock} from '@salesforce/code-analyzer-engine-api/utils';
 import {EventEmitter} from "node:events";
 import {CodeAnalyzerConfig, ConfigDescription, EngineOverrides, FIELDS, RuleOverride} from "./config";
-import {
-    EngineProgressAggregator,
-    SimpleUniqueIdGenerator,
-    toAbsolutePath,
-    UniqueIdGenerator
-} from "./utils";
+import {EngineProgressAggregator, SimpleUniqueIdGenerator, toAbsolutePath, UniqueIdGenerator} from "./utils";
```
jfeingold35: Not sure why my IDE saw fit to change this. I can change it back if we care.

packages/code-analyzer-engine-api/src/events.ts
jfeingold35: Okay, it looks like this rule is being run only on Windows for some reason? Whatever, it doesn't matter, we've got the directive.

packages/code-analyzer-engine-api/src/engines.ts
```diff
@@ -114,6 +114,20 @@ export abstract class Engine {
         });
     }
 
+    /**
+     * Convenience method that subclasses can use to easily emit an event of type {@link TelemetryEvent}
+     * @param eventName A (preferably unique) key to associate with this event
+     * @param data A flexible payload of relevant data
+     * @protected
+     */
+    protected emitTelemetryEvent(eventName: string, data: {[key: string]: null|boolean|number|string}): void {
```
stephen-carter-at-sf: Do we want/need null? What would that be for?

Also, consider using Record instead:
```
data: Record<string, boolean|number|string>
```

packages/code-analyzer-core/src/events.ts
```diff
@@ -55,6 +56,18 @@ export type EngineLogEvent = {
...
+/**
+ * Event emitted when an engine wants to send a telemetry event.
+ * These events are received by callbacks provided to the {@link CodeAnalyzer.onEvent} method for {@link EventType.EngineTelemetryEvent}.
+ */
+export type EngineTelemetryEvent = {
+    type: EventType.EngineTelemetryEvent,
+    timestamp: Date,
+    engineName: string,
+    eventName: string,
+    data: {[key: string]: null|boolean|string|number}
```
stephen-carter-at-sf: ditto

packages/code-analyzer-engine-api/src/events.ts
```diff
@@ -28,6 +29,16 @@ export type LogEvent = {
...
 }
 
+/**
+ * Event emitted when an engine wants to send telemetry data back up through Core.
+ * These events are received by callbacks provided to the {@link Engine.onEvent} method for {@link EventType.TelemetryEvent}.
+ */
+export type TelemetryEvent = {
+    type: EventType.TelemetryEvent,
+    eventName: string,
+    data: {[key: string]: null|string|boolean|number}
```
stephen-carter-at-sf: ditto

packages/code-analyzer-engine-api/src/engines.ts
jfeingold35: The underlying `AnyJson` type lists `null` as one of its accepted primitives, and it seemed reasonable to cover the case where there's a field we want to be intentionally empty.
But yeah, I can switch to `Record`. Stand by for that.

stephen-carter-at-sf: Until we have a use case for null - I think we should avoid having that in my opinion. We can add it in later if we find a use case for it. Because if a user doesn't want the field, the don't need to add it to their telemetry event at all.

packages/code-analyzer-core/src/code-analyzer.ts
```diff
@@ -395,6 +400,17 @@ export class CodeAnalyzer {
...
+    // istanbul ignore next
+    private emitTelemetryEvent(eventName: string, data: TelemetryData): void {
+        this.emitEvent({
+            type: EventType.TelemetryEvent,
+            timestamp: this.clock.now(),
+            eventName,
+            uuid: this.uniqueIdGenerator.getUniversallyUniqueId(),
+            data
+        });
+    }
```
stephen-carter-at-sf, line 413: Probably didn't need to add this until we actually had a use for it. But if you want to update the CLI to prepare for the future scenario of possibly updating core with a core TelemetryEvent, that's fine to do now I suppose. Up to you.

packages/code-analyzer-core/src/code-analyzer.ts
```diff
@@ -395,6 +400,17 @@ export class CodeAnalyzer {
         })
     }
 
+    // istanbul ignore next
```
stephen-carter-at-sf, line 404: If we do add this in even though it isn't being used (which probably means we shouldn't have this added to our code yet), can we at least add in a comment saying that this is unused currently, but when used to remove that istanbul ignore next.

packages/code-analyzer-core/src/utils.ts
```diff
@@ -1,4 +1,5 @@
 import path from "node:path";
+import crypto from "node:crypto";
```
stephen-carter-at-sf: Should both of these instead be `import * as path` and `import * as crypto` instead of directly importing a namespace?

stephen-carter-at-sf (approved):
You have a lint error (unused import to fix). And I left a few comments. Otherwise, approved.

packages/code-analyzer-core/src/utils.ts
jfeingold35: I'll do that for safety's sake, though I'm not positive it actually matters here.

PR #269, opened 2025-04-23
NEW (SFGE) @W-17915999@ Added SFGE telemetry support

jfeingold35:
This PR allows the SFGE engine to transmit telemetry events from the underlying Java module back up through the Typescript layer up to Core.

### DIFF
--- a/packages/code-analyzer-sfge-engine/src/sfge-wrapper.ts
+++ b/packages/code-analyzer-sfge-engine/src/sfge-wrapper.ts
@@ -3,7 +3,8 @@ import fs from 'node:fs';
 import {
     getMessageFromCatalog,
     LogLevel,
-    SHARED_MESSAGE_CATALOG
+    SHARED_MESSAGE_CATALOG,
+    TelemetryData
 } from '@salesforce/code-analyzer-engine-api';
 import {createTempDir, JavaCommandExecutor} from '@salesforce/code-analyzer-engine-api/utils';
 import {getMessage} from "./messages";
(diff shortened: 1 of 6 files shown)


packages/code-analyzer-sfge-engine/src/sfge-wrapper.ts
```diff
@@ -188,7 +201,8 @@ function handleRunStdOut(stdOutMsg: string, emitLog: (logLevel: LogLevel, msg: s
         for (const sfgeMessage of sfgeMessages) {
             if (isSfgeLogMessage(sfgeMessage)) {
                 if (sfgeMessage.messageSeverity === 'TELEMETRY') {
-                    // TODO: TELEMETRY
+                    const telemetryData: TelemetryData = JSON.parse(sfgeMessage.args[0]) as TelemetryData;
+                    emitTelemetry(telemetryData.eventName as string, telemetryData);
```
jfeingold35: Technically, this means that the Telemetry Data will have its own redundant `eventName` parameter, but I don't believe that's worth getting into a tizzy over.

packages/code-analyzer-sfge-engine/test/engine.test.ts
```diff
@@ -347,12 +348,12 @@ describe('SfgeEngine', () => {
 
         it('InternalErrorViolations are thrown as non-fatal errors', async () => {
             // ====== SETUP ======
-            const configWithLowTimeout: SfgeEngineConfig = JSON.parse(JSON.stringify(DEFAULT_SFGE_ENGINE_CONFIG)) as SfgeEngineConfig;
-            configWithLowTimeout.java_thread_timeout = 3; // Set the timeout to an extremely low value, to increase the likelihood of getting a timeout.
-            const engine: SfgeEngine = new SfgeEngine(configWithLowTimeout, fixedClock);
-            const workspace: Workspace = new Workspace('id', [path.join(TEST_DATA_FOLDER, 'sampleRelevantWorkspace')]);
+            const engine: SfgeEngine = new SfgeEngine(DEFAULT_SFGE_ENGINE_CONFIG, fixedClock);
+            const workspace: Workspace = new Workspace('id', [path.join(TEST_DATA_FOLDER, 'sampleUnhandledWorkspace')]);
```
jfeingold35, line 352: Switched this test to work by scanning code that SFGE can't currently handle, instead of setting an absurdly low timeout. For whatever reason, the timeout failure wasn't reliably producing telemetry from within SFGE. That's something we may want to look into, but it's not within the scope of this story.

packages/code-analyzer-sfge-engine/test/test-data/sampleUnhandledWorkspace/MyStringHelper.cls
```diff
@@ -0,0 +1,13 @@
+global virtual class MyStringHelper {
```
jfeingold35, line 1: This code was pulled from [Scanner issue #1177](https://github.com/forcedotcom/sfdx-scanner/issues/1177).

packages/code-analyzer-sfge-engine/src/sfge-wrapper.ts
stephen-carter-at-sf: Whenever possible, we should put JSON.parse in a try/catch in case we don't get the text in full or whatever. Probably could just log with trace level logging if it does fail and just continue as usual. But if you don't like the redundant stuff, then just recreate the data structure over here with what you want. Up to you.

stephen-carter-at-sf (approved):

PR #270, opened 2025-04-23
NEW: @W-18095617@ Support targeting of files in Flow engine

randi274:
Notes:
- Includes latest updates to the python module FlowTest
- Flow engine now supports targeted files to run a subset of a workspace


packages/code-analyzer-flow-engine/test/engine.test.ts
```diff
@@ -468,6 +509,157 @@ describe('Tests for the TestEngine', () => {
...
+
+                        const engineResults: EngineRunResults = await engine.runRules(selectedRuleNames, createRunOptions(
+                            tempFolder, new Workspace('id', [flowA, flowB, flowC], [flowC])));
+
+                        expect(engineResults.violations).toHaveLength(0);
+                    });
+                });
+
+                describe('FlowA references subflow FlowB, they are NOT in the workspace, and both have violations', () => {
+                    // Sanity checks that FlowTest is respecting the workspace boundary
```
randi274: I'm not seeing what I'd expect here - would love a second set of eyes to confirm if either these assumptions are incorrect or FlowTest is not respecting the workspace boundary.

packages/code-analyzer-flow-engine/src/engine.ts
```diff
@@ -78,23 +81,41 @@ export class FlowScannerEngine extends Engine {
...
-        return this.relevantFilesCache.get(cacheKey)!;
+        return this.targetedFlowsCache.get(cacheKey)!;
+    }
+
+    private async hasWorkspaceAndNoTargetedFlows(workspace: Workspace | undefined): Promise<boolean> {
+        if (!workspace) {
+            return false;
+        } else {
+            return (await this.getTargetedFlows(workspace)).length == 0
+        }
```
stephen-carter-at-sf: Lower golf score:
```
return workspace && (await this.getTargetedFlows(workspace)).length === 0;
```

packages/code-analyzer-flow-engine/src/engine.ts
```diff
@@ -64,8 +66,9 @@ export class FlowScannerEngine extends Engine {
 
     public async runRules(ruleNames: string[], runOptions: RunOptions): Promise<EngineRunResults> {
         this.emitRunRulesProgressEvent(0);
-        const relevantFiles: string[] = await this.getRelevantFiles(runOptions.workspace);
-        if (relevantFiles.length == 0) {
+        const targetedFlows: string[] = await this.getTargetedFlows(runOptions.workspace);
+        const workspaceFlows: string[] = await this.getWorkspaceFlows(runOptions.workspace);
+        if (workspaceFlows.length == 0) {
```
stephen-carter-at-sf: I believe this check should be for the targeted flows... and I think you can add it before line 70.

That is I believe this should be:
```
const targetedFlows: string[] = await this.getTargetedFlows(runOptions.workspace);
if (workspaceFlows.length == 0) {
    return { violations: [] };
}
const workspaceFlows: string[] = await this.getWorkspaceFlows(runOptions.workspace);
```

packages/code-analyzer-flow-engine/src/python/FlowScannerCommandWrapper.ts
```diff
@@ -5,7 +5,12 @@ import path from "node:path";
 import fs from "node:fs";
 
 export interface FlowScannerCommandWrapper {
-    runFlowScannerRules(flowFilesToScan: string[], absLogFilePath: string, completionPercentageHandler: (percentage: number) => void): Promise<FlowScannerExecutionResult>;
+    runFlowScannerRules(
+        targetedFilesToScan: string[],
+        workspaceFilesToScan: string[],
```
stephen-carter-at-sf: Can we call this what it is... targetedFlowFiles and workspaceFlowFiles here and in the implementation.

Also i think you have these 2 arguments flipped. Everywhere else you have the 1st input argument as the workspace files and the 2nd as the targeted files.

packages/code-analyzer-flow-engine/src/python/FlowScannerCommandWrapper.ts
```diff
@@ -41,11 +46,17 @@ export class RunTimeFlowScannerCommandWrapper implements FlowScannerCommandWrapp
...
+        targetedFilesToScan: string[],
+        absLogFilePath: string,
+        completionPercentageHandler: (percentage: number) => void
+    ): Promise<FlowScannerExecutionResult> {
         const tempDir: string = await createTempDir();
-        const flowFilesToScanFile: string = path.join(tempDir, 'flowFilesToScan.txt');
-        await fs.promises.writeFile(flowFilesToScanFile, flowFilesToScan.join('\n'), 'utf-8');
+        const workspaceFilesToScanFile: string = path.join(tempDir, 'workspaceFilesToScan.txt');
+        const targetedFilesToScanFile: string = path.join(tempDir, 'targetedFilesToScan.txt');
```
stephen-carter-at-sf: ditto - can we just call these workspaceFlowFiles and targetedFlowFiles

packages/code-analyzer-flow-engine/src/engine.ts
randi274: I tried this swap and had a test failure, so I think what I need is a check on both:

```
const targetedFlows: string[] = await this.getTargetedFlows(runOptions.workspace);
const workspaceFlows: string[] = await this.getWorkspaceFlows(runOptions.workspace);
        
if (workspaceFlows.length == 0 || targetedFlows.length == 0) {
     return { violations: [] };
}
```

stephen-carter-at-sf: But that doesn't make sense. You shouldn't need the or statement. The targeted files must be a subset of the workspace files. Core guarantees that. Did you write a test where this isn't the case? 

When constructing a workspace - you need to keep within the assumptions: https://github.com/forcedotcom/code-analyzer-core/blob/dev/packages/code-analyzer-engine-api/src/workspace.ts#L37

packages/code-analyzer-flow-engine/src/python/FlowScannerCommandWrapper.ts
```diff
@@ -56,8 +67,10 @@ export class RunTimeFlowScannerCommandWrapper implements FlowScannerCommandWrapp
             '--debug',
             '--log_file',
             absLogFilePath,
-            '--infile',
-            flowFilesToScanFile,
+            '--workspace',
+            workspaceFlowsToScanFile,
```
stephen-carter-at-sf: Maybe i'm just being nitpicky here... but my last request to rename this variable was really an attempt to remove the "to scan" piece from it. The workspace flow files are not the ones scanned necessarily ... but the ones available when resolving subflows. So to remove the confusion I was hoping you would remove the "ToScan" piece here from these variable names.

randi274: I like it! Let me remove that as well - I prefer it without "ToScan" as well

packages/code-analyzer-flow-engine/src/engine.ts
```diff
@@ -64,8 +66,10 @@ export class FlowScannerEngine extends Engine {
...
     public async runRules(ruleNames: string[], runOptions: RunOptions): Promise<EngineRunResults> {
         this.emitRunRulesProgressEvent(0);
-        const relevantFiles: string[] = await this.getRelevantFiles(runOptions.workspace);
-        if (relevantFiles.length == 0) {
+        const targetedFlows: string[] = await this.getTargetedFlows(runOptions.workspace);
+        const workspaceFlows: string[] = await this.getWorkspaceFlows(runOptions.workspace);
+        
+        if (workspaceFlows.length == 0 || targetedFlows.length == 0) {
             return { violations: [] };
         }
```
stephen-carter-at-sf: We still have to fix this. We shouldn't be needing to check for workspaceFlows.length === 0 here.

stephen-carter-at-sf (changes requested):

packages/code-analyzer-flow-engine/src/engine.ts
```diff
@@ -50,7 +51,8 @@ export class FlowScannerEngine extends Engine {
 
     public async describeRules(describeOptions: DescribeOptions): Promise<RuleDescription[]> {
         this.emitDescribeRulesProgressEvent(0);
-        if (describeOptions.workspace && (await this.getRelevantFiles(describeOptions.workspace)).length == 0) {
+        const hasWorkspaceAndNoTargetedFlows = await this.hasWorkspaceAndNoTargetedFlows(describeOptions?.workspace);
+        if (hasWorkspaceAndNoTargetedFlows) {
             this.emitLogEvent(LogLevel.Fine, 'No Flow files have been targeted in the workspace. Returning no flow rules.');
```
stephen-carter-at-sf: While you are in here, can you please change this LogLevel.Fine to LogLevel.Debug.

Debug are messages for the user to see what is going on which only appear in their log file but not to the terminal.
Fine are verbose messages really for us to help diagnose issues. This particular message is indeed a user targeted message - not an internal targeted message - so it should be Debug.

packages/code-analyzer-flow-engine/test/engine.test.ts
```diff
@@ -24,15 +24,23 @@ import fs from "node:fs";
...
 const PATH_TO_EXAMPLE3: string = path.join(PATH_TO_MULTIPLE_FLOWS_WORKSPACE, 'example3_containsNoViolations.flow');
 const PATH_TO_EXAMPLE4_PARENTFLOW: string = path.join(PATH_TO_MULTIPLE_FLOWS_WORKSPACE, 'example4_parentFlow.flow-meta.xml');
 const PATH_TO_EXAMPLE4_SUBFLOW: string = path.join(PATH_TO_MULTIPLE_FLOWS_WORKSPACE, 'example4_subflow.flow-meta.xml');
 
+const ALL_FLOW_RULES: string[] = [
+    'PreventPassingUserDataIntoElementWithSharing',
+    'PreventPassingUserDataIntoElementWithoutSharing'
+];
+
 describe('Tests for the TestEngine', () => {
```
stephen-carter-at-sf: Can you fix line 44. It should say "Tests for the FlowScannerEngine" and not "Tests for the TestEngine".

packages/code-analyzer-flow-engine/test/engine.test.ts
```diff
@@ -407,6 +415,17 @@ describe('Tests for the TestEngine', () => {
                 expect(engineResults.violations).toHaveLength(0);
             });
 
+            it('When workspace does not contain flow files but flow files are targeted, then return zero violations', async () => {
+                const selectedRuleNames: string[] = [
+                    'PreventPassingUserDataIntoElementWithSharing',
+                    'PreventPassingUserDataIntoElementWithoutSharing'
+                ];
+                const engineResults: EngineRunResults = await engine.runRules(selectedRuleNames, createRunOptions(
+                    tempFolder, new Workspace('id', [PATH_TO_NO_FLOWS_WORKSPACE], [PATH_TO_EXAMPLE1])));
```
stephen-carter-at-sf: This test is set up incorrectly. You should not be creating a workspace with a list of target files that do not exist in the list of workspace files.

stephen-carter-at-sf (changes requested):

packages/code-analyzer-flow-engine/test/test-data/example workspaces/parent-with-sink-calls-sub-with-source/child_with_source.flow-meta.xml
```diff
@@ -0,0 +1,77 @@
...
+    <formulas>
+        <description>combines id from created case together with user input</description>
+        <name>combine_vars</name>
+        <dataType>String</dataType>
+        <expression>{!create_case} + {!enter_text_subflow}</expression>
+    </formulas>
+    <interviewLabel>child_with_source {!$Flow.CurrentDateTime}</interviewLabel>
+    <label>child_with_source</label>
+    <processType>Flow</processType>
+    <recordCreates>
```
stephen-carter-at-sf: Why does this exist in here? If this is a child with a source - then it shouldn't have a sink.

packages/code-analyzer-flow-engine/test/test-data/example workspaces/parent-with-sink-calls-sub-with-source/parent_with_sink.flow-meta.xml
```diff
@@ -0,0 +1,115 @@
...
+        </connector>
+        <fields>
+            <name>display_result</name>
+            <fieldText>&lt;p&gt;{!call_subflow.output_var1}&lt;/p&gt;</fieldText>
+            <fieldType>DisplayText</fieldType>
+        </fields>
+        <showFooter>true</showFooter>
+        <showHeader>true</showHeader>
+    </screens>
+    <screens>
```
stephen-carter-at-sf: Why is this in here? Isn't this a source... but we want a parent that just has a sink.

stephen-carter-at-sf (approved):

packages/code-analyzer-flow-engine/test/engine.test.ts
```diff
@@ -324,7 +332,7 @@ describe('Tests for the TestEngine', () => {
                         comment: "combine_vars.combine_vars influences assign_enter_to_output.output_var1: Variable Assignment"
                     },
                     {
-                        file: PATH_TO_EXAMPLE4_PARENTFLOW,
+                        file: PATH_TO_EXAMPLE4_SUBFLOW,
```
stephen-carter-at-sf: This seems wrong. It should be the parent. Robert will need to fix again.

packages/code-analyzer-flow-engine/test/engine.test.ts
```diff
@@ -459,15 +459,202 @@ describe('Tests for the TestEngine', () => {
...
+                            comment: "enter_text_subflow.enter_text_subflow: Initialization"
+                        },
+                        {
+                            file: childFlowFile,
+                            startLine: 10,
+                            startColumn: 1,
+                            comment: "enter_text_subflow.enter_text_subflow influences assign_enter_to_output.output_var1: Variable Assignment"
+                        },
+                        {
+                            file: childFlowFile,
```
stephen-carter-at-sf: This seems wrong. It should be the parent. Robert will need to fix again.

stephen-carter-at-sf (changes requested):

stephen-carter-at-sf (approved):

PR #273, opened 2025-04-24
NEW: @W-17701106@: Allow users the ability to control log level for w…

stephen-carter-at-sf:
…hat appears in log file

### DIFF
--- a/packages/code-analyzer-core/src/config.ts
+++ b/packages/code-analyzer-core/src/config.ts
@@ -6,11 +6,13 @@ import * as yaml from 'js-yaml';
 import {getMessage} from "./messages";
 import {deepEquals, toAbsolutePath} from "./utils"
 import {SeverityLevel} from "./rules";
+import {LogLevel} from "./events";
 
 // Only exported internally to share across files
 export const FIELDS = {
     CONFIG_ROOT: 'config_root',
     LOG_FOLDER: 'log_folder',
+    LOG_LEVEL: 'log_level',
     CUSTOM_ENGINE_PLUGIN_MODULES: 'custom_engine_plugin_modules', // Hidden
     RULES: 'rules',
     ENGINES: 'engines',
@@ -36,6 +38,7 @@ export type RuleOverride = {
 type TopLevelConfig = {
     config_root: string
     log_folder: string
+    log_level: LogLevel
     rules: Record<string, RuleOverrides>
     engines: Record<string, EngineOverrides>
     custom_engine_plugin_modules: string[] // INTERNAL USE ONLY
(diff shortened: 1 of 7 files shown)


packages/code-analyzer-core/src/index.ts
```diff
@@ -1,48 +1,65 @@
...
+    CodeAnalyzerConfig
+} from "./config"
+
+export type {
     ConfigDescription,
     ConfigFieldDescription,
     EngineOverrides,
     RuleOverrides,
     RuleOverride
 } from "./config"
```
stephen-carter-at-sf, line 11: From our recent change, where we changed our tsconfig file for jest:

This fixes some of the compiler warnings. We'll want to follow up and do the same for the other packages soon too:
image

packages/code-analyzer-core/src/index.ts
```diff
@@ -1,48 +1,65 @@
 export {
-    CodeAnalyzerConfig,
+    CodeAnalyzerConfig
+} from "./config"
+
+export type {
```
jfeingold35, line 5: Is there a reason that this is formatted the way it is?

packages/code-analyzer-core/src/config.ts
```diff
@@ -304,4 +331,30 @@ function validateAbsolutePath(value: unknown, fieldPath: string): string {
...
+ * @param fieldPath the field path of the value as you want it to appear in validation error messages
+ */
+function validateLogLevel(rawValue: unknown, fieldPath: string): LogLevel {
+    // TODO: We might be able to generalize this method in the engine-api to be a validateNumericEnum
+    //  method instead of having this code which is super close to the validateSeverityLevel implementation.
+
+    let value: unknown = typeof rawValue === 'string' && rawValue.length > 1 ?
+        rawValue.charAt(0).toUpperCase() + rawValue.slice(1).toLowerCase() : rawValue;
+
+    // Note that Object.values(LogLevel) returns [1,2,3,4,5,"Error","Warn","Info","Debug","Fine"]
```
jfeingold35, line 348: Question: is there value in supporting an `Off` option (equal to either 0 or 6), at least for this setting?

stephen-carter-at-sf: I'd prefer to wait for a user to ask for Off before we introduce it. I don't see a use case for it yet do you?

packages/code-analyzer-core/src/index.ts
stephen-carter-at-sf: Not sure what you mean. IntelliJ automatically fixed this file to address the isolatedModules requiring export type on types issue. Formatting looks good to me.

jfeingold35: Okay, I didn't realize it was related to the `isolatedModules` problem.

jfeingold35 (approved):

PR #276, opened 2025-04-28
NEW: @W-17540434@: Implement stylelint engine's describeRules method without workspace awareness

kenamverma:
Implement stylelint engine's describeRules method without workspace awareness - base rules

### DIFF
--- a/packages/code-analyzer-engine-api/src/rules.ts
+++ b/packages/code-analyzer-engine-api/src/rules.ts
@@ -69,11 +69,17 @@ export const COMMON_TAGS = {
         /** Rules that analyze files that have APEX code */
         APEX: "Apex",
 
+        /** Rules that analyze files that have CSS code */
+        CSS: "CSS",
+
         /** Rules that analyze files that have HTML code */
         HTML: "HTML",
 
         /** Rules that analyze files that have JavaScript code */
         JAVASCRIPT: "JavaScript",
+
+        /** Rules that analyze files that have SCSS code */
+        SCSS: "SCSS",
 
         /** Rules that analyze files that have TypeScript code */
         TYPESCRIPT: "TypeScript",
--- a/packages/code-analyzer-stylelint-engine/package.json
+++ b/packages/code-analyzer-stylelint-engine/package.json
@@ -14,7 +14,9 @@
   "types": "dist/index.d.ts",
   "dependencies": {
     "@types/node": "^20.0.0",
-    "@salesforce/code-analyzer-engine-api": "0.23.0"
+    "@salesforce/code-analyzer-engine-api": "0.23.0",
+    "@types/stylelint": "^13.13.3",
+    "stylelint": "^16.18.0"
   },
   "devDependencies": {
     "@eslint/js": "^8.57.1",
--- a/packages/code-analyzer-stylelint-engine/src/engine.ts
+++ b/packages/code-analyzer-stylelint-engine/src/engine.ts
@@ -1,7 +1,20 @@
-import { DescribeOptions, Engine, EngineRunResults, LogLevel, RuleDescription, RunOptions, Violation } from "@salesforce/code-analyzer-engine-api";
-import * as fsp from 'node:fs/promises';
+import {
+    COMMON_TAGS,
+    DescribeOptions,
+    Engine,
+    EngineRunResults,
+    LogLevel,
+    RuleDescription,
+    RunOptions,
+    SeverityLevel,
+    Violation,
+} from "@salesforce/code-analyzer-engine-api";
+import * as fsp from "node:fs/promises";
 import path from "path";
 import { getMessage } from "./messages";
+import stylelint, { RuleMeta } from "stylelint";
+import { RULE_MAPPINGS } from "./rule-mappings";
+import { StylelintRuleStatus } from "./enums";
 
 export class StylelintEngine extends Engine {
     static readonly NAME = "stylelint";
@@ -16,44 +29,58 @@ export class StylelintEngine extends Engine {
     }
 
     public async getEngineVersion(): Promise<string> {
-        const pathToPackageJson: string = path.join(__dirname, '..', 'package.json');
-        const packageJson: {version: string} = JSON.parse(await fsp.readFile(pathToPackageJson, 'utf-8'));
+        const pathToPackageJson: string = path.join(
+            __dirname,
+            "..",
+            "package.json"
+        );
+        const packageJson: { version: string } = JSON.parse(
+            await fsp.readFile(pathToPackageJson, "utf-8")
+        );
         return packageJson.version;
     }
 
     // *** Remove underscore for private naming convention if you need any of the DescribeOptions
-    async describeRules(_describeOptions: DescribeOptions): Promise<RuleDescription[]> {
+    async describeRules(
+        _describeOptions: DescribeOptions
+    ): Promise<RuleDescription[]> {
         // *** Best Practice - Use RunRulesProgressEvents to keep users informed on the scan progress
         this.emitRunRulesProgressEvent(0);
 
         // *** Parse out relevant file types if you engine is language-specific
         //const relevantFiles: string[] | undefined;
 
-        // *** Get your rules!
         const ruleDescriptions: RuleDescription[] = [];
+        // Get all available rules from stylelint
+        const allRules = stylelint.rules;
 
         this.emitRunRulesProgressEvent(50);
 
-        // *** Retrieval Implementation is up to you, but you'll need to map them into RuleDescription form, such as:
-        const exampleRule: RuleDescription = {
-            name: "ExampleStylelintRule",
-            severityLevel: 5,
-            tags: [
-              "Recommended"
-            ],
-            description: "The description of your rule",
-            resourceUrls: []
-        };
-        ruleDescriptions.push(exampleRule);
-        
+        for (const [ruleName] of Object.entries(allRules)) {
+            const rule = await allRules[ruleName as keyof typeof allRules];
+            const ruleMetadata: RuleMeta | undefined = rule.meta;
+            if (ruleMetadata) {
+                // do not include rules that don't have metadata
+
+                ruleDescriptions.push(
+                    toRuleDescription(ruleName, ruleMetadata)
+                ); //no rules are turned on by default.
+            }
+        }
+
         // *** Best Practice - Set RunRulesProgressEvents to 100 before completing
         this.emitDescribeRulesProgressEvent(100);
-        return ruleDescriptions;
+
+        return ruleDescriptions.sort((d1, d2) =>
+            d1.name.localeCompare(d2.name)
+        );
     }
 
     // *** ruleNames comes from a describeRules call made just before running
-    async runRules(_ruleNames: string[], _runOptions: RunOptions): Promise<EngineRunResults> {
-
+    async runRules(
+        _ruleNames: string[],
+        _runOptions: RunOptions
+    ): Promise<EngineRunResults> {
         // *** Best Practice - Use RunRulesProgressEvents to keep users informed on the scan progress
         this.emitRunRulesProgressEvent(2);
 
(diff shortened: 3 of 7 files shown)


packages/code-analyzer-stylelint-engine/src/engine.ts
```diff
@@ -65,12 +92,77 @@ export class StylelintEngine extends Engine {
...
+        return SeverityLevel.Info;
+    } else if (status === StylelintRuleStatus.ERROR) {
+        // The "error" status is typically something users care most about, so we mark these as High severity.
+        return SeverityLevel.High;
+    } else if (status === StylelintRuleStatus.NULL) {
+        // The "null" status are for marking the rules 'off', so we mark these as Low severity.
+        return SeverityLevel.Low;
+    }
+    // All else will give assigned High. Recall that users may override these severities if they wish. All rules are default to error severity: https://stylelint.io/user-guide/configure#severity
+    return SeverityLevel.High;
```
kenamverma, line 153: The severity of all rules by default is error, which is supposed to be high?

kenamverma: Ref: https://stylelint.io/user-guide/configure#severity

packages/code-analyzer-stylelint-engine/src/engine.ts
```diff
@@ -16,44 +29,58 @@ export class StylelintEngine extends Engine {
...
-        
+        for (const [ruleName] of Object.entries(allRules)) {
+            const rule = await allRules[ruleName as keyof typeof allRules];
+            const ruleMetadata: RuleMeta | undefined = rule.meta;
+            if (ruleMetadata) {
+                // do not include rules that don't have metadata
+
+                ruleDescriptions.push(
+                    toRuleDescription(ruleName, ruleMetadata)
+                ); //no rules are turned on by default.
```
kenamverma, line 67: No rules are turned on by default. I am not sure where this info will be used but it is better to be informed.
Ref: https://stylelint.io/user-guide/configure#rules

packages/code-analyzer-stylelint-engine/test/engine.test.ts
```diff
@@ -1,71 +1,307 @@
...
+        it("Should return valid violations when rules are violated", async () => {
+            const engine: StylelintEngine = new StylelintEngine();
+            const results: EngineRunResults = await engine.runRules(
+                ["color-no-invalid-hex"],
+                createRunOptions(new Workspace("id", [TEST_DATA_FOLDER]))
+            );
+
+            // Since we don't have test files with violations, we can at least verify the structure
+            expect(Array.isArray(results.violations)).toBe(true);
+        });
```
kenamverma, line 288: Tests might change acc to new requirements but for now:
image

packages/code-analyzer-stylelint-engine/src/engine.ts
```diff
@@ -65,12 +92,77 @@ export class StylelintEngine extends Engine {
...
+            violations: violations,
         };
     }
-}
\ No newline at end of file
+}
+
+function toRuleDescription(
+    ruleName: string,
+    metadata: RuleMeta
```
kenamverma, line 110: RuleMeta only consists of url, fixable and deprecated. We lack info on description and status.
	export type RuleMeta = {
		url: string;
		deprecated?: boolean;
		fixable?: boolean;
	};

packages/code-analyzer-engine-api/src/rules.ts
```diff
@@ -69,12 +69,18 @@ export const COMMON_TAGS = {
...
+        CSS: "CSS",
+
         /** Rules that analyze files that have HTML code */
         HTML: "HTML",
 
         /** Rules that analyze files that have JavaScript code */
         JAVASCRIPT: "JavaScript",
 
+        /** Rules that analyze files that have SCSS code */
+        SCSS: "SCSS",
```
randi274, line 82: Since Salesforce doesn't directly support SCSS, is this something we should be providing @stephen-carter-at-sf? I know code analyzer can support non-Salesforce products, but just wondering if this fits into our purview.

stephen-carter-at-sf: [kenamverma](https://github.com/kenamverma) do any of your rules target SCSS?

If not, then yeah we should not be adding SCSS as a language tag here. Even if we did want to support SCSS files... we could still keep that under the CSS language tag anyway (sort of like how aura files call under HTML language tag).

packages/code-analyzer-stylelint-engine/package.json
```diff
@@ -14,7 +14,9 @@
   "types": "dist/index.d.ts",
   "dependencies": {
     "@types/node": "^20.0.0",
-    "@salesforce/code-analyzer-engine-api": "0.23.0"
+    "@salesforce/code-analyzer-engine-api": "0.23.0",
+    "@types/stylelint": "^13.13.3",
+    "stylelint": "^16.18.0"
```
stephen-carter-at-sf, line 19: Note that I checked https://gus.lightning.force.com/lightning/o/ADM_Third_Party_Software__c/list and it appears that stylelint v16 is one of the approved 3pp libraries - so this is good.

Note that it looks like 3 days ago they published a new version. So let's get this up to date - I think 16.19.1 is what we want.

packages/code-analyzer-stylelint-engine/src/engine.ts
```diff
@@ -16,44 +29,58 @@ export class StylelintEngine extends Engine {
...
-        const pathToPackageJson: string = path.join(__dirname, '..', 'package.json');
-        const packageJson: {version: string} = JSON.parse(await fsp.readFile(pathToPackageJson, 'utf-8'));
+        const pathToPackageJson: string = path.join(
+            __dirname,
+            "..",
+            "package.json"
+        );
+        const packageJson: { version: string } = JSON.parse(
+            await fsp.readFile(pathToPackageJson, "utf-8")
+        );
```
stephen-carter-at-sf, line 39: Let's try to avoid updating the contents of files for stylelistic reasons. It provides noise in the code review. And it looks like the style before is actually preferable.

packages/code-analyzer-stylelint-engine/src/engine.ts
```diff
@@ -16,44 +29,58 @@ export class StylelintEngine extends Engine {
...
         return packageJson.version;
     }
 
     // *** Remove underscore for private naming convention if you need any of the DescribeOptions
-    async describeRules(_describeOptions: DescribeOptions): Promise<RuleDescription[]> {
+    async describeRules(
+        _describeOptions: DescribeOptions
+    ): Promise<RuleDescription[]> {
```
stephen-carter-at-sf, line 46: ditto - no changes here. Maybe your editor is applying style rules automatically during development - making unnecessary noise.

packages/code-analyzer-stylelint-engine/src/engine.ts
```diff
@@ -16,44 +29,58 @@ export class StylelintEngine extends Engine {
...
         // *** Best Practice - Use RunRulesProgressEvents to keep users informed on the scan progress
         this.emitRunRulesProgressEvent(0);
 
         // *** Parse out relevant file types if you engine is language-specific
         //const relevantFiles: string[] | undefined;
```
stephen-carter-at-sf, line 51: Since you are not looking at the optional workspace provided by the DescribeOptions, then please make this PR more specific saying that it that you are implementing the describeRules method without workspace awareness as a first step.

packages/code-analyzer-stylelint-engine/src/engine.ts
```diff
@@ -1,7 +1,20 @@
...
+    LogLevel,
+    RuleDescription,
+    RunOptions,
+    SeverityLevel,
+    Violation,
+} from "@salesforce/code-analyzer-engine-api";
+import * as fsp from "node:fs/promises";
 import path from "path";
 import { getMessage } from "./messages";
+import stylelint, { RuleMeta } from "stylelint";
```
stephen-carter-at-sf, line 15: When importing from 3pp libraries like this, I recommend doing:
```
import * as stylelint from "stylelint";
```
this way, on line 61 you should be doing
```
const ruleMetadata: stylelint.RuleMeta ...`
```
so that we don't conflict with any other possible RuleMeta classes. Otherwise when I read RuleMeta I assume it is a type from this local package or the engine api.

packages/code-analyzer-stylelint-engine/src/engine.ts
```diff
@@ -16,44 +29,58 @@ export class StylelintEngine extends Engine {
...
         // *** Best Practice - Use RunRulesProgressEvents to keep users informed on the scan progress
         this.emitRunRulesProgressEvent(0);
 
         // *** Parse out relevant file types if you engine is language-specific
         //const relevantFiles: string[] | undefined;
 
-        // *** Get your rules!
         const ruleDescriptions: RuleDescription[] = [];
+        // Get all available rules from stylelint
+        const allRules = stylelint.rules;
```
stephen-carter-at-sf, line 55: Avoid untyped assignments please.

In this case, we don't need to specify the exact type that stylelint is using for its rules property - but instead just specify a type associated with what we are consuming from allRules.

Because now when I look at your line 60 I don't know what rule consists of. 

It looks like you are using stylelint.MetaData from rulemeta and that's it.

So for now, I recommend making the type as follows:
```
        type StyleLintRule = {ruleName: string, meta?: stylelint.RuleMeta};
        const styleLintRuleMap: Record<string, Promise<StyleLintRule>> = stylelint.rules;
```
This will help eslint, the compiler, and a human know why you are using an await when referencing things.

Also, you can make this code much cleaner if you switch to a filter and a map. Here is what you can basically get it down to I believe:

```
        type StyleLintRule = {ruleName: string, meta?: stylelint.RuleMeta};
        const styleLintRuleMap: Record<string, Promise<StyleLintRule>> = stylelint.rules;
        const styleLintRulePromises: Promise<StyleLintRule>[] = Object.values(styleLintRuleMap);
        const styleLintRules: StyleLintRule[] = await Promise.all(styleLintRulePromises);
        const ruleDescriptions: RuleDescription = styleLintRules
            .filter(r => r.meta !== undefined)
            .map(r => toRuleDescription(r.ruleName, r.meta));
```

packages/code-analyzer-stylelint-engine/src/engine.ts
stephen-carter-at-sf: When we get to the runRules method, most likely we'll need to generate the list that the user selected.

 But for adding in the "Recommended" tag or not, that is what the rule-mappings.ts file strategy is for - since we'll also need to assign default tags and severity levels for each rule anyway.

packages/code-analyzer-stylelint-engine/src/engine.ts
stephen-carter-at-sf: So we might also use the deprecated information to decide whether or not to make the rule available or not. That is, we might want to add a filter on `r=> r.meta.deprecated !== true` for example.

I'm glad we have the url - so that's good.

Do you know where the description message lives then? What exactly is in the messages property of the Rule? Is that just for violation messages or is one of those a rule description? Basically the rule description typically is what they use when generating their documentation page. So you might want to take a look at the stylelint repo and figure out how they are generating their documentation. I'm assuming this metadata must live somewhere.
Or does each rule have a packaged README file with a description section or something? How does it work? I wonder if we'll have to hardcode the rule descriptions for now - or figure out a way of extracting this during a build step into a consumable descriptions file.

stephen-carter-at-sf (changes requested):

packages/code-analyzer-stylelint-engine/test/test-data/temp-goldfile.json
```diff
@@ -1,11 +1,1790 @@
...
+  },
+  {
+    "name": "annotation-no-unknown",
+    "severityLevel": 2,
+    "tags": [
+      "ErrorProne",
+      "CSS",
+      "SCSS"
+    ],
+    "description": "annotation-no-unknown",
```
stephen-carter-at-sf, line 23: This gold file should be called something like `all-rules.goldfile.json` or `base-rules.goldfile.json` or whatever we want to call this core set.

We should have description here - not the names of the the rules obviously. We should not have SCSS tags. And we should have this entire file reflect the document that a PM should be reviewing to ensure we have the correct severity levels and tags that we want for each rule in our rule-mappings.ts file.

packages/code-analyzer-stylelint-engine/test/engine.test.ts
```diff
@@ -1,71 +1,307 @@
...
 import path from "path";
 import { StylelintEngine } from "../src/engine";
 import { changeWorkingDirectoryToPackageRoot } from "./test-helpers";
+import { RULE_MAPPINGS as _RULE_MAPPINGS } from "../src/rule-mappings";
+import { StylelintRuleStatus as _StylelintRuleStatus } from "../src/enums";
+import _stylelint from "stylelint";
+import { toSeverityLevel, toTags } from "../src/engine";
 
-changeWorkingDirectoryToPackageRoot();
+jest.mock("stylelint", () => ({
```
stephen-carter-at-sf, line 19: let's not mock out stylelint at all. If you want to provide a custom rule... then do so with an actual example in the test-data directory.

stephen-carter-at-sf (changes requested):

packages/code-analyzer-stylelint-engine/test/engine.test.ts
```diff
@@ -1,71 +1,307 @@
...
+    COMMON_TAGS,
+} from "@salesforce/code-analyzer-engine-api";
 import fs from "node:fs";
 import * as os from "node:os";
 import path from "path";
 import { StylelintEngine } from "../src/engine";
 import { changeWorkingDirectoryToPackageRoot } from "./test-helpers";
+import { RULE_MAPPINGS as _RULE_MAPPINGS } from "../src/rule-mappings";
+import { StylelintRuleStatus as _StylelintRuleStatus } from "../src/enums";
+import _stylelint from "stylelint";
```
stephen-carter-at-sf, line 16: Please update these imports so we only have the 2 styles of imports:

import {a, b, c} from "xyz"
or
import * as xyz from "xyz"

and avoid any default imports like
import abc from "xyz".

packages/code-analyzer-stylelint-engine/src/engine.ts
kenamverma: This is one of the rules in stylelint: [alpha-value-notation](https://github.com/stylelint/stylelint/blob/main/lib/rules/alpha-value-notation/index.cjs). Taking this example:

1. messages only contain the violation messages: https://github.com/stylelint/stylelint/blob/main/lib/rules/alpha-value-notation/index.cjs#L19-L21
2. description is the first line under the ruleName: https://github.com/stylelint/stylelint/blob/main/lib/rules/alpha-value-notation/README.md. I believe they are picking up README for their doc. We always know that the first line after the ruleName in README will be the description.

PR #283, opened 2025-05-01
CHANGE @W-18396029@ Running tests against only affected packages

jfeingold35:
This PR makes the following changes:
- The `verify-pr.yml` job can now identify which specific packages are affected by the changes in the PR, and will only run the tests in those packages.

See [this Github Action run from an experimental branch](https://github.com/forcedotcom/code-analyzer-core/actions/runs/14797216614) to see it only running tests for `code-analyzer-regex-engine` and `code-analyzer-flow-engine`, and see the checks for this PR itself for an example of it running on all packages because no specific package was changed.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -7,6 +7,9 @@ charset = utf-8
 trim_trailing_whitespace = true
 insert_final_newline = true
 
+[package.json]
+indent_size = 2
+
 [*.java]
 indent_style = space
 indent_size = 4
--- a/.github/workflows/verify-pr.yml
+++ b/.github/workflows/verify-pr.yml
@@ -72,6 +72,9 @@ jobs:
       - name: Support long paths
         run: git config --global core.longpaths true
       - uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+          fetch-depth: 0
       - uses: actions/setup-java@v4
         with:
           distribution: 'temurin'
@@ -83,4 +86,12 @@ jobs:
         with:
           python-version: 3.12
       - run: npm install
-      - run: npm run all
+      - name: Test changes in affected packages
+        run: |
+          BASE_SHA=${{ github.event.pull_request.base.sha }}
+          HEAD_SHA=${{ github.event.pull_request.head.sha }}
+          CHANGED_FILES=`git diff --name-only $HEAD_SHA $BASE_SHA`
+          node ./.github/workflows/verify-pr/identify-affected-workspaces.mjs "$CHANGED_FILES"
+          npm run build `cat workspace-args.txt`
+          npm run test `cat workspace-args.txt`
+          npm run lint
--- a/.github/workflows/verify-pr.yml
+++ b/.github/workflows/verify-pr.yml
@@ -54,8 +54,8 @@ jobs:
         run: |
           BASE_SHA=${{ github.event.pull_request.base.sha }}
           HEAD_SHA=${{ github.event.pull_request.head.sha }}
-          CHANGED_FILES=`git diff --name-only $HEAD_SHA $BASE_SHA`
-          node ./.github/workflows/verify-pr/validate-changed-package-versions.js "$CHANGED_FILES"
+          git diff --name-only $HEAD_SHA $BASE_SHA > changed_files.txt
+          node ./.github/workflows/verify-pr/validate-changed-package-versions.js changed_files.txt
       - name: Validate that packages properly depend on each other
         if: ${{ needs.check_for_postrelease_keyword.outputs.is-postrelease == 'false' }}
         run: |
@@ -90,8 +90,8 @@ jobs:
         run: |
           BASE_SHA=${{ github.event.pull_request.base.sha }}
           HEAD_SHA=${{ github.event.pull_request.head.sha }}
-          CHANGED_FILES=`git diff --name-only $HEAD_SHA $BASE_SHA`
-          node ./.github/workflows/verify-pr/identify-affected-workspaces.mjs "$CHANGED_FILES"
+          git diff --name-only $HEAD_SHA $BASE_SHA > changed_files.txt
+          node ./.github/workflows/verify-pr/identify-affected-workspaces.mjs changed_files.txt
           npm run build `cat workspace-args.txt`
           npm run test `cat workspace-args.txt`
           npm run lint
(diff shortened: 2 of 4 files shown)


.editorconfig
```diff
@@ -7,6 +7,9 @@ charset = utf-8
 trim_trailing_whitespace = true
 insert_final_newline = true
 
+[package.json]
```
jfeingold35, line 10: I got sick of fighting with my IDE about formatting on `package.json` files, so I just updated the editor config file about it.

.github/workflows/verify-pr.yml
```diff
@@ -72,6 +72,9 @@ jobs:
       - name: Support long paths
         run: git config --global core.longpaths true
       - uses: actions/checkout@v4
+        with:
```
jfeingold35, line 75: We're still checking out the same ref, but we need to do it like the `validate-packages` job does where the history is pulled too.

.github/workflows/verify-pr.yml
```diff
@@ -83,4 +86,12 @@ jobs:
...
       - run: npm install
-      - run: npm run all
+      - name: Test changes in affected packages
+        run: |
+          BASE_SHA=${{ github.event.pull_request.base.sha }}
+          HEAD_SHA=${{ github.event.pull_request.head.sha }}
+          CHANGED_FILES=`git diff --name-only $HEAD_SHA $BASE_SHA`
+          node ./.github/workflows/verify-pr/identify-affected-workspaces.mjs "$CHANGED_FILES"
+          npm run build `cat workspace-args.txt`
+          npm run test `cat workspace-args.txt`
```
jfeingold35, line 96: I tried assigning `cat workspace-args.txt` to a variable and then using that, but for whatever reason that didn't want to work, so I'm doing this instead. It's a little cumbersome but it'll work.

.github/workflows/verify-pr.yml
```diff
@@ -83,4 +86,12 @@ jobs:
...
+          npm run lint
```
jfeingold35, line 97: LInting is fast, so there's no reason not to lint all the files.

.github/workflows/verify-pr.yml
```diff
@@ -83,4 +86,12 @@ jobs:
         with:
           python-version: 3.12
       - run: npm install
-      - run: npm run all
```
jfeingold35, line 86: Technically we're losing the invocation of `npm package` that is contained by `npm run all`. If we want it, then I can put it back in.

.github/workflows/verify-pr/identify-affected-workspaces.mjs
```diff
@@ -0,0 +1,104 @@
...
+        fs.writeFileSync(tmpFilePath, '');
+        console.log(`WROTE EMPTY WORKSPACE ARGS TO ${tmpFilePath}`);
+        process.exit(0);
+    }
+    displayList('THE FOLLOWING FILES WERE CHANGED:', changedFiles);
+
+    const changedPackages = identifyChangedPackages(changedFiles);
+    if (changedPackages.length === 0) {
+        console.log(`No package-level changes. Using empty workspace arg to test all packages.`);
+        fs.writeFileSync(tmpFilePath, '');
```
jfeingold35, line 24: In this case (and the similar case above), the file will be empty, meaning that the GHA will run `npm run test `,  causing all packages to be tested.
I had to choose between "changing only top-level files causes no packages to be tested" and "changing only top-level files causes all packages to be tested", and I chose the latter because it was easier to implement.
Also, corner case: if you change top-level files and also one package-level file, then only that package and its dependents will be tested. Are we okay with that?

.github/workflows/verify-pr/identify-affected-workspaces.mjs
```diff
@@ -0,0 +1,104 @@
...
+    if (dependentPackages.length > 0) {
+        displayList('THE FOLLOWING PACKAGES HAVE DEPENDENCIES ON CHANGED PACKAGES:', dependentPackages);
+    } else {
+        console.log(`NO PACKAGES HAVE DEPENDENCIES ON CHANGED PACKAGES.\n`);
+    }
+
+    const affectedPackages = [...(new Set([...changedPackages, ...dependentPackages]).keys())];
+    displayList('BASED ON THE ABOVE, THE FOLLOWING PACKAGES ARE AFFECTED BY CHANGES, AND WILL REQUIRE TESTING:', affectedPackages);
+
+    const correspondingWorkspaceArgs = affectedPackages.map(name => `--workspace ${name}`);
```
jfeingold35, line 40: When changes can be tied to specific packages (and the ones that depend on those packages), we can construct a bunch of `--workspace blah` arguments and write them to a file, then we can use that file to turn `npm run test` into `npm run test --workspace blah1 --workspace blah2` and thereby only run the tests in the affected workspaces.

.github/workflows/verify-pr.yml
```diff
@@ -83,4 +86,12 @@ jobs:
...
+      - name: Test changes in affected packages
+        run: |
+          BASE_SHA=${{ github.event.pull_request.base.sha }}
+          HEAD_SHA=${{ github.event.pull_request.head.sha }}
+          CHANGED_FILES=`git diff --name-only $HEAD_SHA $BASE_SHA`
+          node ./.github/workflows/verify-pr/identify-affected-workspaces.mjs "$CHANGED_FILES"
```
stephen-carter-at-sf: Will we have any issues with clipping if the number of changed files is massive?
Would it be better to output this to a file and then read it from the script instead?

.github/workflows/verify-pr/identify-affected-workspaces.mjs
```diff
@@ -52,6 +52,10 @@ function displayList(header, list) {
     console.log('');
 }
 
+function readChangedFilesFile(changedFilesFileName) {
+    return fs.readFileSync(path.join(__dirname, '..', '..', '..', changedFilesFileName), 'utf-8').split('\n').map(s => s.trim());
```
stephen-carter-at-sf: Can you just use pathToRoot here instead of using __dirname, .., .., .. again?

stephen-carter-at-sf (approved):

.github/workflows/verify-pr/identify-affected-workspaces.mjs
jfeingold35: Yeah, good call. Thanks.

PR #202, opened 2025-02-12
FIX (Regex) @W-17809567@ Fixed ENFILE/EMFILE errors for Regex engine

jfeingold35:
This PR resolves https://github.com/forcedotcom/sfdx-scanner/issues/1740.
The root cause is that the `regex` engine was attempting to process every single file in the workspace in parallel via `async/await` and `Promise.all` instead of doing it serially. In a truly massive workspace (like, >20k files), it's possible to hit the system-level limit on how many files are allowed to be open at once.
The fix is that the engine has been modified to process files in batches of 1000, allowing for a more controlled level of parallelism.
I couldn't think of a way to add an automated test for this, short of adding 20,000 new files to repo, so I chose not to.

PR #287, opened 2025-05-08
FIX(regex): @W-18464445@: Use utility based off of p-limit instead of…

stephen-carter-at-sf:
… batching and replace all Promise.all calls for safety

Note that this is an improved fix over https://github.com/forcedotcom/code-analyzer-core/pull/202

### DIFF
--- a/packages/code-analyzer-regex-engine/package.json
+++ b/packages/code-analyzer-regex-engine/package.json
@@ -15,7 +15,8 @@
   "dependencies": {
     "@salesforce/code-analyzer-engine-api": "0.23.0",
     "@types/node": "^20.0.0",
-    "isbinaryfile": "^5.0.4"
+    "isbinaryfile": "^5.0.4",
+    "p-limit": "^3.1.0"
   },
   "devDependencies": {
     "@eslint/js": "^8.57.1",
--- a/packages/code-analyzer-regex-engine/src/engine.ts
+++ b/packages/code-analyzer-regex-engine/src/engine.ts
@@ -14,7 +14,7 @@ import * as fsp from 'node:fs/promises';
 import os from "node:os";
 import {RegexRule, RegexRules} from "./config";
 import {isBinaryFile} from "isbinaryfile";
-import {convertToRegex} from "./utils";
+import {convertToRegex, PromiseExecutionLimiter} from "./utils";
 
 const TEXT_BASED_FILE_EXTS = new Set<string>(
     [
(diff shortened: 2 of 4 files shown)


stephen-carter-at-sf:
fixes https://github.com/forcedotcom/sfdx-scanner/issues/1791

packages/code-analyzer-regex-engine/package.json
```diff
@@ -15,7 +15,8 @@
   "dependencies": {
     "@salesforce/code-analyzer-engine-api": "0.23.0",
     "@types/node": "^20.0.0",
-    "isbinaryfile": "^5.0.4"
+    "isbinaryfile": "^5.0.4",
+    "p-limit": "^3.1.0"
```
stephen-carter-at-sf, line 19: Note that p-limit v4 and higher are strict ESM modules so we need to stick with version 3.

@randi274 FYI.

And we already were shipping p-limit 3 as an indirect dependency... so this doesn't increase the size of our package at all.

packages/code-analyzer-regex-engine/src/engine.ts
```diff
@@ -23,18 +23,13 @@ const TEXT_BASED_FILE_EXTS = new Set<string>(
...
- */
-const FILE_BATCH_SIZE = 1000;
-
 export class RegexEngine extends Engine {
     static readonly NAME = "regex";
     private readonly regexRules: RegexRules;
     private readonly regexValues: Map<string, RegExp> = new Map();
     private readonly textFilesCache: Map<string, string[]> = new Map();
     private readonly ruleResourceUrls: Map<string, string[]>;
+    private readonly promiseLimiter: PromiseExecutionLimiter = new PromiseExecutionLimiter();
```
stephen-carter-at-sf, line 32: Note using a single one on the engine - so that we don't create a bunch of independent limit functions. Like we had a promise.all that calls a promise.all and if we did 1000 limit on the outer and then a 1000 limit on the inner... then that would be 1000000 limit in reality. So instead, we just keep an overall 1000 Promise limit for the entire thing by using the same instance of my PromiseExecutionLimiter.

packages/code-analyzer-regex-engine/src/engine.ts
```diff
@@ -99,26 +94,18 @@ export class RegexEngine extends Engine {
...
-        const violations: Violation[] = [];
-        while (batchMultiplier * FILE_BATCH_SIZE < textFiles.length) {
-            // Turns out there's a system-level limit on how many files can be open at once. In order to avoid hitting
-            // this limit while still reaping the benefits of `Promise.all()`, we'll process the files in batches.
-            const textFileBatch = textFiles.slice(batchMultiplier * FILE_BATCH_SIZE, (batchMultiplier + 1) * FILE_BATCH_SIZE);
-            const ruleRunPromises: Promise<Violation[]>[] = textFileBatch.map(file => this.runRulesForFile(file, ruleNames));
-            const newViolations = (await Promise.all(ruleRunPromises)).flat();
-            violations.push(...newViolations);
-            batchMultiplier++;
-        }
```
stephen-carter-at-sf, line 112: Removing the old batching because in reality, when a batch had a 1000 - then when 999 were finished... we were waiting for just the 1 before we created another batch of 1000.

With p-limit we guarantee that 1000 are always running. So when 1 finished... it allows the next one to come in to keep 1000 running at any given time.

So we actually get a performance boost from this change away from batching.

jfeingold35 (approved):
The function-based setup is a little counterintuitive, but it seems fine to me. Approved.

PR #290, opened 2025-05-12
NEW: @W-18495430@: Update devDependencies to use eslint v9

stephen-carter-at-sf:
Note that this the code-analyzer-eslint8-engine module obviously will be stuck on eslint v8. And note that this simply prepares the way for us to use eslint 9 with our code-analyzer-eslint-engine but that engine implementation is still to come.

### DIFF
--- a/packages/ENGINE-TEMPLATE/package.json
+++ b/packages/ENGINE-TEMPLATE/package.json
@@ -39,7 +39,7 @@
     "test": "jest --coverage",
     "lint": "eslint src/**/*.ts",
     "package": "npm pack",
-    "all": "npm run build && npm run test && npm run lint && npm run package",
+    "all": "npm run build && npm run lint && npm run test && npm run package",
     "clean": "tsc --build tsconfig.build.json --clean",
     "postclean": "rimraf dist && rimraf coverage && rimraf ./*.tgz && rimraf vulnerabilities",
     "scrub": "npm run clean && rimraf node_modules",
--- a/packages/ENGINE-TEMPLATE/package.json
+++ b/packages/ENGINE-TEMPLATE/package.json
@@ -14,17 +14,17 @@
   "types": "dist/index.d.ts",
   "dependencies": {
     "@types/node": "^20.0.0",
-    "@salesforce/code-analyzer-engine-api": "0.23.0"
+    "@salesforce/code-analyzer-engine-api": "0.24.0-SNAPSHOT"
   },
   "devDependencies": {
-    "@eslint/js": "^8.57.1",
+    "@eslint/js": "^9.26.0",
     "@types/jest": "^29.5.14",
-    "eslint": "^8.57.1",
+    "eslint": "^9.26.0",
     "jest": "^29.7.0",
     "rimraf": "*",
-    "ts-jest": "29.3.2",
+    "ts-jest": "^29.3.2",
     "typescript": "^5.8.3",
-    "typescript-eslint": "^8.30.1"
+    "typescript-eslint": "^8.32.1"
   },
   "engines": {
     "node": ">=20.0.0"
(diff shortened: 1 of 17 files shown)


packages/ENGINE-TEMPLATE/package.json
```diff
@@ -14,17 +14,17 @@
...
   "devDependencies": {
-    "@eslint/js": "^8.57.1",
+    "@eslint/js": "^9.26.0",
     "@types/jest": "^29.5.14",
-    "eslint": "^8.57.1",
+    "eslint": "^9.26.0",
     "jest": "^29.7.0",
     "rimraf": "*",
-    "ts-jest": "29.3.2",
+    "ts-jest": "^29.3.2",
```
stephen-carter-at-sf, line 25: Back with https://github.com/forcedotcom/code-analyzer-core/pull/261/files#diff-8c263af063232c1af728f80b0ebe7407da27dc2a4805c3f9c8a75725f81046bd it looks like @randi274 forgot to put back the `^` on `ts-jest` after we were unblocked. So fixing that here as well.

packages/code-analyzer-eslint8-engine/package.json
```diff
@@ -35,9 +35,9 @@
     "cross-env": "^7.0.3",
     "jest": "^29.7.0",
     "rimraf": "*",
-    "ts-jest": "29.3.2",
+    "ts-jest": "^29.3.2",
     "typescript": "^5.8.3",
-    "typescript-eslint": "^8.30.1",
+    "typescript-eslint": "8.30.1",
```
stephen-carter-at-sf: Forgot to lock this in for the eslint8-engine ... so eslint8-engine will have everything it needs in its own node_modules folder local to its package separate from top level one.

packages/code-analyzer-eslint8-engine/src/strategy.ts
```diff
@@ -181,7 +181,7 @@ export class LegacyESLintStrategy implements ESLintStrategy {
             overrideConfigFile: userConfigInfo.getUserConfigFile(),                              // This is applied third.
             overrideConfig: overrideConfig as Linter.Config,                                     // This is applied fourth (on top).
             ignorePath: userConfigInfo.getUserIgnoreFile()
-        };
+        } as ESLint.Options;
```
stephen-carter-at-sf, line 184: Makes IDE Happy because IDE struggles a little bit resolving this one type with both versions of ESLint installed.

jfeingold35 (approved):

packages/code-analyzer-eslint8-engine/package.json
```diff
@@ -13,18 +13,18 @@
...
-    "@salesforce/code-analyzer-engine-api": "0.23.0",
+    "@salesforce/code-analyzer-engine-api": "0.24.0-SNAPSHOT",
     "@salesforce/eslint-config-lwc": "3.7.2",
     "@salesforce/eslint-plugin-lightning": "1.0.1",
     "@types/eslint": "8.56.12",
     "@types/node": "^20.0.0",
-    "@typescript-eslint/eslint-plugin": "8.30.1",
-    "@typescript-eslint/parser": "8.30.1",
+    "@typescript-eslint/eslint-plugin": "8.32.1",
+    "@typescript-eslint/parser": "8.32.1",
```
stephen-carter-at-sf: There seemed to be some sort of odd issue with the 8.30.1 version. I was seeing the same error as shown over at https://github.com/typescript-eslint/typescript-eslint/issues/10353 with https://github.com/forcedotcom/code-analyzer-core/actions/runs/14983203873/job/42091861005?pr=290#step:8:1178

Upgrading fixed this... but it also pulled in the 1 new typescript rule - which is good.

stephen-carter-at-sf: OK actually fixed the issue in a different way. To ensure that @typescript-eslint/eslint-plugin doesn't show up at the top level (because it can depend back on eslint - which has the other version at the top level) I instead filled in the dependencies for the eslint-engine to be different than the eslint8-engine to force the eslint8-engine to not hoise its dependencies up. Resolved.

PR #292, opened 2025-05-13
NEW: @W-18495437@: Detect state of user's eslint configuration files …

stephen-carter-at-sf:
…and emit fine log


packages/code-analyzer-eslint-engine/test/user-config-info.test.ts
```diff
@@ -0,0 +1,607 @@
...
+const TEST_DATA_FOLDER: string = path.join(__dirname, 'test-data');
+const WORKSPACE_WITH_LEGACY_CONFIG1: string = path.join(TEST_DATA_FOLDER, 'workspaceWithLegacyConfig1');
+const WORKSPACE_WITH_LEGACY_CONFIG2: string = path.join(TEST_DATA_FOLDER, 'workspaceWithLegacyConfig2');
+const WORKSPACE_WITH_LEGACY_CONFIG3: string = path.join(TEST_DATA_FOLDER, 'workspaceWithLegacyConfig3');
+const WORKSPACE_WITH_LEGACY_IGNORE: string = path.join(TEST_DATA_FOLDER, 'workspaceWithLegacyIgnoreFile');
+const WORKSPACE_WITH_FLAT_CONFIG1: string = path.join(TEST_DATA_FOLDER, 'workspaceWithFlatConfig1');
+const WORKSPACE_WITH_FLAT_CONFIG2: string = path.join(TEST_DATA_FOLDER, 'workspaceWithFlatConfig2');
+const WORKSPACE_WITH_FLAT_CONFIG3: string = path.join(TEST_DATA_FOLDER, 'workspaceWithFlatConfig3');
+
+describe('Tests for the UserConfigInfo class', () => {
```
stephen-carter-at-sf, line 16: This file is rather exhaustive to cover all the different scenarios for the most part.

packages/code-analyzer-eslint-engine/src/messages.ts
```diff
@@ -98,7 +98,10 @@ const MESSAGE_CATALOG : { [key: string]: string } = {
         `  engines:\n` +
         `    eslint:\n` +
         `      eslint_ignore_file: "%s"\n` +
-        `Alternatively, to have Code Analyzer automatically discover and apply any ESLint configuration and ignore files found in your workspace, set the auto_discover_eslint_config value to true.`
+        `Alternatively, to have Code Analyzer automatically discover and apply any ESLint configuration and ignore files found in your workspace, set the auto_discover_eslint_config value to true.`,
+
+    IgnoringFlatConfigFile:
+        `Ignoring '%s' since ESLint "Flat" configuration files are not yet supported by this version of the 'eslint' engine.`
```
stephen-carter-at-sf, line 104: Note that this doesn't actually get displayed unless the file is auto discovered when auto_discover_eslint_config is enabled. The config still errors today when a user tries to explicitely set a flat config file.

packages/code-analyzer-eslint-engine/src/user-config-info.ts
```diff
@@ -0,0 +1,96 @@
...
+
+    toString(): string {
+        return JSON.stringify({
+            state: this.getState(),
+            userConfigFile: this.getUserConfigFile() || null,
+            userIgnoreFile: this.getUserIgnoreFile() || null
+        });
+    }
+
+    private initIfNeeded(): void {
```
jfeingold35, line 53: Question: Is this lazy-loading just a performance measure, or is there a functional reason for it? If it's just performance, I'd honestly say it's premature optimization and just load it at create time so that you can have `userConfigState` not be `?`-annotated.

packages/code-analyzer-eslint-engine/test/engine.test.ts
```diff
@@ -330,32 +358,38 @@ describe('Typical tests for the runRules method of ESLintEngine', () => {
     //        * When custom eslint config exists but is not applied, then runRules emits info message
     //        * When runRules is called on workspace with a config that ignores files and auto discover is true, then those files are ignored
     //        * When runRules is called and a ".eslintignore" file is provided that ignores files, then those files are ignored
-    // For now we just maintain a test that the v9 eslint engine errors until it has been implemented:
-    it('When directly calling describeRules on the v9 ESLintEngine, then we error since it has not been implemented', async () => {
-        const engine: ESLintEngine = new ESLintEngine(DEFAULT_CONFIG);
+    // For now we just maintain a test that the v9 eslint engine defers when auto discovered flat config is discovered until it has been implemented:
```
jfeingold35: "Defers" is unclear to me. What's it deferring to? V8?

packages/code-analyzer-eslint-engine/test/engine.test.ts
```diff
@@ -418,8 +452,6 @@ function createRunOptions(workspace: Workspace): RunOptions {
     }
 }
 
-// To be used temporarily while we are still migrating tests from using v8 to v9.
```
jfeingold35, line 421: Seems to me that this comment is still relevant. What was it removed?

packages/code-analyzer-eslint-engine/test/test-data/workspaceWithFlatConfig1/dummy1.js
jfeingold35, line 1: Looks like the only difference between these three workspaces is the file extension of the config. Could we rename the workspaces to `workspaceWithJsConfig`, `workspaceWithMjsConfig`, and `workspaceWithCjsConfig` for clarity?

packages/code-analyzer-eslint-engine/test/test-data/workspaceWithLegacyConfig2/.eslintrc.yml
jfeingold35, line 1: Again, seems like the only difference in these workspaces is that the config is of a different suffix. Can we rename the workspaces to make that clearer?

packages/code-analyzer-eslint-engine/src/user-config-info.ts
stephen-carter-at-sf: 3 reasons for this:
1) It is an antipattern if constructors do work (which is why constructors can't be async by the way)... so doing the "work" to calculate state is done when a getter is called - not during construction.
2) We can have a circumstance where a userConfigFile or userIgnoreFile is undefined even after calculation... so forcing a central place to check (in this case userConfig state) makes sense to know whether to work is still needed to be done or not.
3) The configuration state calculation already requires us to do the work of looking up the config file or ignore file if auto discover is true. That is... we need to do all the work in one place anyway because they are actually all intertwined.

So with this in mind we have 2 choices... either do what I'm doing and use initIfNeeded for lazy initialization/caching. Or make the constructor to be private and instead use a static method to do the full initialization. The downsize of using a static method to do the initialization is that we have more variables to pass around into sub functions instead of being able to set them as state and use private methods. Thus I chose the first choice because it seemed cleaner.

But happy to switch to the static method approach if you think that makes sense. This is all just implementation detail anyway.

packages/code-analyzer-eslint-engine/test/test-data/workspaceWithFlatConfig1/dummy1.js
stephen-carter-at-sf: that is ambiguous though. js and cjs files can either be legacy or flat. 
But yeah I could rename things to be workspaceWithJsFlatConfig, workspaceWithCjsFlatConfig, workspaceWithMjsFlatConfig but the workspaceWithFlatConfig3 also has a legacy eslint ignore file in it so I can test that case as well. So I don't know how specific I want to make each of these scenarios. I was just thinking of scenario 1, 2, and 3. And I'll be adding to these soon - so it isn't just that 1 is js based, 2 is cjs based, and 3 is mjs based. I just wanted to have 3 scenarios.

Notice that in the LegacyConfig1, LegacyConfig2, and LegacyConfig3 cases I could have kept going with 4, 5, 6 etc... because I'm not even testing "yaml" for example. It's not fully exhaustive. It just gives me 3 slots to sprinkle in differences as needed moving forward. This sprinkling strategy avoids having to have a million test scenarios. So hopefully we'll be able to stick to 3 each and I can sprinkle what I need into these 3 to cover nearly all cases that actually might catch bugs.

packages/code-analyzer-eslint-engine/test/test-data/workspaceWithLegacyConfig2/.eslintrc.yml
stephen-carter-at-sf: Currently it is just the file extension. As I move tests back in from the "TODO"s section ... then these tests will change soon.
You are still seeing step 3 out of 6 here.

packages/code-analyzer-eslint-engine/test/engine.test.ts
stephen-carter-at-sf: Yeah. I can update this to make it more clear.

packages/code-analyzer-eslint-engine/test/engine.test.ts
stephen-carter-at-sf: No the use of createEngineFromPlugin is no longer temporary. Since the engine now receives in its constructor  a delegate eslint 8 engine. It's easiest to just test at this level for now and see the delegation as implementation detail.

packages/code-analyzer-eslint-engine/src/user-config-info.ts
jfeingold35: No, you've convinced me. This is fine.

packages/code-analyzer-eslint-engine/test/engine.test.ts
jfeingold35: Fair enough.

packages/code-analyzer-eslint-engine/test/test-data/workspaceWithFlatConfig1/dummy1.js
jfeingold35: Fair enough. If we can't have more detail in the names of the folders, could we maybe have a README in each folder that describes the scenarios it covers?

packages/code-analyzer-eslint-engine/test/test-data/workspaceWithLegacyConfig2/.eslintrc.yml
jfeingold35: Fair.

packages/code-analyzer-eslint-engine/test/test-data/workspaceWithFlatConfig1/dummy1.js
stephen-carter-at-sf: I just did the rename for now. Numbers, extensions... all the same to me.

packages/code-analyzer-eslint-engine/test/engine.test.ts
stephen-carter-at-sf: done.

jfeingold35 (approved):

PR #293, opened 2025-05-14
CHANGE: @W-18515717@: Make sure to wrap arguments that have spaces wi…

stephen-carter-at-sf:
…th quotes when using shell=true

### DIFF
--- a/packages/code-analyzer-retirejs-engine/tsconfig.json
+++ b/packages/code-analyzer-retirejs-engine/tsconfig.json
@@ -3,7 +3,8 @@
   "compilerOptions": {
     "composite": true,
     "outDir": "./dist",
-    "rootDir": "."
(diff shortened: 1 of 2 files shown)


packages/code-analyzer-retirejs-engine/tsconfig.json
```diff
@@ -3,11 +3,12 @@
   "compilerOptions": {
     "composite": true,
     "outDir": "./dist",
-    "rootDir": "."
+    "rootDir": ".",
+    "isolatedModules": true
```
stephen-carter-at-sf, line 7: Another fix as you go item since I was in here. It looks like @randi274 updated all other packages with isolatedModules: true but missed this one so it was producing warnings still.

jfeingold35 (approved):

PR #296, opened 2025-05-16
CHANGE @W-18308576@ Updated dependencies

jfeingold35:


packages/code-analyzer-pmd-engine/test/test-data/pmdGoldfiles/rules_apexOnly.goldfile.json
```diff
@@ -958,6 +958,19 @@
...
   },
+  {
+    "name": "TypeShadowsBuiltInNamespace",
+    "severityLevel": 2,
+    "tags": [
+      "Recommended",
+      "ErrorProne",
+      "Apex"
+    ],
+    "description": "This rule finds Apex classes, enums, and interfaces that have the same name as a class, enum, or interface in the `System` or `Schema` namespace. Shadowing these namespaces in this way can lead to confusion and unexpected behavior. Code that intends to reference a `System` or `Schema` class, enum, or interface may inadvertently reference the locally defined type instead. This can... Learn more: https://docs.pmd-code.org/pmd-doc-7.13.0/pmd_rules_apex_errorprone.html#typeshadowsbuiltinnamespace",
```
stephen-carter-at-sf: In the learn more... you should update 7.13.0 with `{{PMD_VERSION}}` as well.

packages/code-analyzer-sfge-engine/sfge/build.gradle.kts
```diff
@@ -35,6 +35,7 @@ dependencies {
     }
     implementation(libs.reflections)
     implementation(libs.asm)
+    implementation(libs.netty.all)
```
stephen-carter-at-sf: What's this for?

packages/code-analyzer-pmd-engine/test/test-data/pmdGoldfiles/rules_apexOnly.goldfile.json
jfeingold35: Good catch; copy-pasted that one.

packages/code-analyzer-sfge-engine/sfge/build.gradle.kts
jfeingold35: We got flagged for our dependency on an insecure version of `netty-handler`, which is part of `netty-all`, so I updated the transitive dependency on `netty-all` to keep the related modules in lockstep instead of just updating `netty-handler`.

stephen-carter-at-sf: To not make it look like a direct dependency, should you instead just use a constraints section? If that doesn't make sense - then at least just put in a comment saying that this dependency is to lock in an indirect dependency to a later version because the other one is vulnerable or something.

image

stephen-carter-at-sf: That might be groovy syntax Kotlin syntax might be:
image

jfeingold35: Fair enough. Stand by for that.

stephen-carter-at-sf (approved):

PR #298, opened 2025-05-19
FIX(eslint8): @W-18558198@: Use cwd as basedir if no base configs are used …

stephen-carter-at-sf:
…so that custom overrides work


stephen-carter-at-sf:
fixes: https://github.com/forcedotcom/code-analyzer/issues/1807

stephen-carter-at-sf:
Unfortunately, the only way to test this will be to create a temp folder and unzip a massive folder with a bunch of node modules. Since we will be moving towards ESLint 9 soon... I don't think it's worth adding in a big zip file for this one case. Instead, we should just QA this fix.

randi274 (approved):

PR #301, opened 2025-05-29
NEW(eslint): @W-18495508@: Implement the "no-user-config" scenario for ESLint v9

stephen-carter-at-sf:


packages/code-analyzer-engine-api/src/utils/fs-utils.ts
```diff
@@ -10,4 +12,45 @@ const tmpDirAsync = promisify((options: tmp.DirOptions, cb: tmp.DirCallback) =>
...
+}
+
+
+/**
+ * Returns the longest common parent folder of the provided paths.
+ *   If empty or if no common parent folder exists, like in the case on Windows machines of using two different drives
+ *   C: and D:, then null is returned (to allow the client of this function to handle this case without try/catch).
+ * @param paths It is assumed that paths is a non-empty array of absolute value paths.
+ */
+export function calculateLongestCommonParentFolderOf(paths: string[]): string | null {
```
stephen-carter-at-sf, line 24: Note that this was just moved (not modified in any way).
I had to share this utility with the eslint9 engine to help with the calculation of the base directory for ESLint 9. This meant a change to the engine-api package which meant all the engines in this monorepo needed to increment their dependency on it. Sorry for the noise.

packages/code-analyzer-eslint-engine/package.json
```diff
@@ -18,14 +18,16 @@
...
+    "@salesforce/code-analyzer-engine-api": "0.25.0-SNAPSHOT",
+    "@salesforce/code-analyzer-eslint8-engine": "0.3.0-SNAPSHOT",
     "@types/node": "^20.0.0",
+    "@types/tmp": "^0.2.6",
     "@typescript-eslint/eslint-plugin": "^8.32.1",
     "@typescript-eslint/parser": "^8.32.1",
     "eslint": "^9.26.0",
     "eslint-plugin-import": "^2.31.0",
     "eslint-plugin-jest": "^28.11.0",
+    "tmp": "^0.2.3",
```
stephen-carter-at-sf, line 30: I look forward to the day I have time to do W-17875752 so that we can just have core pass in the temp folder so that we don't need to have the tmp dependency on all our engines.

packages/code-analyzer-eslint-engine/src/rule-mappings.ts
```diff
@@ -636,6 +624,10 @@ export const RULE_MAPPINGS: Record<string, {severity: SeverityLevel, tags: strin
         severity: SeverityLevel.High,
         tags: [/* NOT RECOMMENDED */    COMMON_TAGS.CATEGORIES.ERROR_PRONE,    COMMON_TAGS.LANGUAGES.JAVASCRIPT] // Not available with TypeScript
     },
+    "no-useless-assignment": {
```
stephen-carter-at-sf, line 627: ESLint v9 has removed a few rules and added a few rules for base javascript and typescript. So the the updates to the gold files and rule-mappings reflect this.

packages/code-analyzer-eslint-engine/test/engine.test.ts
```diff
@@ -418,7 +419,8 @@ describe('Tests for emitting events', () => {
 
     it('When describeRules is called, then it emits correct progress events', async () => {
         await engine.describeRules(createDescribeOptions());
-        expect(describeRulesProgressEvents.map(e => e.percentComplete)).toEqual([0, 10, 40, 80, 100]);
+        // TODO: We should make our DescribeRulesProgressEvents more refined while calculating the eslint context information
+        expect(describeRulesProgressEvents.map(e => e.percentComplete)).toEqual([0, 10, 90, 95, 100]);
```
stephen-carter-at-sf, line 423: This file was filled with the tests for the default (no user config) case...and it's really great that the only thing behavior change was the progress events (which I plan on enhancing at some point by the way) and gold files.
This proves that between v8 and v9 - I have successfully maintained the default behavior!

packages/code-analyzer-eslint-engine/src/utils.ts
```diff
@@ -5,4 +5,56 @@ export function makeUnique(values: string[]): string[] {
...
+ *
+ * @param value - top level value to be sanitized
+ * @param replacer - function that can replace a value before it is sanitized
+ */
+export function makeStringifiable(
+    value: unknown,
+    replacer: (value: unknown, path: string) => unknown = v => v
+): object | null | string | number | boolean
+{
+    const seen = new WeakMap<object, string>();
```
jfeingold35, line 26: Just to make sure I understand, is the `WeakMap` being used here so that entries added during recursive calls fall out as the recursion un-nests?

packages/code-analyzer-eslint-engine/src/base-config.ts
```diff
@@ -0,0 +1,169 @@
...
+                        // config file. See https://typescript-eslint.io/packages/parser/#projectservice
+                        projectService: true
+                    }
+                }
+            });
+        }
+        return configs;
+    }
+
+    private useJsConfig(): boolean {
```
jfeingold35: Could we maybe call these methods something like `useBaseJsConfig` etc for clarity's sake?

packages/code-analyzer-eslint-engine/src/engine.ts
```diff
@@ -41,43 +55,216 @@ export class ESLintEngine extends Engine {
     }
 
     async describeRules(describeOptions: DescribeOptions): Promise<RuleDescription[]> {
+        this.emitDescribeRulesProgressEvent(0);
+
         const userConfigInfo: UserConfigInfo = this.getUserConfigInfo(describeOptions.workspace);
-        this.emitLogEvent(LogLevel.Fine, `Detected the following regarding the user's eslint configuration files: ${userConfigInfo}`);
+        this.emitLogEvent(LogLevel.Fine, `Detected the following user's ESLint configuration state: ${userConfigInfo}`);
```
jfeingold35: "Following user's configuration state" sounds like it's indicating who the configuration state belongs to, not that it's indicating the configuration state itself. Maybe "Following user-provided configuration state"?

packages/code-analyzer-eslint-engine/src/engine.ts
```diff
@@ -41,43 +55,216 @@ export class ESLintEngine extends Engine {
...
+        name: ruleName,
+        severityLevel: severityLevel,
+        tags: tags,
+        description: metadata.docs?.description || '',
+        resourceUrls: ruleUrl ? [ruleUrl] : []
+    }
+}
+
+function toSeverityLevelForCustomRule(metadata: RulesMeta, status: ESLintRuleStatus): SeverityLevel {
+    if (status === ESLintRuleStatus.WARN) {
```
jfeingold35, line 211: I understand why you made the choices you made, but it feels really weird to me that `warn` produces a lower-severity violation than `layout`. I feel like if `warn` is being turned to `info`, then so should `layout`.

packages/code-analyzer-eslint-engine/src/missing-types-for-dependencies.d.ts
```diff
@@ -0,0 +1,37 @@
+// This declaration adds in the missing types for "@lwc/eslint-plugin-lwc-platform" whose package.json file's main field points to:
+//     node_modules/@lwc/eslint-plugin-lwc-platform/lib/index.js
+declare module '@lwc/eslint-plugin-lwc-platform' {
```
jfeingold35, line 3: Is this being declared on top of modules that already exist, or are we declaring these modules from whole cloth?

stephen-carter-at-sf: See https://salesforce-internal.slack.com/archives/C5W3E40TC/p1746136385115359

We depend on those packages but they have no types. Until the other team adds in types for typescript, I created the types for them here locally with our own .d.ts file.

packages/code-analyzer-eslint-engine/src/engine.ts
stephen-carter-at-sf: An ESLint warn is meant to not fail eslint in any way but to just inform the user about something. This is indeed a lower severity that layout errors.

I didn't change these - these came from our original eslint 8 engine design which we already had reviewed, etc. Keeping the rules the exact same as they were before.

packages/code-analyzer-eslint-engine/src/engine.ts
stephen-carter-at-sf: Yeah I struggled with this... in fact I am finding this is a super noisy and won't provide much value anyway - so let me remove this log.

The log that I need to add in is displaying the eslint options. But I'll do that in a separate PR since it'll requires me somehow sending down the log emitter function down a few levels (which i'll do last).

But for now, i'll just remove this log line.

packages/code-analyzer-eslint-engine/src/base-config.ts
stephen-carter-at-sf: Sure. Done.

packages/code-analyzer-eslint-engine/src/utils.ts
stephen-carter-at-sf: I just know from experience that you need to use a WeakMap when doing infinite loop detection. I'll let chatGPT answer this one for me:

image

packages/code-analyzer-eslint-engine/src/engine.ts
stephen-carter-at-sf: Done.

stephen-carter-at-sf: Sure I can fix the wording here.

Done.

jfeingold35 (approved):

PR #305, opened 2025-06-05
FIX(regex): @W-18696814@: Lower p-limit for windows and only read fil…

stephen-carter-at-sf:
…e once

### DIFF
--- a/packages/code-analyzer-core/test/rule-selection.test.ts
+++ b/packages/code-analyzer-core/test/rule-selection.test.ts
@@ -290,9 +290,14 @@ describe('Tests for selecting rules', () => {
             workspace: new engApi.Workspace('FixedId', [path.resolve('src'), path.resolve('test')])
         };
         const stubEngine1: stubs.StubEngine1 = plugin.getCreatedEngine('stubEngine1') as stubs.StubEngine1;
-        expect(stubEngine1.describeRulesCallHistory).toEqual([{describeOptions: expectedDescribeOptions}]);
+        expect(stubEngine1.describeRulesCallHistory).toHaveLength(1);
+        expect(stubEngine1.describeRulesCallHistory[0].describeOptions.logFolder).toEqual(expectedDescribeOptions.logFolder);
+        expect(stubEngine1.describeRulesCallHistory[0].describeOptions.workspace!.getWorkspaceId()).toEqual('FixedId');
         const stubEngine2: stubs.StubEngine2 = plugin.getCreatedEngine('stubEngine2') as stubs.StubEngine2;
-        expect(stubEngine2.describeRulesCallHistory).toEqual([{describeOptions: expectedDescribeOptions}]);
+        expect(stubEngine2.describeRulesCallHistory).toHaveLength(1);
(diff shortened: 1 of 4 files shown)


packages/code-analyzer-regex-engine/src/engine.ts
```diff
@@ -101,29 +101,37 @@ export class RegexEngine extends Engine {
...
 
     private shouldScanFile(fileName: string, ruleName: string): boolean {
         const fileExtensions: string[] | undefined = this.regexRules[ruleName].file_extensions;
         return !fileExtensions || fileExtensions.some(ext => fileName.toLowerCase().endsWith(ext));
     }
 
-    private async scanFile(fileName: string, ruleName: string): Promise<Violation[]> {
+    private async scanFileContents(fileName: string, fileContents: string, metaDataContents: string, ruleName: string): Promise<Violation[]> {
         const violations: Violation[] = [];
-        let fileContents: string = await fs.promises.readFile(fileName, {encoding: 'utf8'})
```
stephen-carter-at-sf, line 118: Before we were reading the file for every single rule. Now I changed it so that we only read a file once and then run the regex for each of its rules the on the file contents.

packages/code-analyzer-regex-engine/src/engine.ts
```diff
@@ -101,30 +101,41 @@ export class RegexEngine extends Engine {
         };
     }
 
-    private async runRulesForFile(file: string, ruleNames: string[]): Promise<Violation[]>{
-        const rulesToRun: string[] = ruleNames.filter(rule => this.shouldScanFile(file, rule));
-        const violationPromiseFunctions:  (() => Promise<Violation[]>)[] = rulesToRun.map(
-            ruleName => () => this.scanFile(file, ruleName));
-        return (await this.promiseLimiter.execute(violationPromiseFunctions)).flat();
+    private async runRulesForFile(fileName: string, ruleNames: string[]): Promise<Violation[]>{
```
randi274, line 104: I'm surprised there weren't any test changes for this fix. Are there any that make sense to add as a result of this change?

stephen-carter-at-sf: What do you mean? These changes are implementation detail only changes for performance boost. I wouldn't expect a test to fail.

packages/code-analyzer-core/test/rule-selection.test.ts
```diff
@@ -290,9 +290,14 @@ describe('Tests for selecting rules', () => {
             workspace: new engApi.Workspace('FixedId', [path.resolve('src'), path.resolve('test')])
         };
         const stubEngine1: stubs.StubEngine1 = plugin.getCreatedEngine('stubEngine1') as stubs.StubEngine1;
-        expect(stubEngine1.describeRulesCallHistory).toEqual([{describeOptions: expectedDescribeOptions}]);
```
stephen-carter-at-sf, line 293: With cache warmed up, this test failed because it was doing an exact check on the workspace object. So reworked this test to just look to see if it has the workspace by id.

jfeingold35 (approved):

stephen-carter-at-sf:
fixes https://github.com/forcedotcom/code-analyzer/issues/1832

PR #306, opened 2025-06-05
NEW(eslint): @W-18669872@: Enhance progress events for eslint v9 engine

stephen-carter-at-sf:


packages/code-analyzer-eslint-engine/src/base-config.ts
```diff
@@ -104,7 +104,6 @@ export class BaseConfigFactory {
             configs.push({
                 ...conf,
                 files: this.engineConfig.file_extensions.typescript.map(ext => `**/*${ext}`),
-                ignores: this.engineConfig.file_extensions.javascript.map(ext => `**/*${ext}`), // TODO: Confirm whether this works or not
```
stephen-carter-at-sf, line 107: Yeah turns out this wasn't needed. :-)

packages/code-analyzer-eslint-engine/src/engine.ts
```diff
@@ -150,7 +152,13 @@ export class ESLintEngine extends Engine {
...
                 }
                 const violation: Violation = toViolation(eslintResult.filePath, resultMsg);
-                violations.push(violation);
+
+                if (specifiedRules.has(violation.ruleName)) {
+                    violations.push(violation);
+                } else {
+                    // This may be possible if a user tries to suppress an eslint rule in their code that isn't available. We just ignore it but debug it just in case.
+                    this.emitLogEvent(LogLevel.Debug, getMessage('ViolationFoundFromUnregisteredRule', violation.ruleName, JSON.stringify(violation,null,2)))
+                }
```
stephen-carter-at-sf, line 161: Fix as you go item: This code was with the eslint 8 engine... I didn't think we would need it with eslint 9's ruleFilter but as I did my full gambit of tests... it turns out we still do.

packages/code-analyzer-eslint-engine/src/messages.ts
```diff
@@ -66,6 +66,11 @@ const MESSAGE_CATALOG : { [key: string]: string } = {
         `The eslint engine encountered an unexpected error thrown from '%s':\n%s\n\n` +
         'ESLint options used:\n%s',
 
+    ViolationFoundFromUnregisteredRule:
+        `A rule with name '%s' produced a violation, but this rule was not registered with the 'eslint' engine so it will not be included in the results.\n` +
+        `This may occur if in your file you are using inline comments to attempt to disable or configure this rule even though it is unknown to ESLint and Code Analyzer.\n` +
+        `Ignored Violation:\n%s`,
```
stephen-carter-at-sf, line 72: This already exististed with eslint 8... just pulled it into eslint 9... so no need for UI text review here.

packages/code-analyzer-eslint-engine/src/base-config.ts
```diff
@@ -25,7 +26,13 @@ export class BaseConfigFactory {
...
                     "$Label": "readonly",        // ^
                     "$Locale": "readonly",       // ^
-                    "$Resource": "readonly"      // ^
+                    "$Resource": "readonly",     // ^
+
+                    // ESLint doesn't natively know about various browser and node globals. So we add them here to
+                    // remove false positives for our users.
+                    ... globals.node,
+                    ... globals.browser,
+                    ... globals.es2017
```
stephen-carter-at-sf, line 35: Fix as you go item in response to seeing noise when scanning a ton of repos. This should make users happy :-)

packages/code-analyzer-eslint-engine/test/engine.test.ts
```diff
@@ -651,6 +651,41 @@ describe('Typical tests for the runRules method of ESLintEngine', () => {
...
+                endLine: 4,
+                endColumn: 31
+            }],
+            primaryLocationIndex: 0
+        };
+
+        expect(logEvents).toContainEqual({
+            type: EventType.LogEvent,
+            logLevel: LogLevel.Debug,
+            message: getMessage('ViolationFoundFromUnregisteredRule', 'oops-rule', JSON.stringify(filteredViolation, null, 2))
```
jfeingold35, line 686: What is our rationale for logging these violations instead of just including them with the results? I can't remember, and now that I'm seeing it again it looks weird to me.

jfeingold35 (approved):

packages/code-analyzer-eslint-engine/test/engine.test.ts
stephen-carter-at-sf: Good question. We do the same thing with eslint 8. Basically the rule name associated with the violation that eslint produces isn't one of the rules selected.

Basically ESLint 8 and 9 both create a proper violation that ends up in our official engine results whenever it wants to complain that a directive is used for an unknown rule. But Core complains when a violation associated with a rule that wasn't selected show up. This makes the engine fail completely since it is a post engine validation check - which we don't want.


Ideally ESLint 8 and 9 should just emit a warning. It's odd that it adds this as a violation. This is even more odd given that the ruleFilter doesn't include this rule. So ESLint is kind of violating their own contract in reporting a violation for a rule that wasn't in the ruleFilter gate.

jfeingold35: Okay, yeah, the core requirement that all violations be for rules that actually exist does totally justify the current behavior.

PR #312, opened 2025-06-20
FIX(pmd): @W-18808343@: Prevent AppExchange rules from running in an infinite loop

### LINKED ISSUE #1811: [BUG][code-analyzer] The AppExchange:AvoidChangeProtectionUnprotected freezes execution
### Have you tried to resolve this issue yourself first?

Yes

### Bug Description

It appears that the AvoidChangeProtectionUnprotected rule can hang the entire pmd engine. See the repro steps below.

It get's stuck at:

Image

### Output / Logs

_No response_

### Steps To Reproduce

Simple repro steps:
```
git clone https://github.com/certinia/ffhttp-googledrive.git
cd ffhttp-googledrive
sf code-analyzer run -r pmd:AppExchange:AvoidChangeProtectionUnprotected
```
Then things just hang forever. 

### Expected Behavior

PMD rules should always finish (and should be as fast as possible).

### Operating System

MacOs 15.4.1

### Salesforce CLI Version

@salesforce/cli/2.88.6

### Code Analyzer Plugin (code-analyzer) Version

code-analyzer 5.0.0

### Node Version

v23.6.1

### Java Version

openjdk 21.0.5.0.101 2025-01-21 LTS

### Python Version

Python 3.11.8

### Additional Context (Screenshots, Files, etc)

_No response_

### Workaround

_No response_

### Urgency

Low

stephen-carter-at-sf:
fixes https://github.com/forcedotcom/code-analyzer/issues/1811

### DIFF
--- a/packages/code-analyzer-pmd-engine/pmd-rules/src/main/java/com/salesforce/security/pmd/apex/DangerousChangeProtectionMethods.java
+++ b/packages/code-analyzer-pmd-engine/pmd-rules/src/main/java/com/salesforce/security/pmd/apex/DangerousChangeProtectionMethods.java
@@ -25,17 +25,11 @@ public class DangerousChangeProtectionMethods extends AbstractApexRule {
 
     @Override
     public Object visit(ASTMethodCallExpression node, Object data) {
-        if (Helper.isTestMethodOrClass(node)) {
-            super.visit(node, data);
-        }
-
-        if (node.getFullMethodName().compareToIgnoreCase(
-            CHANGE_PROTECTION) == 0
-            || node.getFullMethodName().compareToIgnoreCase(SYSTEM_CHANGE_PROTECTION) == 0) {
+        if (!Helper.isTestMethodOrClass(node) && (node.getFullMethodName().compareToIgnoreCase(CHANGE_PROTECTION) == 0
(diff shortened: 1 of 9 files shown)


packages/code-analyzer-pmd-engine/pmd-rules/src/main/java/com/salesforce/security/pmd/apex/DangerousChangeProtectionMethods.java
```diff
@@ -25,17 +25,11 @@ public class DangerousChangeProtectionMethods extends AbstractApexRule {
 
     @Override
     public Object visit(ASTMethodCallExpression node, Object data) {
-        if (Helper.isTestMethodOrClass(node)) {
-            super.visit(node, data);
```
stephen-carter-at-sf, line 29: The issue was because we super.visit was called twice (once here and once below). Ravindra should have returned the super.visit output here instead of continuing.

It caused an issue with this one file because this one file has a name that is referenced within itself. Since super was called twice it wasn't able to use its recursion detection I believe.

jfeingold35 (approved):

jfeingold35 (approved):

PR #315, opened 2025-07-14
CHANGE: @W-19015089@: Update dependencies as much as possible

stephen-carter-at-sf:


packages/ENGINE-TEMPLATE/package.json
```diff
@@ -14,17 +14,17 @@
...
-    "@salesforce/code-analyzer-engine-api": "0.25.0"
+    "@salesforce/code-analyzer-engine-api": "0.26.0-SNAPSHOT"
   },
   "devDependencies": {
-    "@eslint/js": "^9.28.0",
-    "@types/jest": "^29.5.14",
-    "eslint": "^9.28.0",
-    "jest": "^29.7.0",
+    "@eslint/js": "^9.31.0",
+    "@types/jest": "^30.0.0",
```
jfeingold35, line 21: This is a dumb question, but why isn't `@types/jest` in the top-level package.json?

jfeingold35 (approved):

packages/ENGINE-TEMPLATE/package.json
stephen-carter-at-sf: The top level doesn't have any tests where we need the types defined. The top level package only needs jest so that we can call the jest script for the whole monorepo instead of calling jest individually on each package at a time.

PR #319, opened 2025-08-01
FIX: @W-19184117@: Our JSON schema fields primaryLocationIndex, locat…

stephen-carter-at-sf:
…ions, resources should not be marked as optional.

Rationale:

We prefer to allow consumers of our json output to not have to check if the field exists, but instead to just check the size.
Since this does not currently reflect the truth that comes from the core data structure ( see https://github.com/forcedotcom/code-analyzer-core/blob/dev/packages/code-analyzer-core/src/results.ts#L43-L53) and the translation to JSON (see https://github.com/forcedotcom/code-analyzer-core/blob/dev/packages/code-analyzer-core/src/output-formats/results/json-run-results-format.ts#L136-L139)... we should fix this.

### DIFF
--- a/packages/code-analyzer-core/src/output-formats/results/json-run-results-format.ts
+++ b/packages/code-analyzer-core/src/output-formats/results/json-run-results-format.ts
@@ -56,7 +56,7 @@ export type JsonViolationOutput = {
     // The index of the primary code location within the code locations array
     primaryLocationIndex: number
 
-    // An array of code locations associated with the violation
+    // An non-empty array of code locations associated with the violation
     locations: JsonCodeLocationOutput[]
 
     // The violation message
--- a/packages/code-analyzer-core/src/output-formats/results/json-run-results-format.ts
+++ b/packages/code-analyzer-core/src/output-formats/results/json-run-results-format.ts
@@ -54,16 +54,16 @@ export type JsonViolationOutput = {
     tags: string[]
 
     // The index of the primary code location within the code locations array
-    primaryLocationIndex?: number
+    primaryLocationIndex: number
 
     // An array of code locations associated with the violation
-    locations?: JsonCodeLocationOutput[]
+    locations: JsonCodeLocationOutput[]
 
     // The violation message
     message: string
 
     // An array of urls for resources associated with the violation
-    resources?: string[]
+    resources: string[]
 }
 export type JsonCodeLocationOutput = {
     // The path, relative to runDir, of the file associated with the violation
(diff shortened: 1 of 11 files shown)


packages/code-analyzer-core/src/output-formats/results/json-run-results-format.ts
```diff
@@ -54,16 +54,16 @@ export type JsonViolationOutput = {
     tags: string[]
 
     // The index of the primary code location within the code locations array
-    primaryLocationIndex?: number
+    primaryLocationIndex: number
 
     // An array of code locations associated with the violation
-    locations?: JsonCodeLocationOutput[]
+    locations: JsonCodeLocationOutput[]
```
jfeingold35, line 60: Correct me if I'm wrong, but it's possible for this list to be empty, right? So in that case, would you still want to have a primaryLocationindex at all? And  if so, what would the value be?

stephen-carter-at-sf: No it should not possible for this to ever be empty right now... but look over at https://github.com/forcedotcom/code-analyzer-core/blob/dev/packages/code-analyzer-core/src/code-analyzer.ts#L651-L661 I see that although we validation allows for this. Let me fix that as well.

I think the original thinking was that we would allow for a violation to not have code locations... but when we implemented the HTML report we decided its best if consumers should not have to check the size of the code locations and instead always know that they can get a Code Location object with v.codeLocations[v.primaryLocationIndex] always... and then you check if "file" exists in the code location or not to see if it is a location without a file... which is why we introduced the "UndefinedCodeLocation" object:
https://github.com/forcedotcom/code-analyzer-core/blob/dev/packages/code-analyzer-core/src/results.ts#L151

The alternative to this design would be that violations don't need code locations - but then the question is... what would the primaryLocationIndex be? Would it be -1 in that case? So choose your poison. Should we be checking for a -1 primaryCodeLocation index every single time.. or doing the null object pattern with a an AnbstractLocationlessViolation pattern which returns a UndfiedCodeLocation.Instace - see https://github.com/forcedotcom/code-analyzer-core/blob/dev/packages/code-analyzer-core/src/results.ts#L234

stephen-carter-at-sf: OK Updated PR.

packages/code-analyzer-core/src/output-formats/results/xml-run-results-format.ts
```diff
@@ -38,42 +38,39 @@ export class XmlRunResultsFormatter implements RunResultsFormatter {
             for (const tag of violationOutput.tags) {
                 tagsNode.node('tag').text(tag);
             }
-            if (violationOutput.primaryLocationIndex != null) {
```
stephen-carter-at-sf, line 41: Removed unnecessary if statements

jfeingold35 (approved):

PR #320, opened 2025-08-01
NEW @W-19212468@ Changed text when there are no violations

jfeingold35:


randi274 (commented):
Looking good! I did find some additional occurrences of a similar pattern, but not seeing test failures - are these dead code paths or should they be updated as well? 
Screenshot 2025-08-04 at 9 20 51 AM

jfeingold35:
@randi274 , I believe those cases are legitimate, since those are the cases for when a non-zero number of violations is found. It makes sense to say "Found 3 violations across 2 files." It doesn't make sense to say "Found 0 violations across 0 files".

randi274 (approved):

PR #326, opened 2025-08-14
NEW @W-17977608@ CSV output format for rules

jfeingold35:


stephen-carter-at-sf (changes requested):
I think you are missing the command help text updates.

Also why is graph engine creating files still locally? Let’s not check these in - we should make it so these don’t get created locally

packages/code-analyzer-core/test/stubs.ts
```diff
@@ -474,6 +474,41 @@ class EmptyTagEngine extends engApi.Engine {
     }
 }
 
+export class NewlineDescriptionEnginePlugin extends engApi.EnginePluginV1 {
```
stephen-carter-at-sf: Instead of creating yet another plugin and engine, consider just modifying the description for stub1RuleA or something with a newline and I would also suggest we add in various quotes as well.

Notice how we did this with the violation on line 365 above.

This should prevent us from having to have another file inside of exectedOutputFiles ... i.e. can we remove ruleSelectionWithNewlines.goldfile.csv in favor of just updating multipleRules.goldfile.csv

stephen-carter-at-sf (changes requested):

packages/code-analyzer-core/test/stubs.ts
jfeingold35: Yeah, modifying the rule description is probably a good idea. I'll look into that.

jfeingold35:
> I think you are missing the command help text updates.
> 
> Also why is graph engine creating files still locally? Let’s not check these in - we should make it so these don’t get created locally

The command help text is on the CLI side, right?
Also, yeah, good catch on the SFGE files. Not sure how those ended up there. I think my local config might have done something. I'll remove them.

packages/code-analyzer-core/test/output-format.test.ts
```diff
@@ -197,6 +198,38 @@ describe("RuleSelectionFormatter Tests", () => {
...
+            });
+            const rule2: RuleImpl = new RuleImpl('stubEngine1', {
+                name: 'stub1RuleB',
+                severityLevel: SeverityLevel.Low,
+                tags: ['Recommended', 'Performance'],
+                description: 'A simple description this time',
+                resourceUrls: []
+            });
+            complicatedRuleSelection.addRule(rule1);
+            complicatedRuleSelection.addRule(rule2);
```
jfeingold35, line 226: Rather than modify the actual hardcoded description of any of the stub rules (which would have likely impacted other tests), I decided it was just easier to instantiate a rule selection from scratch with some made up rules.
It's my opinion that this is acceptable, but I'm prepared for disagreement.

packages/code-analyzer-core/test/stubs.ts
stephen-carter-at-sf: Thanks. is this used anymore? If not, then can we remove it and its corresponding engine?

stephen-carter-at-sf (approved):

PR #328, opened 2025-08-14
NEW: @W-18395429@ Add SLDS Config Settings and Rules to ESLint (v9)

randi274:
This PR:
- Adds new SLDS rules for Lightning Web Components HTML and Salesforce CMP files
- Adds a new setting for the ESLint engine `disable_slds_base_config` to turn off these new SLDS rules
- Adds a new file extension for the ESLint engine `html` (.html, .htm, and .cmp) to be used for rule discovery of particular files
- Note: settings are available with v9's flat config only


packages/code-analyzer-eslint-engine/test/test-data/workspaceWithFlatConfigJs/a-config-file-that-uses-ignores.js
```diff
@@ -5,19 +5,5 @@ module.exports = defineConfig([
         // Should globally ignore js files - making the base js and lwc config not applicable
         // See https://eslint.org/docs/latest/use/configure/configuration-files#globally-ignoring-files-with-ignores
         ignores: ["**/*.js"]
-    },
```
randi274, line 8: This created some big testing issues. I'd love to figure out why, but for now I wanted 🟢  tests. Additionally, removing this didn't fail any tests, so it was clearly not needed explicitly. If I put it back, I'd like to make sure we're testing against this, since the config title just calls out ignoring JS.

Screenshot 2025-08-14 at 4 26 16 PM

packages/code-analyzer-eslint-engine/test/engine.test.ts
```diff
@@ -734,14 +774,14 @@ describe('Tests for emitting events', () => {
...
-            [0, 10, 14, 18, 33, 48, 63, 78, 82, 86, 90, 95, 100]);
+            [0, 10, 14, 18, 28, 38, 48, 58, 68, 78, 82, 86, 90, 95, 100]);
     });
 
     it('When runRules is called, then it emits correct progress events', async () => {
         const runOptions: RunOptions = createRunOptions(new Workspace('id', [workspaceWithNoCustomConfig]));
         await engine.runRules(['no-unused-vars'], runOptions);
         expect(runRulesProgressEvents.map(e => e.percentComplete)).toEqual(
-            [0, 1.5, 3, 14.25, 25.5, 27, 28.5, 30, 62.5, 95, 100]);
+            [0, 1.5, 3, 10.5, 18, 25.5, 27, 28.5, 30, 51.67, 73.33, 95, 100]);
```
randi274, line 807: I did find it interesting that these changed, but there's more work being done now with more rules.

packages/code-analyzer-eslint-engine/src/rule-mappings.ts
```diff
@@ -1595,5 +1595,21 @@ export const RULE_MAPPINGS: Record<string, {severity: SeverityLevel, tags: strin
...
         severity: SeverityLevel.Moderate,
         tags: [/* NOT RECOMMENDED */    COMMON_TAGS.CATEGORIES.BEST_PRACTICES, COMMON_TAGS.LANGUAGES.TYPESCRIPT]
+    },
+    
+    // =================================================================================================================
+    //   SLDS/HTML BASE RULES - @salesforce-ux/eslint-plugin-slds
+    // =================================================================================================================
+    "@salesforce-ux/slds/enforce-bem-usage": {
+        severity: SeverityLevel.High,
+        tags: [COMMON_TAGS.RECOMMENDED, COMMON_TAGS.LANGUAGES.HTML] //TODO: are there other tags?
```
randi274: Might be something we can answer, or have to ask the team on.

packages/code-analyzer-eslint-engine/src/missing-types-for-dependencies.d.ts
```diff
@@ -35,3 +35,18 @@ declare module '@salesforce/eslint-config-lwc' {
     };
     export = moduleObject;
 }
+
+// This declaration adds in the missing types for "@salesforce-ux/eslint-plugin-slds" whose package.json file's main field points to:
+//     node_modules/@salesforce-ux/eslint-plugin-slds/build/index.js
```
randi274, line 40: Hoping I got this part right! I opted not to type the v8 exports since it's not used.

packages/code-analyzer-eslint-engine/src/rule-mappings.ts
stephen-carter-at-sf: Every rule should have a category tag... so yes we need to ask what each of these rules should fall under regarding our category tags.

packages/code-analyzer-eslint-engine/src/base-config.ts
```diff
@@ -105,6 +109,21 @@ export class BaseConfigFactory {
...
+        //todo: Should this need to be updated now to take CSS into consideration?
+        return sldsEslintPlugin.configs['flat/recommended'].map(conf => ({
+            ...conf,
+            files: this.engineConfig.file_extensions.html.map(ext => `**/*${ext}`),
+            languageOptions: {
+                    ... (conf.languageOptions ?? {}),
+                    parserOptions: {
+                        ... (conf.languageOptions?.parserOptions ?? {}),
+                        projectService: true
+                    }
```
stephen-carter-at-sf: What is this for?

Typescript needs the project service so that it'll auto discover the tsconfig.json files for each typescript file that eslint lints.

But I don't think there is anything like that needed for the html parser is there?

packages/code-analyzer-eslint-engine/test/engine.test.ts
stephen-carter-at-sf: I believe it is the number of files actually. Between 0 and 30 it is calculating the config for each file (with some of the points fixed)... and between 30 and 95 it is linting each file. So because you added in workspace_NoCustomConfig/dummy1.html which is another file that is now valid to process, I would have expected exactly what we see here.

packages/code-analyzer-eslint-engine/test/test-data/workspaceWithFlatConfigJs/a-config-file-that-uses-ignores.js
stephen-carter-at-sf: What test was causing this failure?

stephen-carter-at-sf: Ah yes I know what's going on here now. While it is calculating the config for dummy.html it is reconsoling the config to only what is relevant. So the other config options which have a files filter on it for typescript go away here this config for the rules remains. 3 fix options could have happened: 1 add in the `        files: ["**/*.ts"],` to that config object, 2 add in the plugin to that config option, or 3 just remove that config option. You are right to point out that it isn't being used from a test perspective (even though behind the scenes that one rule is indeed modified for typescript file cases. So I think to keep things simple, yeah your removal of this config option makes sense.

But do me a favor and remove the unnecessary use of defineConfig here... and just make the a-config-file-that-uses-ignores.ts file just look like this:
```
module.exports = [
    {
        // Should globally ignore js files - making the base js and lwc config not applicable
        // See https://eslint.org/docs/latest/use/configure/configuration-files#globally-ignoring-files-with-ignores
        ignores: ["**/*.js"]
    }
];
```

packages/code-analyzer-eslint-engine/src/base-config.ts
randi274: Let me try and back this out - I was having some weird test failures, so was pulling things from one spot to another to try and get them to work.

randi274: Ok yes! This is great to clean up 👍

packages/code-analyzer-eslint-engine/test/test-data/workspaceWithFlatConfigJs/a-config-file-that-uses-ignores.js
randi274: Ah that makes sense - thanks for clarifying! I'll make that change as well. Good cleanup opportunity!

packages/code-analyzer-eslint-engine/test/test-data/workspaceWithFlatConfigMjs/eslint.config.mjs
```diff
@@ -1,12 +1,6 @@
...
-            "no-useless-escape": [0]
-        },
-    },
-]);
+export const rules = {
+    "array-callback-return": ["error"],
+    "no-unused-vars": ["error", { argsIgnorePattern: "^_" }],
+    "no-useless-backreference": ["warn"],
+    "no-useless-escape": [0]
+};
```
stephen-carter-at-sf: This is wrong.
You are not exporting a default export and you are exporting a rules object (which aren't being used).

What an eslint config should be returning is an array of config objects.
So this should be
export default [
  {
     "rules": ...
  }
]

stephen-carter-at-sf (changes requested):

packages/code-analyzer-eslint-engine/src/config.ts
```diff
@@ -56,10 +62,12 @@ export const DEFAULT_CONFIG: ESLintEngineConfig = {
     auto_discover_eslint_config: false,
     disable_javascript_base_config: false,
     disable_lwc_base_config: false,
+    disable_slds_base_config: false,
     disable_typescript_base_config: false,
     file_extensions: {
         javascript: ['.js', '.cjs', '.mjs'],
         typescript: ['.ts'],
+        html: ['.html', '.cmp'],
```
stephen-carter-at-sf: You are missing '.htm'

stephen-carter-at-sf (changes requested):

packages/code-analyzer-eslint-engine/src/missing-types-for-dependencies.d.ts
stephen-carter-at-sf: Yes this looks correct.

packages/code-analyzer-eslint-engine/src/rule-mappings.ts
```diff
@@ -1595,5 +1595,21 @@ export const RULE_MAPPINGS: Record<string, {severity: SeverityLevel, tags: strin
     "@typescript-eslint/use-unknown-in-catch-callback-variable": {
         severity: SeverityLevel.Moderate,
         tags: [/* NOT RECOMMENDED */    COMMON_TAGS.CATEGORIES.BEST_PRACTICES, COMMON_TAGS.LANGUAGES.TYPESCRIPT]
+    },
+    
+    // =================================================================================================================
+    //   SLDS/HTML BASE RULES - @salesforce-ux/eslint-plugin-slds
+    // =================================================================================================================
+    "@salesforce-ux/slds/enforce-bem-usage": {
+        severity: SeverityLevel.High,
```
stephen-carter-at-sf: Just waiting on the other team to give the correct tags and severities.

stephen-carter-at-sf (changes requested):

packages/code-analyzer-eslint-engine/src/config.ts
randi274: Added, thanks!

packages/code-analyzer-eslint-engine/test/test-data/workspaceWithFlatConfigMjs/eslint.config.mjs
```diff
@@ -1,12 +1,11 @@
-import { defineConfig } from "eslint/config";
-
-export default defineConfig([
+export default [
     {
         rules: {
+            "array-callback-return": ["error"],
```
jfeingold35, line 4: Is this duplicate line intentional? If so, please add a comment explaining its purpose, because it's not at all clear at this point in the code review.

packages/code-analyzer-eslint-engine/src/rule-mappings.ts
```diff
@@ -1595,5 +1595,21 @@ export const RULE_MAPPINGS: Record<string, {severity: SeverityLevel, tags: strin
...
         severity: SeverityLevel.Moderate,
         tags: [/* NOT RECOMMENDED */    COMMON_TAGS.CATEGORIES.BEST_PRACTICES, COMMON_TAGS.LANGUAGES.TYPESCRIPT]
+    },
+    
+    // =================================================================================================================
+    //   SLDS/HTML BASE RULES - @salesforce-ux/eslint-plugin-slds
+    // =================================================================================================================
+    "@salesforce-ux/slds/enforce-bem-usage": {
+        severity: SeverityLevel.Low,
+        tags: [COMMON_TAGS.RECOMMENDED, COMMON_TAGS.CATEGORIES.BEST_PRACTICES, COMMON_TAGS.LANGUAGES.HTML] //TODO: are there other tags?
```
jfeingold35: Was this TODO meant to be dealt with before merging?

stephen-carter-at-sf (approved):

PR #339, opened 2025-08-27
NEW @W-19365711@ Converted package versioning enforcement to git hooks

jfeingold35:


.github/workflows/verify-pr.yml
```diff
@@ -21,48 +21,6 @@ jobs:
...
-        id: main
-        if: github.base_ref == 'dev'
-        run: |
-          title="${{ github.event.pull_request.title }}"
-          if [[ "$title" =~ ^POSTRELEASE ]]; then
-            echo "is_postrelease=true" >> "$GITHUB_OUTPUT"
-          else
-            echo "is_postrelease=false" >> "$GITHUB_OUTPUT"
-          fi
-  validate_packages:
```
jfeingold35: This job will need to be removed from the branch protection rules.

jfeingold35 (commented):
I didn't need to change anything in the publishing flow, because the commit in the publishing flow is done via graphQL, which I believe should sidestep Husky entirely.

.github/workflows/verify-pr.yml
stephen-carter-at-sf: I assumed we would have both: github hooks and branch protection rules that still run here.
We can't just rely on husky commit hooks because they can easily be bypassed. We should definitely still have the branch protection here - just simply call to the script(s).

jfeingold35: The language we were using was to have these checks be done at commit time so that the GHA would have to do less work, so I got mislead.

package.json
```diff
@@ -17,7 +17,8 @@
     "scrub": "npm run clean && npm run scrub --workspaces --if-present && rimraf node_modules && rimraf package-lock.json",
     "showcoverage-java": "npm run showcoverage-java --workspaces --if-present",
     "showcoverage-typescript": "open ./coverage/lcov-report/index.html",
-    "showcoverage": "npm run showcoverage-java && npm run showcoverage-typescript"
+    "showcoverage": "npm run showcoverage-java && npm run showcoverage-typescript",
+    "prepare": "husky"
```
stephen-carter-at-sf, line 21: Was
npm install husky --save-dev
needed?
I don’t see the package.json updated at the top level here with the dev dependency.

stephen-carter-at-sf (changes requested):

.github/workflows/verify-pr.yml
stephen-carter-at-sf: No - it's so that we don't commit and then afterward find ci/cd fail and then have to commit again. We should have both.

.husky/node-scripts/validate-changed-package-versions.js
```diff
@@ -74,18 +75,31 @@ function isFileInTestFolder(changedFile) {
...
-        //A temporary workaround for the rename of the flowtest-engine package to flow-engine
-        if (changedPackage === 'packages/code-analyzer-flowtest-engine') {
-            continue;
-        }
         const packageVersion = getPackageVersion(changedPackage);
         if (!packageVersion.endsWith('-SNAPSHOT')) {
             incorrectlyVersionedPackages.push(`${changedPackage} (currently versioned as ${packageVersion}) lacks a trailing "-SNAPSHOT"`);
+            continue;
+        }
+        const releasedPackageVersions = getReleasedPackageVersions(changedPackage);
```
stephen-carter-at-sf: For each version that had a -SNAPSHOT added to it, I was looking for a NEW_VERSION_MINUS_SNAPSHOT >= CURRENT_VERSION check. I wouldn't use includes here... it isn't robust because you are checking against all the releasedPackageVersions here aren't you?

stephen-carter-at-sf (approved):

.github/workflows/verify-pr.yml
```diff
@@ -49,20 +49,21 @@ jobs:
...
+      - run: npm install
       - name: Validate that changed packages are versioned as snapshots
         if: ${{ needs.check_for_postrelease_keyword.outputs.is-postrelease == 'false' }}
         run: |
           BASE_SHA=${{ github.event.pull_request.base.sha }}
           HEAD_SHA=${{ github.event.pull_request.head.sha }}
-          git diff --name-only $HEAD_SHA $BASE_SHA > changed_files.txt
-          node ./.github/workflows/verify-pr/validate-changed-package-versions.js changed_files.txt
+          git diff --name-only $HEAD_SHA $BASE_SHA > `pwd`/changed_files.txt
+          node ./.node-scripts/validate-changed-package-versions.js `pwd`/changed_files.txt
```
stephen-carter-at-sf, line 59: This is causing a commit issue: 
image

We shouldn't have a changed_files.txt created locally here. Worse case, we can add it to .gitignore... best case - let's move this to a temp folder so we don't see it (or even better just pipe the diff directly into the javascript instead of writing first to a file.

stephen-carter-at-sf (changes requested):

.github/workflows/verify-pr.yml
stephen-carter-at-sf, line 59: In addition to the husky precommit stuff... do you want to also update these to use the .scatchfile as well?

.node-scripts/validate-changed-package-versions.js
```diff
@@ -74,18 +76,31 @@ function isFileInTestFolder(changedFile) {
...
-            continue;
-        }
         const packageVersion = getPackageVersion(changedPackage);
         if (!packageVersion.endsWith('-SNAPSHOT')) {
             incorrectlyVersionedPackages.push(`${changedPackage} (currently versioned as ${packageVersion}) lacks a trailing "-SNAPSHOT"`);
+            continue;
+        }
+        const releasedPackageVersion = getLatestReleasedVersion(changedPackage);
+        if (semver.lte(semver.parse(packageVersion.slice(0, packageVersion.length - 9)), semver.parse(releasedPackageVersion))) {
+            incorrectlyVersionedPackages.push(`${changedPackage} (currently versioned as ${packageVersion}) is semantically behind latest published release ${releasedPackageVersion}`);
```
stephen-carter-at-sf: I think this should say `is semantically not ahead of latest`... because it could be equal.

stephen-carter-at-sf (changes requested):

stephen-carter-at-sf (approved):
Please fix my last 2 comments. Then you are good to merge.

.github/workflows/verify-pr.yml
jfeingold35: We could, but it doesn't matter terribly because there's no chance of the file being committed or anything like that.

PR #342, opened 2025-09-08
CHANGE: @W-19582414@ Export ViolationCounts

randi274:
This PR:
- Exports ViolationCounts for external use

### DIFF
--- a/packages/code-analyzer-core/src/index.ts
+++ b/packages/code-analyzer-core/src/index.ts
@@ -47,10 +47,6 @@ export {
 } from "./output-format"
 
 export type {
-    ViolationCounts
-} from "./output-formats/results/json-run-results-format"
-
-export type {
     CodeLocation,
     EngineRunResults,
     RunResults,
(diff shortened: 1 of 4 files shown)


packages/code-analyzer-core/src/index.ts
```diff
@@ -46,6 +46,10 @@ export {
     RunResultsFormatter
 } from "./output-format"
 
+export type {
+    ViolationCounts
+} from "./output-formats/results/json-run-results-format"
```
randi274: I first hoisted this into output-format.ts, but that felt strange since I wasn't using it in there, but I could do that to make this look cleaner if it's preferable

stephen-carter-at-sf: Instead of exporting this from the top level... I'd prefer if we instead switched to use "exports" instead of "main" sort of like we do with our engine-api module. See: https://github.com/forcedotcom/code-analyzer-core/blob/dev/packages/code-analyzer-engine-api/package.json#L13-L16

Then you can put an index.ts inside of the output-formats folder... so that the import over on the MCP tool provider side looks something like
```
import { ViolationCounts } from "@salesforce/code-analyzer-core/output-formats"
```

stephen-carter-at-sf (approved):

PR #346, opened 2025-09-09
CHANGE: @W-19397345@: Use provided workingFolder instead of creating new temp dirs

stephen-carter-at-sf:


packages/T-E-M-P-L-A-T-E/package.json
stephen-carter-at-sf, line 1: We have the ENGINE-TEMPLATE which is sufficient now. We don't need this other template package anymore.

packages/code-analyzer-stylelint-engine/package.json
stephen-carter-at-sf, line 1: Fix as you go item. We do not need this stylelint engine code anymore since we won't be using it. Removing it instead of maintaining this anymore.

.node-scripts/validate-changed-package-versions.js
```diff
@@ -49,7 +49,7 @@ function identifyMeaningfullyChangedPackages(changedFiles) {
 
     for (const changedFile of changedFiles) {
         const changedPackage = convertFileNameToPackageNameIfPossible(changedFile);
-        if (changedPackage && !isFileInTestFolder(changedFile)) {
+        if (changedPackage && !isFileInTestFolder(changedFile) && !isPackageThatHasNotPublished(changedPackage)) {
```
stephen-carter-at-sf: npm view is called on packages associated with any changed files... so to avoid that, it's best to not keep these packages in our changedPackage list. This allowed me to make a change to our template without this script dying.

packages/ENGINE-TEMPLATE/package.json
```diff
@@ -36,7 +36,7 @@
   ],
   "scripts": {
     "build": "tsc --build tsconfig.build.json --verbose",
-    "test": "jest --coverage",
+    "test": "tsc --build tsconfig.json --noEmit && jest --coverage",
```
stephen-carter-at-sf: Sorry for all the noise in the tests. I had to update a bunch of tests since it turns out that Josh's last change failed to make the tests compile. jest doesn't attempt to do any compile when it executes... so I'm adding in a --noEmit bit before running our tests so we don't miss this anymore.

packages/code-analyzer-engine-api/src/utils/fs-utils.ts
```diff
@@ -1,20 +1,6 @@
-import * as tmp from 'tmp';
```
randi274, line 1: Can tmp be removed as a dependency in this package now? I see we also have the types in the eslint engine and I'm not sure we need those either:
Screenshot 2025-09-10 at 11 46 54 AM

stephen-carter-at-sf: Nice catch. Fixed!

randi274 (approved):

PR #355, opened 2025-09-19
NEW: @W-19397375@: Preserve working folders on error

stephen-carter-at-sf:


packages/code-analyzer-core/src/code-analyzer.ts
```diff
@@ -110,9 +112,15 @@ export class CodeAnalyzer {
     private readonly rulesCache: Map<string, RuleImpl[]> = new Map();
     private readonly engineRuleDiscoveryProgressAggregator: EngineProgressAggregator = new EngineProgressAggregator();
 
-    constructor(config: CodeAnalyzerConfig, version: string = process.version) {
-        this.validateEnvironment(version);
+    constructor(config: CodeAnalyzerConfig, fileSystem: FileSystem = new RealFileSystem(), nodeVersion: string = process.version) {
+        this.validateEnvironment(nodeVersion);
         this.config = config;
+        this.tempFolder = new TempFolder(fileSystem);
+        /* istanbul ignore next */
```
randi274, line 119: Could you clarify why this needed the ignore?

packages/code-analyzer-core/src/code-analyzer.ts
```diff
@@ -344,32 +365,62 @@ export class CodeAnalyzer {
...
+            const rulesWorkingFolderName: string = `rules-${this.clock.formatToDateTimeString()}`;
+
+            await this.tempFolder.makeSubfolder(rulesWorkingFolderName);
+
+            const rulePromises: Promise<RuleImpl[]>[] = this.getEngineNames().map(async (engineName) => {
+                const workingFolder: string = await this.tempFolder.makeSubfolder(rulesWorkingFolderName, engineName);
+                const describeOptions: engApi.DescribeOptions = {
                     workspace: engApiWorkspace,
-                    workingFolder: await this.tempFolder.createSubfolder(describeWorkingFolderName, engineName),
+                    workingFolder: workingFolder,
```
randi274, line 379:
```suggestion
                    workingFolder,
```

randi274: Nice cleanup 😎

packages/code-analyzer-core/src/utils.ts
```diff
@@ -34,25 +30,82 @@ export class RuntimeUniqueIdGenerator implements UniqueIdGenerator {
     }
 }
 
-export interface TempFolder {
-    getPath(): Promise<string>;
-    createSubfolder(...subfolderPathSegs: string[]): Promise<string>;
+export interface FileSystem {
```
randi274, line 33: This is good - I feel like it could use elevating out of `utils.ts`

packages/code-analyzer-core/src/code-analyzer.ts
stephen-carter-at-sf: The callback is really what's ignored. The test could never measure this because the node process is on exit when the call back is fired. So no way to cover this with a unit test. Demo (end-to-end) shows things working though.

packages/code-analyzer-core/src/code-analyzer.ts
stephen-carter-at-sf: I actually find that this syntax is super confusing and it ties the variable name to the field name. So I personally choose not to use this type of syntax from javascript typically. Just a personal preference.

packages/code-analyzer-core/src/utils.ts
stephen-carter-at-sf: But it is an implementation detail utility in my opinion.

packages/code-analyzer-core/src/code-analyzer.ts
stephen-carter-at-sf: For example, if I use my editor to rename a variable - it causes issues if I have "workingFolder" as both the fieldname and the variable. I might not want to change the field name... but may want to change the variable name and editors don't know this typically and get it wrong. Sometimes convenience and short code causes bugs... which is why I love static typing as well. :-)

randi274: Works for me!

randi274 (approved):
This will be great to for us to add for some additional debugging support!

PR #359, opened 2025-09-23
NEW: @W-19053333@ Add CSS Config to ESLint Engine with New Rules

randi274:
This PR:
- Adds a CSS Configuration to the ESLint engine.
- Adds CSS Rules from the [salesforce-ux/eslint-plugin-slds](https://www.npmjs.com/package/@salesforce-ux/eslint-plugin-slds).

### DIFF
--- a/packages/code-analyzer-eslint-engine/src/base-config.ts
+++ b/packages/code-analyzer-eslint-engine/src/base-config.ts
@@ -9,11 +9,11 @@ import globals from "globals";
 
 export class BaseConfigFactory {
     private readonly engineConfig: ESLintEngineConfig;
-
+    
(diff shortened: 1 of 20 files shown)


packages/code-analyzer-eslint-engine/src/base-config.ts
jfeingold35, line 1: Seems like there's some extraneous whitespace changes in this file?

jfeingold35 (approved):

packages/code-analyzer-eslint-engine/src/base-config.ts
randi274: Agreed - I think it was trying to make the whitespace consistent! I'll merge this one in, but be mindful about other changes I see come later.

PR #365, opened 2025-10-03
FIX @W-19772057@ Fixed oversight in selector code

jfeingold35:
There was a slight oversight in my initial selector code. Previously, it would try to split `(2,3):pmd` into `(2` and `3):pmd` instead of `(2,3)` and `pmd`. This has been addressed.

### DIFF
--- a/packages/code-analyzer-core/src/selectors.ts
+++ b/packages/code-analyzer-core/src/selectors.ts
@@ -7,10 +7,13 @@ export interface Selector {
 export function toSelector(selectorString: string): Selector {
     // We parse the selector back-to-front, so that the front-most selectors end up at the bottom of the tree we create
     // and therefore get resolved first.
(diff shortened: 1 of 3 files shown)


packages/code-analyzer-core/src/selectors.ts
```diff
@@ -7,10 +7,13 @@ export interface Selector {
 export function toSelector(selectorString: string): Selector {
     // We parse the selector back-to-front, so that the front-most selectors end up at the bottom of the tree we create
     // and therefore get resolved first.
+    console.log(`selector is ${selectorString}`);
```
randi274: Can the console statements be removed?

jfeingold35: Good catch!

randi274 (approved):

PR #367, opened 2025-10-09
FIX @W-19772057@ Refactored selection algorithm

jfeingold35:
Previously, `eslint:(2,BestPractices:(3,4))` was being parsed as `eslint:((2,BestPractices):(3,4))` instead of `eslint:(2,(BestPractices:(3,4)))`. This has been corrected, and additional test coverage has been added.

### DIFF
--- a/packages/code-analyzer-core/src/selectors.ts
+++ b/packages/code-analyzer-core/src/selectors.ts
@@ -5,81 +5,63 @@ export interface Selector {
 }
 
 export function toSelector(selectorString: string): Selector {
-    // We parse the selector back-to-front, so that the front-most selectors end up at the bottom of the tree we create
-    // and therefore get resolved first.
-    if (selectorString === '') {
+    const trimmedSelector: string = selectorString.trim();
+
+    if (trimmedSelector === '') {
         // ERROR CASE: The selector is empty. Possible if you do something like "()" or "a:()".
         throw new Error(getMessage("SelectorCannotBeEmpty"));
-    } else if (selectorString.endsWith(')')) {
-        // If the selector ends in close-paren, then we need to find the open-paren that matches it.
-        const correspondingOpenParen: number = identifyCorrespondingOpenParen(selectorString);
-        if (correspondingOpenParen === 0) {
-            // RECURSIVE CASE: The entire selector is wrapped in parens. Pop them off and call recursively.
-            return toSelector(selectorString.slice(1, -1))
-        } else {
-            // RECURSIVE CASE: The open-paren is somewhere in the middle of the selector and accompanied by an operator.
-            const left: string = selectorString.slice(0, correspondingOpenParen - 1);
-            const right: string = selectorString.slice(correspondingOpenParen);
-            const op: string = selectorString[correspondingOpenParen - 1];
-            return toComplexSelector(left, right, op);
-        }
-    } else {
-        // If there's a close-paren in the string, only look for operators after it.
-        const lastCloseParen: number = Math.max(selectorString.lastIndexOf(')'), 0);
-        const lastComma: number = selectorString.slice(lastCloseParen).lastIndexOf(',');
-        const lastColon: number = selectorString.slice(lastCloseParen).lastIndexOf(':');
-
-        // BASE CASE: The selector contains no commas or colons.
-        if (lastComma === -1 && lastColon === -1) {
-            // Parens only make sense in conjunction with operators, so if we find any, the selector is malformed.
-            if (selectorString.includes(')') || selectorString.includes('(')) {
-                throw new Error(getMessage('SelectorLooksIncorrect', selectorString));
-            }
-            return new SimpleSelector(selectorString);
-        } else if (lastComma !== -1) {
-            // Commas resolve before colons, so that "x,a:b" and "a:b,x" both resolve equivalently the combination of
-            // "x" and "a:b".
-            const left: string = selectorString.slice(0, lastComma + lastCloseParen);
-            const right: string = selectorString.slice(lastComma + lastCloseParen + 1);
-            return toComplexSelector(left, right, ',');
-        } else {
-            const left: string = selectorString.slice(0, lastColon + lastCloseParen);
-            const right: string = selectorString.slice(lastColon + lastCloseParen + 1);
-            return toComplexSelector(left, right, ':');
-        }
     }
-}
 
-function identifyCorrespondingOpenParen(selectorString: string): number {
-    const reversedLetters: string[] = selectorString.split('').reverse();
+    let commaIdx: number|null = null;
+    let colonIdx: number|null = null;
     let parenBalance: number = 0;
-    let idx = 0;
-    for (const letter of reversedLetters) {
-        if (letter === ')') {
+    for (let i = 0; i < trimmedSelector.length; i++) {
+        const char: string = trimmedSelector[i];
+        if (char === '(') {
             parenBalance += 1;
-        } else if (letter === '(') {
+        } else if (char === ')') {
             parenBalance -= 1;
+            // If our parenthesis balance is negative, it means there are more close-parens than open-parens, which is a problem.
+            if (parenBalance < 0) {
+                throw new Error(getMessage("SelectorLooksIncorrect", selectorString));
+            }
+        } else if (char === ',') {
+            // If we're not inside of parentheses, and we haven't already found a comma, note the location of this one.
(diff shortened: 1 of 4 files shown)


packages/code-analyzer-core/src/selectors.ts
```diff
@@ -5,81 +5,63 @@ export interface Selector {
...
+        } else if (char === ',') {
+            // If we're not inside of parentheses, and we haven't already found a comma, note the location of this one.
+            if (parenBalance === 0 && commaIdx === null) {
+                commaIdx = i;
+            }
+        } else if (char === ':') {
+            // If we're not inside of parentheses, and we haven't already found a colon, note the location of this one.
+            if (parenBalance === 0 && colonIdx === null) {
+                colonIdx = i;
+            }
```
stephen-carter-at-sf, line 46: Instead of keeping track of the commaIdx, since commas trump colons, can't we just do a quick return here to avoid processing more. I only think we need to keep track of the colonIdx the whole way through.

That is, can't we just do:
```
} else if (parenBalance > 0) { // Ignore chars inside of parens
    continue;
} else if (char === ',') {
        const left: string = trimmedSelector.slice(0, i);
        const right: string = trimmedSelector.slice(i + 1);
        return new OrSelector(toSelector(left), toSelector(right));
} else if (char === ':' && colonIdx === null) {
    colonIdx = i;
}
```
Then you don't need lines 47-50 below.

Also maybe rename colonIdx to firstColonIdx

packages/code-analyzer-core/src/selectors.ts
```diff
@@ -5,81 +5,63 @@ export interface Selector {
...
+        // If there are colons but no commas, split along the first colon.
+        const left: string = trimmedSelector.slice(0, colonIdx);
+        const right: string = trimmedSelector.slice(colonIdx + 1);
         return new AndSelector(toSelector(left), toSelector(right));
+    } else if (trimmedSelector[0] === '(' && trimmedSelector[trimmedSelector.length - 1] === ')') {
+        // If the first and last character are parentheses, then pop those off and run again.
+        return toSelector(trimmedSelector.slice(1, trimmedSelector.length - 1));
+    } else if (trimmedSelector.includes('(') || trimmedSelector.includes(')')) {
+        // There shouldn't be parentheses in the middle of a selector that has no operators.
+        throw new Error(getMessage('SelectorLooksIncorrect', selectorString));
```
stephen-carter-at-sf: Running through the string 2 more times is unneeded. Instead add a boolean that is called "hasParens" that is true anytime a "(" is found up on line 21. The ")" case shouldn't be needed because the negative check should already handle this. Then here you can just check if hasParents is true or not without doing more work.

packages/code-analyzer-core/src/selectors.ts
```diff
@@ -5,81 +5,63 @@ export interface Selector {
 }
 
 export function toSelector(selectorString: string): Selector {
-    // We parse the selector back-to-front, so that the front-most selectors end up at the bottom of the tree we create
-    // and therefore get resolved first.
-    if (selectorString === '') {
+    const trimmedSelector: string = selectorString.trim();
+
+    if (trimmedSelector === '') {
```
stephen-carter-at-sf: Instead of trimming, we should assume that coming spaces in Core are not allowed (they are already handled ont he client side), and so we should validate that there are no spaces and that the selector doesn't start or end with a special character like `,` or `:`. We don't want users giving a selector like `,something` or `something:`

stephen-carter-at-sf (changes requested):

packages/code-analyzer-core/test/rule-selection.test.ts
```diff
@@ -266,9 +281,31 @@ describe('Tests for selecting rules', () => {
...
+            case: 'leading colon',
+            selector: ':asdfasdf'
+        },{
+            case: 'trailing comma',
+            selector: 'asdfasdf,'
+        },{
+            case: 'trailing colon',
+            selector: 'asdfasdf:'
+        }
+    ])('Empty selectors are rejected. Case: $case', async ({selector}) => {
```
stephen-carter-at-sf: '' should be rejected, but it probably should have a different validation message than ' ', or ',something'. So after validating not empty, validate no spaces, and validate does not start or end with ':' or ',' maybe separately.

packages/code-analyzer-core/src/selectors.ts
jfeingold35: I thought about that, but doing the full iteration allows us to catch situations where the number of open-parens exceeds the number of close-parens, since the balance will be positive at the end.

packages/code-analyzer-core/src/selectors.ts
jfeingold35: Yeah, that's reasonable, I think.

packages/code-analyzer-core/src/selectors.ts
jfeingold35: Cool cool, I'll change the test into an error check instead.

packages/code-analyzer-core/test/rule-selection.test.ts
jfeingold35: Fair. I was trying to keep the `if-else` simple, but the separate error messages is worth having, so yeah.

packages/code-analyzer-core/src/selectors.ts
```diff
@@ -5,81 +5,70 @@ export interface Selector {
 }
 
 export function toSelector(selectorString: string): Selector {
-    // We parse the selector back-to-front, so that the front-most selectors end up at the bottom of the tree we create
-    // and therefore get resolved first.
-    if (selectorString === '') {
+    if (selectorString.includes(' ')) {
+        // ERROR CASE: The selector contains whitespace.
+        throw new Error(getMessage("SelectorLooksIncorrect", selectorString));
```
stephen-carter-at-sf: When using includes or regex testing, you are increasing the runtime since you are doing additional passes unnecessarily. We want this part of the code to be as performant as possible. So you should move the space check as in if statement inside of the loop since you are already checking character by character.

packages/code-analyzer-core/src/selectors.ts
```diff
@@ -5,81 +5,70 @@ export interface Selector {
...
-            const right: string = selectorString.slice(lastComma + lastCloseParen + 1);
-            return toComplexSelector(left, right, ',');
-        } else {
-            const left: string = selectorString.slice(0, lastColon + lastCloseParen);
-            const right: string = selectorString.slice(lastColon + lastCloseParen + 1);
-            return toComplexSelector(left, right, ':');
-        }
+    } else if (new RegExp('^[,:]').test(selectorString) || new RegExp('[,:]$').test(selectorString)) {
+        // ERROR CASE: The selector cannot start with a binary operator, because that's nonsense.
+        throw new Error(getMessage("SelectorStartsOrEndsWithOperator", selectorString));
```
stephen-carter-at-sf: Likewise, you can test the first and last character of each selectorString (since the algorithm is recursive) by just checking the first and last characters without a regular expression.

So 
```
} else if (selectorString.startsWith(',') ||  selectorString.startsWith(':') || selectorString.endsWith(',') ||  selectorString.endsWith(':')) {
```
would be better here.

Or better yet... you could just add it to the if statement in your loop below. For example
```
    if (i===0 && (char === ',' || char === ':'))  {
```
.

If this wasn't a recursive algorithm where the operations get multiplied then it wouldn't be as big of a deal - but it is, so we need to try to keep O(1) operations inside of a single O(n) loop if possible.

stephen-carter-at-sf (changes requested):

packages/code-analyzer-core/src/selectors.ts
stephen-carter-at-sf: You will catch it anyway already because all the parens need to be balanced on both sides of the top level comma anyway and so it'll get caught in the next iteration of the loop.... so I do believe that you can indeed short circuit when you find the first top level comma.

stephen-carter-at-sf (approved):
I'd still consider reducing complexity by removing the nested if statements. Consider:
```
} else if (parenBalance > 0) { // Ignore chars inside of parens
    continue;
} else if (char === ',') {
    const left: string = trimmedSelector.slice(0, i);
    const right: string = trimmedSelector.slice(i + 1);
    return new OrSelector(toSelector(left), toSelector(right));
} else if (char === ':' && firstColonIdx === null) {
    firstColonIdx = i;
}
```

But approved.

PR #369, opened 2025-10-15
NEW @W-19397402@ Implemented hidden config property for keeping working d…

jfeingold35:
…irectories

### DIFF
--- a/packages/code-analyzer-core/src/code-analyzer.ts
+++ b/packages/code-analyzer-core/src/code-analyzer.ts
@@ -326,14 +326,18 @@ export class CodeAnalyzer {
 
         const runPromises: Promise<EngineRunResults>[] = ruleSelection.getEngineNames().map(async (engineName) => {
             const workingFolder: string = await this.tempFolder.makeSubfolder(runWorkingFolderName, engineName);
+            if (this.config.getPreserveAllWorkingDirectories()) {
+                this.tempFolder.markToBeKept(runWorkingFolderName, engineName);
+            }
             const engineRunOptions: engApi.RunOptions = {
                 logFolder: this.config.getLogFolder(),
                 workingFolder: workingFolder,
                 workspace: engApiWorkspace
             };
             const errorCallback: () => void = () => {
+                // istanbul ignore else
                 if (!this.tempFolder.isKept(runWorkingFolderName, engineName)) {
-                    this.emitLogEvent(LogLevel.Debug, getMessage('EngineWorkingFolderKept', engineName, workingFolder));
+                    this.emitLogEvent(LogLevel.Debug, getMessage('EngineWorkingFolderKeptDueToError', engineName, workingFolder));
                     this.tempFolder.markToBeKept(runWorkingFolderName, engineName);
                 }
             };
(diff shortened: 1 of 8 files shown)


packages/code-analyzer-core/src/config.ts
```diff
@@ -41,6 +42,7 @@ type TopLevelConfig = {
     log_level: LogLevel
     rules: Record<string, RuleOverrides>
     engines: Record<string, EngineOverrides>
+    preserve_all_working_directories: boolean // INTERNAL USE ONLY
```
stephen-carter-at-sf: For properties that are potentially customer facing, let's be consistent. We are calling it "working folders" and not directories. So I propose we call this something like "preserve_working_folders_always" instead.

packages/code-analyzer-core/src/config.ts
```diff
@@ -226,6 +230,14 @@ export class CodeAnalyzerConfig {
         return this.config.custom_engine_plugin_modules;
     }
 
+    /**
+     * Returns a boolean indicating whether working directories should be preserved even if their engine succeeds, as opposed
+     * to being kept only for failed engines.
```
stephen-carter-at-sf: May I suggest:

```
     * Returns a boolean indicating whether working folders are preserved always.
     * If false, then the working folders for an engine are deleted unless the engine issues an error.
     * If true, then the working folders for each engine remain, even if an engine does not issues any error.
```

packages/code-analyzer-core/src/code-analyzer.ts
```diff
@@ -326,21 +326,29 @@ export class CodeAnalyzer {
...
                 }
             };
             const results: EngineRunResults = await this.runEngineAndValidateResults(engineName, ruleSelection, engineRunOptions, errorCallback);
             await this.tempFolder.removeIfNotKept(runWorkingFolderName, engineName);
             return results;
         });
+        if (this.config.getPreserveAllWorkingDirectories()) {
+            this.tempFolder.markToBeKept(runWorkingFolderName);
+            this.emitLogEvent(LogLevel.Debug, getMessage('AllWorkingFoldersKept', await this.tempFolder.getPath(runWorkingFolderName)));
+        }
```
stephen-carter-at-sf: This `this.tempFolder.markToBeKept(runWorkingFolderName);` isn't needed. Line 330 already marks the subfolders as kept and therefore the parent folder (runWorkingFolderName) is already kept by the way the temp folder preserves parents.

I'm on the fence about needing a single debug message. We could just debug each engine folder instead to keep a log code simple with the log being below line 330. But I see the motivation here - to have a single log for all things under the run folder.

packages/code-analyzer-core/src/code-analyzer.ts
```diff
@@ -393,6 +406,11 @@ export class CodeAnalyzer {
                 return rules;
             });
 
+            if (this.config.getPreserveAllWorkingDirectories()) {
+                this.tempFolder.markToBeKept(rulesWorkingFolderName);
```
stephen-carter-at-sf: ditto - not needed.

packages/code-analyzer-core/src/messages.ts
```diff
@@ -202,7 +202,10 @@ const MESSAGE_CATALOG : MessageCatalog = {
     EngineReturnedViolationWithCodeLocationWithEndColumnBeforeStartColumnOnSameLine:
         `Engine failure. The engine '%s' returned a violation for rule '%s' that contains a code location with the endLine equal to the startLine and the endColumn %d before the startColumn %d.`,
 
-    EngineWorkingFolderKept:
+    AllWorkingFoldersKept:
+        `All temporary working folders have been kept, and are located in: %s`,
```
stephen-carter-at-sf: I'm on the fence regarding this message. It will show up once for the rules and once for the run. My gut was to just for each engine's working folder have something similar to the EngineWorkingFolderKeptDueToError message... and just have each individual engine folder have a debug message that says "Since preserve_working_folders_always is true, the following temporary working folder will not be removed: %s"

Or if you wanted to keep it at a top level... then just use that message for the top level rules and run folders. I think the term "All" in your message is what is tripping me up because the way you say that "All" "are located in"... when in reality it is just all the run folders or all the rules folders.

stephen-carter-at-sf (changes requested):

stephen-carter-at-sf (approved):

PR #370, opened 2025-10-16
NEW @W-19397402@ Added hidden root_working_folder config property

jfeingold35:


packages/code-analyzer-core/src/config.ts
```diff
@@ -13,6 +13,7 @@ export const FIELDS = {
     CONFIG_ROOT: 'config_root',
     LOG_FOLDER: 'log_folder',
     LOG_LEVEL: 'log_level',
+    WORKING_FOLDERS_ROOT: 'working_folders_root', // Hidden
```
stephen-carter-at-sf: hmm - part of me wants to have this say "root_working_folder" instead... because it parallels "log_folder" in a way. I'm even on the fence wondering if we instead call this "working_folder" and remove root as it can be implied - sort of like how log_folder is a bit implied. But since it is hidden - we can do naming if we ever decide to make this exposed.

packages/code-analyzer-core/src/utils.ts
```diff
@@ -64,9 +64,9 @@ export class TempFolder {
     private rootFolder?: string;
     private relPathsToKeep: Set<string> = new Set();
 
-    constructor(fileSystem: FileSystem = new RealFileSystem(), rootFolderPrefix: string = path.join(os.tmpdir(), 'code-analyzer-')) {
+    constructor(fileSystem: FileSystem = new RealFileSystem(), rootFolderPath: string = os.tmpdir()) {
```
stephen-carter-at-sf, line 67: Notice how you have this say rootFolderPath - which is another indication that we are dealing with a rootWorkingFolder instead of a workingFoldersRoot. Just an observation.

packages/code-analyzer-core/src/utils.ts
```diff
@@ -64,9 +64,9 @@ export class TempFolder {
...
         this.fileSystem = fileSystem;
-        this.rootFolderPrefix = rootFolderPrefix;
+        this.rootFolderPrefix = path.join(rootFolderPath, 'code-analyzer-');
```
stephen-carter-at-sf, line 69: I see you removed option to supply the prefix. I think this is fine since I was on the fence of whether to do so before. Until we reuse the TempFolder for other purposes - this seems fine.

stephen-carter-at-sf (approved):

PR #372, opened 2025-10-20
FIX: @W-19990843@ Turn off jest/no-deprecated-functions rule

stephen-carter-at-sf:
This PR is a temporary workaround of the issue: https://github.com/salesforce/eslint-config-lwc/issues/161


randi274 (approved):
Looks good, and allows us to migrate to vitest while ensuring we're not showing an error to all users missing jest when loading this rule in the next release.

PR #374, opened 2025-10-20
CHANGE @W-19688830@ Updated node dependencies

jfeingold35:


packages/code-analyzer-eslint-engine/src/base-config.ts
```diff
@@ -56,22 +56,22 @@ export class BaseConfigFactory {
         }
         return configArray;
     }
-    
+
     private createJavascriptPlusLwcConfigArray(): Linter.Config[] {
         let configs: Linter.Config[] = validateAndGetRawLwcConfigArray();
-        
+
         // TODO: Remove the For the following 2 updates when https://github.com/salesforce/eslint-config-lwc/issues/158 is fixed
```
randi274, line 63: @jfeingold35 is turning this off still necessary? It looks like the original bug was turned off (not sure if this eslint plugin still needs to update or if that og bug being fixed is sufficient)

packages/code-analyzer-core/package.json
```diff
@@ -16,26 +16,26 @@
   },
   "types": "dist/index.d.ts",
   "dependencies": {
-    "@salesforce/code-analyzer-engine-api": "0.30.0",
-    "@types/node": "^20.0.0",
+    "@salesforce/code-analyzer-engine-api": "0.31.0-SNAPSHOT",
+    "@types/node": "^24.9.0",
```
stephen-carter-at-sf: I've gone back and forth on this and researched it to death without getting a great answer... should we be updating @types/node if we want to ensure support of node 20?

We probably can I don't know if there are any risks here - which is why I've kept these at 20.0.0 for a while. It also helps us not accidentally rely on a node type that is a later version I believe. Since core modules are imported by clients... we wouldn't want a client being prevented from compiling because it compiles using node 20. But maybe @types/node has versioned types?

packages/code-analyzer-engine-api/package.json
```diff
@@ -16,17 +16,17 @@
   },
   "types": "dist/index.d.ts",
   "dependencies": {
-    "@types/node": "^20.0.0"
+    "@types/node": "^24.9.0"
```
stephen-carter-at-sf: ditto for all of them.

packages/code-analyzer-eslint8-engine/package.json
```diff
@@ -18,10 +18,10 @@
     "@eslint/js": "8.57.1",
     "@lwc/eslint-plugin-lwc": "2.2.0",
     "@lwc/eslint-plugin-lwc-platform": "5.2.0",
-    "@salesforce/code-analyzer-engine-api": "0.30.0",
+    "@salesforce/code-analyzer-engine-api": "0.31.0-SNAPSHOT",
     "@salesforce/eslint-config-lwc": "3.7.2",
     "@salesforce/eslint-plugin-lightning": "1.0.1",
-    "@types/node": "^20.0.0",
+    "@types/node": "^24.9.0",
```
stephen-carter-at-sf: ditto

stephen-carter-at-sf (commented):
Maybe you can l

packages/code-analyzer-eslint-engine/src/base-config.ts
jfeingold35: The Issue is still open, so I think it might still be necessary.

randi274 (approved):

PR #396, opened 2026-01-16
NEW @W-20774650@ - Integrate with eslint-plugin-jsx-a11y plugin

aruntyagiTutu:
1- Add all the rules from flatConfigs.strict. [58 rules]

2- Add mapping for these rules in code analyzer.

3- Add UTs for these rules.

4- Manually test that the rules are working as expected.

### DIFF
--- a/packages/code-analyzer-eslint-engine/package.json
+++ b/packages/code-analyzer-eslint-engine/package.json
@@ -37,7 +37,6 @@
     "typescript-eslint": "^8.50.0"
   },
   "devDependencies": {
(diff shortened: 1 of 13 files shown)


packages/code-analyzer-eslint-engine/package.json
```diff
@@ -36,6 +37,7 @@
     "typescript-eslint": "^8.50.0"
   },
   "devDependencies": {
+    "@types/eslint-plugin-jsx-a11y": "^6.10.1",
```
namrata111f: We need to add this as devDependency as well?

namrata111f (approved):

PR #408, opened 2026-02-05
NEW @W-21101982@ Adding ignores object in config file

namrata111f:
Adding new ignore object in config file which users can use to add file paths/glob patterns to skip running rules for all the engines.

Testing:
Case-1) Tested with the ignore patterns:
```
ignores:
  files:
    - "**/*.js"
    - "**/*.jsx"
```

```
ignores:
  files:
    - "**/utils.js"
```
Violations reported skipped the files covered in these patterns.
Case-2) Tested with no ignores object in the config files. The output contained all the expected violations.


packages/code-analyzer-engine-api/src/workspace.ts
```diff
@@ -1,5 +1,6 @@
 import fs from "node:fs";
 import path from "node:path";
+import {Minimatch} from "minimatch";
```
aruntyagiTutu, line 3: Ignore patterns likely won’t match on Windows due to path separator mismatch. minimatch expects /-separated paths, but matchesIgnorePattern feeds it path.sep (backslashes on Windows). This will make ignores.files silently ineffective on Windows unless users write \\-escaped patterns. Normalize to POSIX separators before matching (e.g., pathToMatch.replaceAll(path.sep, "/")).

were you able to test it for windows ?

namrata111f: That's a good catch. But shouldn't windows user be giving the file path as '\\' as well in the ignore section if that is how the paths comes up for them?

namrata111f: Have added POSIX normalization for now.

aruntyagiTutu (approved):

PR #409, opened 2026-02-10
FIX @W-21101982@ Adding changes to avoid sfge scan when no target files are there

namrata111f:
SFGE was scanning all the Apex files in the workspace when the same file was present in the target and the same file was present in the ignore as well. This can impact the performance especially for the extension when we try to scan the file which is matching the ignored pattern.

### DIFF
--- a/packages/code-analyzer-sfge-engine/src/engine.ts
+++ b/packages/code-analyzer-sfge-engine/src/engine.ts
@@ -68,6 +68,13 @@ export class SfgeEngine extends Engine {
             return { violations: [] };
         }
 
+        // Get targeted files and return early if empty - prevents SFGE from analyzing all workspace files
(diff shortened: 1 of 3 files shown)


packages/code-analyzer-sfge-engine/src/engine.ts
```diff
@@ -68,6 +68,13 @@ export class SfgeEngine extends Engine {
             return { violations: [] };
         }
 
+        // Get targeted files and return early if empty - prevents SFGE from analyzing all workspace files
+        const targetedFiles: string[] = await runOptions.workspace.getTargetedFiles();
```
aruntyagiTutu, line 72: is it possible to ignore folder / files for sfge after this implementation ?

aruntyagiTutu (approved):

PR #410, opened 2026-02-13
NEW @W-19860299@ Adding changes to disable rules from being selected

namrata111f:
We are making this change to allow users to disable particular rules for all the files of their project.Testing details:
Case-1) When no value for the rule's disabled property is added:
```
Violation file paths relative to '/Users/namrata.gupta/workspace/dreamhouse/dreamhouse-sfdx/force-app/main/default/classes'.
  #    Severity   Rule                         Location              Message                                                 
 ────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
  1    5 (Info)   regex:NoTrailingWhitespace   BotMessage.cls:6:1    Found trailing whitespace at the end of a line of code.
  2    5 (Info)   regex:NoTrailingWhitespace   BotMessage.cls:12:1   Found trailing whitespace at the end of a line of code.
  3    5 (Info)   regex:NoTrailingWhitespace   BotMessage.cls:16:1   Found trailing whitespace at the end of a line of code.
  4    5 (Info)   regex:NoTrailingWhitespace   BotMessage.cls:20:1   Found trailing whitespace at the end of a line of code.
  5    5 (Info)   regex:NoTrailingWhitespace   BotMessage.cls:24:1   Found trailing whitespace at the end of a line of code.
  6    5 (Info)   regex:NoTrailingWhitespace   BotMessage.cls:28:1   Found trailing whitespace at the end of a line of code.
  7    5 (Info)   regex:NoTrailingWhitespace   BotMessage.cls:34:1   Found trailing whitespace at the end of a line of code.
  8    5 (Info)   regex:NoTrailingWhitespace   BotMessage.cls:40:1   Found trailing whitespace at the end of a line of code.
  9    5 (Info)   regex:NoTrailingWhitespace   BotMessage.cls:46:1   Found trailing whitespace at the end of a line of code.
  10   5 (Info)   regex:NoTrailingWhitespace   BotMessage.cls:50:1   Found trailing whitespace at the end of a line of code.
  11   5 (Info)   regex:NoTrailingWhitespace   BotMessage.cls:54:1   Found trailing whitespace at the end of a line of code.
  12   5 (Info)   regex:NoTrailingWhitespace   BotMessage.cls:58:1   Found trailing whitespace at the end of a line of code.
  13   5 (Info)   regex:NoTrailingWhitespace   BotMessage.cls:62:1   Found trailing whitespace at the end of a line of code.
  ```
  Case-2) When we explicitly disable the rule from config:
  ```
rules:
  regex:
    NoTrailingWhitespace:
      severity: 2
      disabled: true
```
No violations are reported for it and we are also adding an info log:
```
Selecting rules... done.
Code Analyzer [11:19:14.872]:
    Rule 'NoTrailingWhitespace' from engine 'regex' was disabled according to the specified configuration and will not be included in the rule selection.

Executing rules... done. Executed rules from regex, cpd, pmd, sfge.

=== Summary

Found 0 violations.
```
Case-3) If user tries to give any value other than boolean value for this property:
```
% sf code-analyzer run --workspace . -t ./force-app/main/default/classes/BotMessage.cls -r All -c /Users/namrata.gupta/workspace/dreamhouse/dreamhouse-sfdx/force-app/main/default/react-components/code-analyzer.yml
 ›   Warning: @salesforce/plugin-code-analyzer is a linked ESM module and cannot be auto-transpiled. Existing compiled source will be used instead.
 ›   Warning: @salesforce/cli update available from 2.113.6 to 2.122.6.
Error (1): The 'rules.regex.NoTrailingWhitespace.disabled' configuration value must be of type 'boolean' instead of type 'string'.
```

### DIFF
--- a/packages/code-analyzer-core/src/code-analyzer.ts
+++ b/packages/code-analyzer-core/src/code-analyzer.ts
@@ -301,9 +301,19 @@ export class CodeAnalyzer {
 
         const ruleSelection: RuleSelectionImpl = new RuleSelectionImpl();
         for (const rule of allRules) {
-            if (selectorObjects.some(o => rule.matchesRuleSelector(o))) {
-                ruleSelection.addRule(rule);
+            // Skip rules that don't match any selector
+            if (!selectorObjects.some(o => rule.matchesRuleSelector(o))) {
+                continue;
             }
+
+            // Skip rules that are disabled in the config
+            const ruleOverride = this.config.getRuleOverrideFor(rule.getEngineName(), rule.getName());
+            if (ruleOverride.disabled === true) {
+                this.emitLogEvent(LogLevel.Info, getMessage('RuleDisabledInConfig', rule.getName(), rule.getEngineName()));
+                continue;
+            }
+
+            ruleSelection.addRule(rule);
         }
 
         this.emitEvent({type: EventType.RuleSelectionProgressEvent, timestamp: this.clock.now(), percentComplete: 100});
--- a/packages/code-analyzer-core/src/code-analyzer.ts
+++ b/packages/code-analyzer-core/src/code-analyzer.ts
@@ -300,6 +300,8 @@ export class CodeAnalyzer {
         const allRules: RuleImpl[] = await this.getAllRules(selectOptions?.workspace);
 
         const ruleSelection: RuleSelectionImpl = new RuleSelectionImpl();
+        const disabledRules: {ruleName: string, engineName: string}[] = [];
+
         for (const rule of allRules) {
             // Skip rules that don't match any selector
             if (!selectorObjects.some(o => rule.matchesRuleSelector(o))) {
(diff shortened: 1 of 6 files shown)


packages/code-analyzer-core/src/code-analyzer.ts
```diff
@@ -301,9 +301,19 @@ export class CodeAnalyzer {
...
-                ruleSelection.addRule(rule);
+            // Skip rules that don't match any selector
+            if (!selectorObjects.some(o => rule.matchesRuleSelector(o))) {
+                continue;
             }
+
+            // Skip rules that are disabled in the config
+            const ruleOverride = this.config.getRuleOverrideFor(rule.getEngineName(), rule.getName());
+            if (ruleOverride.disabled === true) {
+                this.emitLogEvent(LogLevel.Info, getMessage('RuleDisabledInConfig', rule.getName(), rule.getEngineName()));
```
stephen-carter-at-sf: OK a few things. First, this should not be an "info" level log because then this would appear in the terminal in the CLI and would be unnecessarily noisy. This in my opinion should just be a Debug level message. This would also be consistent with the fact we use Debug log level for the RulePropertyOverridden message over at https://github.com/forcedotcom/code-analyzer-core/blob/2ed7f84d80a8ed12f4818390d5743cfb797fe29c/packages/code-analyzer-core/src/code-analyzer.ts#L631

Second, instead of having a ton of logs (one per disabled rule), I would encourage you to collect the disabled rules and have a single log that logs all the disabled rules at once that were skipped based on the provided rule selector.

stephen-carter-at-sf (changes requested):

packages/code-analyzer-core/src/code-analyzer.ts
namrata111f: Collecting the disabled rules and then logging makes sense, will add. But for the log level I intentionally kept it "INFO" level so that the user has visibility to the rules they have disabled in the run. While I agree existing ones are of the Debug level but they are not that much critical as well. We are just changing the severity and tags, the violations are still reported in output. @stephen-carter-at-sf do you think it would be too much of noise?

stephen-carter-at-sf: I would check with PM to see if they want it to be noisy or not. I personally think this should just end up in a log and not at the terminal.

aruntyagiTutu (approved):

PR #426, opened 2026-03-05
FIX @W-21462752@ - Smart parser selection to support decorators when LWC base confi…

aruntyagiTutu:
@W-21462752@

### User Impact

Customers who set `disable_lwc_base_config: true` in their Code Analyzer configuration experienced parsing failures when scanning Lightning Web Component (LWC) files. The scanner failed with the following error:

```
Error: Parsing error: Unexpected character '@'
```

This prevented customers from:

* Disabling overly strict LWC rules while still analyzing their LWC code
* Using custom ESLint configurations on LWC files
* Running Code Analyzer successfully on projects that use LWC decorators

### Customer Scenario

Customers may disable the LWC base configuration for valid reasons:

1. They want to use their own ESLint rules instead of the default LWC rules provided by Code Analyzer.
2. Some default LWC rules may be too strict for their specific use case.
3. They want more control over which rules apply to their codebase.

However, when `disable_lwc_base_config: true` was set, LWC files containing decorators such as `@api`, `@track`, and `@wire` failed to parse completely.

### Root Cause

The issue was related to how the ESLint engine selected the JavaScript parser.

Two parsers are available:

* **Espree** (ESLint default): fast and lightweight, supports JSX
* **Babel**: slightly slower but supports additional language features including decorators

The original logic worked as follows:

If `disable_lwc_base_config: true`
→ Use Espree parser for better performance

Else
→ Use Babel parser

The problem is that Espree does not support decorator syntax such as `@api` and `@track`, which are fundamental to LWC. As a result, switching to Espree caused LWC files to fail parsing.

The original design assumed that if a user disabled the LWC configuration, they likely did not have LWC files in their project. In practice, customers disable it mainly to avoid certain rules, not because they do not use LWC. This meant the configuration unintentionally affected the parser, even though its purpose was to control rules.

### Solution

The fix introduces smarter parser selection based on the file types being analyzed.

Updated logic:

If `.js` files are included in the configuration
→ Use the Babel parser, which supports both LWC decorators and React JSX

Else if only `.jsx`, `.mjs`, or `.cjs` files are included
→ Use the Espree parser for better performance

### Why This Works

For LWC projects:

* Default configuration includes `.js` files
* Babel parser is automatically selected
* LWC decorators work correctly without requiring any configuration changes

For React-only projects:

* If the configuration only includes `.jsx` files
* Espree parser is used, providing faster performance
* React projects do not require decorator support

For mixed projects:

* If `.js` files are present, Babel handles all JavaScript parsing
* This provides a safe default that supports both frameworks

### Benefits

For customers:

* Customers can now disable LWC rules without breaking parsing of LWC files
* No configuration changes are required
* Existing configurations continue to work without modification
* Users no longer see confusing errors such as “Unexpected character '@'”

For Code Analyzer:

* Reduces a common category of support issues
* Configuration behavior now matches user expectations
* Maintains performance benefits for React-only projects
* Both available parsers remain useful and correctly applied

### Testing

Manual testing was performed using the customer’s configuration.

Before the fix:

```yaml
eslint:
  disable_lwc_base_config: true
```

Result:

```
Error: Parsing error: Unexpected character '@'
```

After the fix:

```yaml
eslint:
  disable_lwc_base_config: true
```

Result:

```
Found 1 violation(s) across 1 file(s)
1 High severity violation(s) found.
```

Test scenarios covered include:

* `.js` files with LWC decorators and `disable_lwc_base_config: true`
* `.jsx` files containing React JSX
* Mixed `.js` and `.jsx` projects
* Configurations where both LWC and React rule sets are disabled

### Risk Assessment

Risk Level: Low

Reasons:

* The change only affects parser selection logic
* No breaking changes to existing configurations
* Babel parser usage increases slightly but only when required
* All pre-commit checks passed, including linting and version checks

Deployment details:

* Change is limited to the `eslint-engine` package
* Will be included in the next Code Analyzer release
* No migration or configuration updates are required for existing users

### Files Changed

* `packages/code-analyzer-eslint-engine/package.json` – version bump to `0.41.0-SNAPSHOT`
* `packages/code-analyzer-eslint-engine/src/base-config.ts` – updated parser selection logic

### Documentation

The configuration field `disable_lwc_base_config` is documented as:

“Whether to turn off the default base configuration that supplies the LWC rules for JavaScript files.”

After this change, the behavior now correctly matches the documentation. The setting disables LWC rules but no longer affects the ability to parse decorator syntax.

### Related Issues

This change resolves the issue where users could not use `disable_lwc_base_config` in projects that contain LWC files with decorators.

### DIFF
--- a/packages/code-analyzer-eslint-engine/src/base-config.ts
+++ b/packages/code-analyzer-eslint-engine/src/base-config.ts
@@ -148,17 +148,39 @@ export class BaseConfigFactory {
     }
 
     private createJavascriptConfigArray(): Linter.Config[] {
-        return [{
-            ... eslintJs.configs.all,
-            files: this.engineConfig.file_extensions.javascript.map(ext => `**/*${ext}`),
-            languageOptions: {
-                parserOptions: {
-                    ecmaFeatures: {
-                        jsx: true  // Enable JSX parsing for React/JSX files
+        // Smart parser selection based on file extensions:
+        // - .js files may contain LWC decorators (@api, @track, @wire) → need Babel
+        // - .jsx, .mjs, .cjs are typically React or modules without decorators → can use Espree (faster)
+        const hasJsExtension = this.engineConfig.file_extensions.javascript.includes('.js');
+
+        if (hasJsExtension) {
+            // .js files might have LWC decorators - use Babel parser with decorator support
+            const lwcConfig = validateAndGetRawLwcConfigArray()[0];
+            const babelParser = lwcConfig.languageOptions?.parser;
+            const babelParserOptions = lwcConfig.languageOptions?.parserOptions;
+
+            return [{
+                ... eslintJs.configs.all,
+                files: this.engineConfig.file_extensions.javascript.map(ext => `**/*${ext}`),
+                languageOptions: {
+                    parser: babelParser,
+                    parserOptions: babelParserOptions
+                }
+            }];
+        } else {
+            // Only .jsx, .mjs, .cjs (no .js) - use Espree for better performance
+            return [{
+                ... eslintJs.configs.all,
+                files: this.engineConfig.file_extensions.javascript.map(ext => `**/*${ext}`),
+                languageOptions: {
+                    parserOptions: {
+                        ecmaFeatures: {
+                            jsx: true  // Enable JSX parsing for React/JSX files
+                        }
                     }
                 }
-            }
-        }];
+            }];
+        }
     }
 
     private createSldsHTMLConfigArray(): Linter.Config[] {
--- a/packages/code-analyzer-eslint-engine/src/base-config.ts
+++ b/packages/code-analyzer-eslint-engine/src/base-config.ts
@@ -157,14 +157,26 @@ export class BaseConfigFactory {
             // .js files might have LWC decorators - use Babel parser with decorator support
             const lwcConfig = validateAndGetRawLwcConfigArray()[0];
             const babelParser = lwcConfig.languageOptions?.parser;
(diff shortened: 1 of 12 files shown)


packages/code-analyzer-eslint-engine/src/base-config.ts
```diff
@@ -148,17 +148,51 @@ export class BaseConfigFactory {
...
-                        jsx: true  // Enable JSX parsing for React/JSX files
+        // Smart parser selection based on file extensions:
+        // - .js files may contain LWC decorators (@api, @track, @wire) → need Babel
+        // - .jsx, .mjs, .cjs are typically React or modules without decorators → can use Espree (faster)
+        const hasJsExtension = this.engineConfig.file_extensions.javascript.includes('.js');
+
+        if (hasJsExtension) {
+            // .js files might have LWC decorators - use Babel parser with decorator support
+            const lwcConfig = validateAndGetRawLwcConfigArray()[0];
+            const babelParser = lwcConfig.languageOptions?.parser;
```
nikhil-mittal-165, line 159: will lwcConfig.languageOptions only have babel parser ?
as the variable is named  const babelParser

aruntyagiTutu: Good question LWC config is guaranteed to use Babel parser because LWC requires decorator support (@api, @track, etc.). Babel is the only ESLint parser that supports decorators, so this is a hard dependency of
  @salesforce/eslint-config-lwc

nikhil-mittal-165 (approved):

PR #431, opened 2026-03-06
FIX @W-21462752@ - Fix parser not configured when both JS and LWC base configs d…

aruntyagiTutu:
Issue:
When users set both disable_javascript_base_config: true AND disable_lwc_base_config: true, ESLint failed to parse .js files with decorators with error: 'Unexpected character @'

Root Cause:
The createBaseConfigArray() method had no fallback case when both base configs were disabled. This caused no parser to be configured at all, making ESLint default to Espree which cannot parse decorators.

Solution:
Added createMinimalJavascriptParserConfig() method that configures ONLY the parser (without base rules) when both configs are disabled:
- Uses Babel parser for .js files (supports decorators + JSX)
- Uses Espree parser for .jsx/.mjs/.cjs only (performance optimization)

Changes:
- base-config.ts: Added fallback case and createMinimalJavascriptParserConfig()
- parser-selection.test.ts: Added 6 comprehensive tests covering all edge cases
- package.json: Bumped version to 0.40.2-SNAPSHOT

Test Coverage:
- Both JS and LWC disabled with .js files
- Both disabled with only .jsx files
- Both disabled with .mjs/.cjs files
- Both disabled with mixed .js and .jsx files
- All three configs disabled (JS, LWC, React)
- No JavaScript extensions present

All 282 tests passing with 99.83% code coverage.

### DIFF
--- a/packages/code-analyzer-eslint-engine/src/base-config.ts
+++ b/packages/code-analyzer-eslint-engine/src/base-config.ts
@@ -60,6 +60,10 @@ export class BaseConfigFactory {
         }
         if (this.useTsBaseConfig()) {
             configArray.push(...this.createTypescriptConfigArray());
+        } else if (this.engineConfig.file_extensions.typescript.length > 0) {
+            // When TS base config is disabled, we still need to configure a parser for TypeScript files
+            // to avoid ESLint falling back to a parser that can't handle TypeScript syntax
+            configArray.push(...this.createMinimalTypescriptParserConfig());
         }
         // Add React plugin config for JSX files
         if (this.useReactBaseConfig()) {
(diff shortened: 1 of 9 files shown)


nikhil-mittal-165 (approved):

packages/code-analyzer-eslint-engine/src/base-config.ts
```diff
@@ -56,6 +102,10 @@ export class BaseConfigFactory {
         }
         if (this.useTsBaseConfig()) {
             configArray.push(...this.createTypescriptConfigArray());
+        } else if (this.engineConfig.file_extensions.typescript.length > 0) {
+            // When TS base config is disabled, we still need to configure a parser for TypeScript files
+            // to avoid ESLint falling back to a parser that can't handle TypeScript syntax
+            configArray.push(...this.createMinimalTypescriptParserConfig());
```
namrata111f, line 108: Do we need to introduce new function here? Can we directly use useTsBaseConfig? Or maybe combine this.useTsBaseConfig() with or of this.engineConfig.file_extensions.typescript.length > 0.

packages/code-analyzer-eslint-engine/src/base-config.ts
```diff
@@ -195,6 +245,87 @@ export class BaseConfigFactory {
...
+
+            // Add @babel/preset-react to support JSX in React files alongside LWC files
+            const enhancedParserOptions = {
+                ...originalParserOptions,
+                babelOptions: {
+                    ...originalBabelOptions,
+                    configFile: false,
+                    // Add React preset for JSX support (.jsx files and React in .js files)
+                    presets: [...(originalBabelOptions.presets || []), require.resolve('@babel/preset-react')]
+                }
```
namrata111f, line 270: Same here can we use the createJavascriptPlusLwcConfigArray and useJsBaseConfig instead here.

PR #433, opened 2026-03-09
FIX @W-21506501@ Adding changes to enable parsing by default

namrata111f:
Enabling parsing of the files irrespective of the config values selected. Currently we sometimes add parser for the file or sometimes skip it depending on config values. This can lead to some inconsistent states where parsing runs into error. Going forward we will have consistent behavior of adding the parser and skipping execution of the rules depending on the config values. 
Have manually tested for all the possible combinations of config values we can have: https://docs.google.com/document/d/18n5OmWVVRlqI3meBmWi1cSjnqbUEqLZ6nG1hVGsuwsU/edit?usp=sharing

### DIFF
--- a/packages/code-analyzer-eslint-engine/package.json
+++ b/packages/code-analyzer-eslint-engine/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@salesforce/code-analyzer-eslint-engine",
   "description": "Plugin package that adds 'eslint' as an engine into Salesforce Code Analyzer",
-  "version": "0.40.2",
+  "version": "0.41.0-SNAPSHOT",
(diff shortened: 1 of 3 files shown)


aruntyagiTutu (commented):
are existing tests enough to catch all scenrio with this optimization ?

namrata111f:
> are existing tests enough to catch all scenrio with this optimization ?

Have manually tested all the possible scenarios. Did not add any new UT since UTs were covered as part of your previous PR itself.

aruntyagiTutu (approved):

namrata111f:
Have added new UTs as well to verify unnecessary rules are not being executed depending on the config.

PR #438, opened 2026-03-13
NEW @W-20485780@ - Add Fixes and Suggestions to the Engine APITask Description

nikhil-mittal-165:
@W-20485780@ -
Define Fix and Suggestion types in engine-api, add core interfaces and implementations, update JSON/SARIF/XML output formats, and pass includeFixes/includeSuggestions flags through RunOptions to all engines.


packages/code-analyzer-retirejs-engine/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@salesforce/code-analyzer-retirejs-engine",
   "description": "Plugin package that adds 'retire-js' as an engine into Salesforce Code Analyzer",
-  "version": "0.32.0",
+  "version": "0.32.1-SNAPSHOT",
```
aruntyagiTutu: why patch ?

packages/code-analyzer-core/src/output-formats/results/sarif-run-results-format.ts
```diff
@@ -69,6 +69,35 @@ function toSarifResult(violation: Violation, runDir: string, ruleIndex: number)
...
         relatedLocations: violation.getCodeLocations().map(codeLoc => toSarifLocation(codeLoc, runDir))
     };
+    const fixes = violation.getFixes();
+    if (fixes.length > 0) {
+        result.fixes = fixes.map(fix => toSarifFix(fix, runDir));
+    }
+    return result;
+}
+
+function toSarifFix(fix: Fix, runDir: string): sarif.Fix {
```
aruntyagiTutu, line 79: The SARIF format shows toSarifFix() but no toSarifSuggestion():
  - Are suggestions also mapped to SARIF format?

packages/code-analyzer-core/src/code-analyzer.ts
```diff
@@ -755,6 +763,8 @@ function validateEngineRunResults(engineName: string, apiEngineRunResults: engAp
         validateViolationRuleName(violation, engineName, ruleSelection);
         validateViolationCodeLocations(violation, engineName);
         validateViolationPrimaryLocationIndex(violation, engineName);
+        validateFixCodeLocations(violation, engineName);
```
aruntyagiTutu, line 817: - Are these called automatically during the run process?
  - Or do consumers need to call them explicitly?
  - What happens if validation fails - is an error thrown?

aruntyagiTutu:
- How do individual engines know to include fixes/suggestions?
  - Is there a base interface or contract they implement?

aruntyagiTutu:
Are there any performance benchmarks or limits?

nikhil-mittal-165:
> * How do individual engines know to include fixes/suggestions?
>   
>   * Is there a base interface or contract they implement?

refer to this PR : https://github.com/forcedotcom/code-analyzer-core/pull/441

the include fixes and suggestions will be passed on to the respective engines where the decision is made

nikhil-mittal-165:
> Are there any performance benchmarks or limits?

currently only eslint provides fixes and suggestions [already providing we are just not using it as of now] so no changes on the performance that i can see

once apex guru comes in we do need to do some benchmarking

packages/code-analyzer-core/src/output-formats/results/sarif-run-results-format.ts
nikhil-mittal-165: sarif has the ability to add fixes but not suggestions 

according to spike doc

image

packages/code-analyzer-core/src/code-analyzer.ts
nikhil-mittal-165: Yes, these are called automatically during the run process. 
If validation fails (e.g., a fix has an invalid file path or out-of-range line number), it throws an error, same as the existing violation validation.

packages/code-analyzer-retirejs-engine/package.json
nikhil-mittal-165: updated

packages/code-analyzer-core/src/code-analyzer.ts
```diff
@@ -87,6 +87,12 @@ export type SelectOptions = {
 export type RunOptions = {
     /** Object that specifies the user's workspace and which files should be targeted in the run analysis. */
     workspace: Workspace
+
+    /** When true, engines should include fix data on violations when available. */
+    includeFixes?: boolean
+
+    /** When true, engines should include suggestion data on violations when available. */
+    includeSuggestions?: boolean
```
namrata111f, line 95: What will be the default value of these properties?

nikhil-mittal-165: default will be false check the cli PR run.ts file https://github.com/forcedotcom/code-analyzer/pull/2007/changes

aruntyagiTutu (approved):

namrata111f (approved):

PR #441, opened 2026-03-17
NEW @W-20485780@ - Add Fixes and Suggestions support in eslint

nikhil-mittal-165:


aruntyagiTutu:
implementation treats byte offsets as character indices, which will produce incorrect results for files with multi-byte characters

do you agree ?

aruntyagiTutu:
Assumes UTF-8 encoding. ESLint may analyze files with different encodings?

nikhil-mittal-165:
> implementation treats byte offsets as character indices, which will produce incorrect results for files with multi-byte characters
> 
> do you agree ?

we always read the sourxce file in utf-8 encoding see this (https://github.com/forcedotcom/code-analyzer-core/blob/d0493716c027dee6f6dd6b60ce4942ae23b800f3/packages/code-analyzer-eslint-engine/src/engine.ts#L58) 


this gets internally converted to javascript strings and the ranges are calculated based on character indices which will be uniform no matter the source encoding, i have added a test case now for multibyte files as well 

please check the commit titled "multibyte test case"

packages/code-analyzer-eslint-engine/src/engine.ts
```diff
@@ -276,6 +288,87 @@ function toViolation(file: string, resultMsg: Linter.LintMessage): Violation {
...
+                convertEslintSuggestion(file, s, lineStartOffsets));
+        }
+    }
+
+    return violation;
+}
+
+function getFileContent(filePath: string, cache: Map<string, string>): string {
+    if (!cache.has(filePath)) {
+        cache.set(filePath, fsSync.readFileSync(filePath, 'utf8'));
```
stephen-carter-at-sf: This is a big no-no... you will kill the performance and responsiveness of Code Analyzer if you add in sync based operations. Why are you not using promises?

stephen-carter-at-sf (changes requested):

packages/code-analyzer-eslint-engine/src/engine.ts
```diff
@@ -276,6 +288,87 @@ function toViolation(file: string, resultMsg: Linter.LintMessage): Violation {
...
+    }
+    return cache.get(filePath)!;
```
stephen-carter-at-sf: If there are a lot of files, your cache is going to get very very large I'm guessing.

All this code is sensitive... so please find the best approach and test on very large projects with lots of violations and measure performance before and after.

nikhil-mittal-165: not using cache anymore , did a comparison of before and after post your review comment fix ,

the repo i used was this contains 12k tarket file. (https://github.com/SalesforceFoundation/NPSP)

took 23- 24 seconds to run all engines (without sfge) in both cases with fixes and without fixes.



Screenshot 2026-03-24 at 12 45 13 PM



Screenshot 2026-03-24 at 12 49 05 PM



i have also attached the evidences of the fixes here testing all the scenarios 

https://docs.google.com/document/d/1dTAXfDjgbx2nLrytjGV2PkjCDUpUZGh8TwubaxbOEtk

packages/code-analyzer-eslint-engine/src/engine.ts
nikhil-mittal-165: fixed it will now consume it directly from lintResult

packages/code-analyzer-eslint-engine/src/engine.ts
```diff
@@ -276,6 +295,77 @@ function toViolation(file: string, resultMsg: Linter.LintMessage): Violation {
         }],
         primaryLocationIndex: 0
     };
+
+    if (lineStartOffsets) {
+        if (includeFixes && resultMsg.fix) {
+            violation.fixes = [convertEslintFix(file, resultMsg.fix, lineStartOffsets)];
+        }
```
namrata111f: nit: Nested if increases code complexity we can simply return violations if lineStartOffsets is null/empty. Something like this:
```
if (!lineStartOffsets) {
    return violation;
}
 if (includeFixes && resultMsg.fix) {
            violation.fixes = [convertEslintFix(file, resultMsg.fix, lineStartOffsets)];
 }
if (includeSuggestions && resultMsg.suggestions?.length) {
            violation.suggestions = resultMsg.suggestions.map(s =>
                convertEslintSuggestion(file, s, lineStartOffsets));
}

nikhil-mittal-165: fixed

stephen-carter-at-sf (approved):

aruntyagiTutu (approved):

PR #449, opened 2026-04-10
NEW @W-21741601@ Adding changes for bulk suppressions in core

namrata111f:
These changes are to add the bulk suppression feature.
Design doc: https://docs.google.com/document/d/15bDsqLY8V0WDai7yd-A3CdXIU9D6ceGAZt7V1g9eajw/edit?usp=sharing
Testing
Demo: https://drive.google.com/file/d/1ov3D2ummBoRqtM1QthabQ1_ar9MfHB8Z/view?usp=sharing
1- Added UTs
2- Manual Testing: https://docs.google.com/document/d/1J5kE4tG51JD5K612qTk35xFXAyhXyIbv2Vry09gZAmM/edit?usp=sharing


packages/code-analyzer-core/src/code-analyzer.ts
```diff
@@ -672,6 +745,9 @@ export class CodeAnalyzer {
         // Apply inline suppressions per-engine BEFORE emitting EngineResultsEvent
         engineRunResults = await this.applyInlineSuppressionsToEngineResults(engineRunResults);
 
+        // Apply bulk suppressions per-engine AFTER inline suppressions, still BEFORE emitting EngineResultsEvent
+        engineRunResults = this.applyBulkSuppressionsToEngineResults(engineRunResults);
+
```
nikhil-mittal-165, line 750: the analysis can produce non deterministic results for each scan as it depends on which engine execution completes first are we fine with this experience

packages/code-analyzer-core/src/config.ts
```diff
@@ -50,11 +50,21 @@ export type Ignores = {
     files: string[]
 }
 
+/**
+ * Rule for bulk suppression of violations
+ */
+export type BulkSuppressionRule = {
+    rule_selector: string;
+    max_suppressed_violations?: number | null;
```
nikhil-mittal-165, line 58: what if user enters -1 for max_suppressed_violations ?

namrata111f: That's a good catch. Currently we are treating it as good as zero. I can add validation to report it as error if negative numbers are used. Shall we do that because treating it as zero can be misleading for the user?

packages/code-analyzer-core/src/code-analyzer.ts
namrata111f: As discussed with John/Sudha we are okay with this experience for now. This can only happen in case of complex selectors where multiple engines are mixed in rule selector. We will call this out in documentation. Currently there are no network calls involved so engines' execution order do not vary over multiple runs but can start happening once network call is involved.

packages/code-analyzer-core/src/config.ts
namrata111f: Have started throwing error for -ve values.

nikhil-mittal-165 (approved):

aruntyagiTutu (approved):

PR #462, opened 2026-05-05
NEW @W-22178016@ Implement Org JWT Minting

nikhil-mittal-165:


packages/code-analyzer-apexguru-engine/src/services/ApexGuruAuthService.ts
```diff
@@ -86,4 +96,76 @@ export class ApexGuruAuthService {
...
+     *
+     * @param featureId - Feature ID for tracking (default: 'VibesService')
+     * @returns Promise<string> - The Org JWT token
+     * @throws Error if minting fails
+     */
+    async mintOrgJwt(featureId: string = 'VibesService'): Promise<string> {
+        const accessToken = this.getAccessToken();
+        const instanceUrl = this.getInstanceUrl();
+
+        const endpoint = `${instanceUrl}/ide/auth`;
```
namrata111f: We can use variables for passing feature id and using endpoint rather than directly hard-coding.

packages/code-analyzer-apexguru-engine/src/services/ApexGuruAuthService.ts
```diff
@@ -86,4 +96,76 @@ export class ApexGuruAuthService {
...
+        this.emitLogEvent(LogLevel.Fine, 'Minting Org JWT for SFAP API access');
+
+        try {
+            const response = await fetch(endpoint, {
```
namrata111f: This is boiler plate code we can consider moving to a helper file.

packages/code-analyzer-apexguru-engine/src/services/ApexGuruAuthService.ts
franciscoperezsammartino: This endpoint was also created by you? Im wondering which ECA are you using to mint the JWT from. Probably if this endpoint was also created by you, that is present there.

franciscoperezsammartino: Otherwise this code is to call a core endpoint, that should be fine. (my only concern is if security approves exposing an endpoint to mint a JWT (although this is not part of this story), im assuming that as you already have the opaque access token it might be ok)

nikhil-mittal-165: the endpoint already exists and we are just adding a feature id for our use case. and as you mentioned the jwt is generated by already existing access token only

namrata111f (approved):

jag-j (approved):

PR #491, opened 2026-07-28
CHANGE @W-22462048@ apex guru lib update

nikhil-mittal-165:
the PR involves dependency upgrade for code-analyzer-apexguru-engine and code-analyzer-core both only

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10704,7 +10704,7 @@
     },
     "packages/code-analyzer-apexguru-engine": {
       "name": "@salesforce/code-analyzer-apexguru-engine",
-      "version": "0.41.0-SNAPSHOT",
+      "version": "0.42.0-SNAPSHOT",
       "license": "BSD-3-Clause",
       "dependencies": {
         "@salesforce/code-analyzer-engine-api": "0.40.0-SNAPSHOT",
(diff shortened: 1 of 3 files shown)


aruntyagiTutu (commented):
The title (\"apex guru lib update\") and commit message (\"dependency upgrade\") suggest a dependency bump, but the diff only contains self-package version increments — `code-analyzer-apexguru-engine` 0.41.0→0.42.0-SNAPSHOT and `code-analyzer-core` 0.51.0→0.52.0-SNAPSHOT (plus the matching lockfile entries). No actual dependency (internal or third-party) version changed anywhere in the diff.

Per the PR Hygiene guideline, the description/title should match the code. Could you clarify:
- What library/dependency is actually being upgraded here? If this is prep for an upcoming ApexGuru library bump that hasn't landed yet, it'd help to say so in the PR description.
- If this is just a manual version bump ahead of a release (similar to what the automated POSTRELEASE PRs do), consider noting that explicitly so reviewers aren't looking for a dependency change that isn't there.

Not blocking since nothing here is incorrect, just flagging for clarity.

aruntyagiTutu (approved):

